PR #617, opened 2023-07-13
fix: nft descriptor base64

PaulRBerg:
@andreivladbrg, please review urgently 🙏

- Closes https://github.com/sablier-labs/v2-core/issues/616
- Closes https://github.com/sablier-labs/v2-core/issues/608
- Closes https://github.com/sablier-labs/v2-core/issues/607


andreivladbrg (commented):
You missed this:

https://github.com/sablier-labs/v2-core/blob/055159fc8a6a2deec278342233d41c1cd9b09ba4/src/SablierV2NFTDescriptor.sol#L95

It should be changed to: 

```solidity
assetAddress: vars.asset.toHexString()
```

The contracts formatting fails, I think we should set it of in the test file.

PaulRBerg:
Looks like it's working onw

PR #626, opened 2023-07-20
refactor: convert integration tests to unit tests

### LINKED ISSUE #620: Convert integration tests to unit tests
As per the discussion with Matt in https://github.com/sablier-labs/v2-core/discussions/603.

- [x] [`flashFee`](https://github.com/sablier-labs/v2-core/blob/f742cf602d9adee31942f64d1cbab2dc51c405eb/test/integration/concrete/comptroller/flash-fee/flashFee.t.sol)
- [x] [`setFlashFee`](https://github.com/sablier-labs/v2-core/blob/main/test/integration/concrete/comptroller/set-flash-fee/setFlashFee.t.sol)

PraneshASP:
Resolves #620


test/unit/concrete/comptroller/flash-fee/flashFee.t.sol
```diff
@@ -3,9 +3,23 @@ pragma solidity >=0.8.19 <0.9.0;
...
+contract FlashFee_Unit_Concrete_Test is Base_Test {
+    /*//////////////////////////////////////////////////////////////////////////
+                                  SET-UP FUNCTION
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function setUp() public virtual override {
+        Base_Test.setUp();
+
+        // Deploy V2 Core.
+        deployCoreConditionally();
```
PaulRBerg, line 17: This is unfortunately too wasteful - it deploys all core contracts when just the comptroller is needed.

What we should do instead is:

- Write a new `Comptroller_Concrete_Test` contract in `test/unit/comptroller` (similar to [`NFTDescriptor`](https://github.com/sablier-labs/v2-core/blob/3df030516c7e9044742313c7cf17f15fdc1e9b05/test/unit/concrete/nft-descriptor/NFTDescriptor.t.sol#L26), which inherits from `Base_Test`
- Write a new `deployComptrollerConditionally` utility in that test contract
- Call the new deployer function in `setUp`
- Inherit from `Comptroller_Concrete_Test` instead of `Base_Test`

PaulRBerg (changes requested):
Thanks for your PR! Left a suggestion below

test/unit/concrete/comptroller/flash-fee/flashFee.t.sol
PraneshASP: Thanks for the clarification. Update the tests as mentioned above :+1:

PaulRBerg: Thanks, @PraneshASP. I turned this into a `Comptroller_Unit_Shared_Test` contract so that the comptroller fuzz tests could be moved to `unit`, too.

PaulRBerg (approved):
I pushed a commit to move these [comptroller fuzz tests](https://github.com/sablier-labs/v2-core/tree/3df030516c7e9044742313c7cf17f15fdc1e9b05/test/integration/fuzz/comptroller) over to `unit`, but otherwise LGTM.

Thanks @PraneshASP!

PR #632, opened 2023-07-21
chore: mv remappings to foundry config

pegahcarter:
Simple improvement to the codebase for one less file in the base directory.


PaulRBerg:
Hey, thanks for your PR. We want to keep the remappings in a separate file because:

1. Some VSCode extensions don't work with `foundry.toml`-based remappings.
2. They can be seen more easily in this way.

PR #641, opened 2023-07-30
refactor: using Given, When and Then keywords in branching Trees

### LINKED ISSUE #621: Separate current state from current action in Branching Trees
## Problem

Copy-pasting @DrakeEvans's feedback from a private chat on Telegram:

> I mostly see the "when" keyword on the branching tree here, do you differentiate between a given state and action in the `.tree` files?  I typically think of tests in the Arrange, Act, Assert phases which correlate to Given, When, Then in Cucumber style notation.  But it looks like you are combining the current state of the system and the current action into a single statement like "when the admin is not the same as the current admin" (line 8).  I think this is a little confusing (or at least it was to me on the first read) because it's not clear in which instances we are talking about function arguments and where we are talking about the state of the system without context clues.  And admittedly this situation wasn't too hard to parse.
>
> Looking at [`transferAdmin`](https://github.com/sablier-labs/v2-core/blob/main/test/unit/concrete/adminable/transfer-admin/transferAdmin.tree):
>
> - On line 2, "the admin" refers to the currently set admin (state of the system)
> - On line 8, "the admin" refers to the function parameter, and "current admin" is referring to the state of the system
>
> To be clear I really like the branching idea for keeping the information DRY, but I wonder if there is a way to make it slightly more explicit without the need for a full description of the prior state which can be verbose.  Some way to delineate between how the action takes place and the current state prior to an action

Drake is absolutely right - our tree terminology commingles the current state (contract state) with the current action (function arguments). It would be helpful to draw a distinction between the two

## Potential Solution

Follow Cucumber's [Gherkin](https://cucumber.io/docs/gherkin/) syntax and replace "when" with "given" for the tree nodes that refer to the contract state.

## Related

- ETHDenver talk: [Better Smart Contracts with Behavior-Driven Development](https://www.youtube.com/watch?v=T70txXsSxig)

smol-ninja:
Partially addresses #621.

### DIFF
--- a/test/integration/concrete/comptroller/protocol-fees/protocolFees.tree
+++ b/test/integration/concrete/comptroller/protocol-fees/protocolFees.tree
@@ -1,5 +1,5 @@
 protocolFees.t.sol
 ├── when the protocol fee has not been set
-│  └── it should return zero
+│  └── then should return zero
 └── when the protocol fee has been set
-   └── it should return the correct protocol fee
+   └── then should return the correct protocol fee
--- a/test/integration/concrete/comptroller/protocol-fees/protocolFees.tree
+++ b/test/integration/concrete/comptroller/protocol-fees/protocolFees.tree
@@ -1,5 +1,5 @@
 protocolFees.t.sol
 ├── when the protocol fee has not been set
-│  └── then should return zero
+│  └── it should return zero
 └── when the protocol fee has been set
-   └── then should return the correct protocol fee
+   └── it should return the correct protocol fee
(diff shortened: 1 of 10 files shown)


PaulRBerg (approved):
Thanks for this, @smol-ninja. It's looking good!

I changed my mind about "then" - "it" was better.

Pushed a commit to undo that change, but otherwise LGTM, merging now.

smol-ninja:
> I changed my mind about "then" - "it" was better.

Yeah "it" looked better to me as well. Other idea I had was using "then it" but that could have introduced 2 keywords for the same job.

PaulRBerg:
agree

PR #642, opened 2023-07-31
refactor: use of "Given" keyword in test contracts

smol-ninja:
fully addresses #621.


smol-ninja:
Closing until we resolve this [discussion](https://github.com/sablier-labs/v2-core/discussions/647).

test/integration/concrete/lockup/is-cancelable/isCancelable.tree
```diff
@@ -2,9 +2,9 @@ isCancelable.t.sol
 ├── when the id references a null stream
 │  └── it should revert
 └── when the id does not reference a null stream
-   ├── when the stream is cold
+   ├── given the stream is cold
    │  └── it should return true
-   └── when the stream is not cold
+   └── given the stream is not cold
       ├── when the stream is cancelable
```
PaulRBerg, line 8: Should be "given when the stream is cancelable".

test/integration/concrete/lockup/status-of/statusOf.t.sol
```diff
@@ -31,27 +31,27 @@ abstract contract StatusOf_Integration_Concrete_Test is Integration_Test, Lockup
         assertEq(actualStatus, expectedStatus);
     }
 
-    modifier whenAssetsNotFullyWithdrawn() {
+    modifier givenwhenAssetsNotFullyWithdrawn() {
```
PaulRBerg, line 34: Typo

test/integration/concrete/lockup/status-of/statusOf.tree
```diff
@@ -2,12 +2,12 @@ statusOf.t.sol
...
+   ├── given assets have been fully withdrawn
    │  └── it should return DEPLETED
-   └── when assets have not been fully withdrawn
-      ├── when the stream has been canceled
+   └── given assets have not been fully withdrawn
+      ├── given the stream has been canceled
       │  └── it should return CANCELED
-      └── when the stream has not been canceled
+      └── given the stream has not been canceled
          ├── when the start time is in the future
```
PaulRBerg, line 11: Should be "given when the start time is in the future"

And same for the "given when the refundable amount is zero" branch

test/integration/concrete/lockup/withdraw-max-and-transfer/withdrawMaxAndTransfer.tree
```diff
@@ -8,9 +8,9 @@ withdrawMaxAndTransfer.t.sol
       ├── when the caller is not the current recipient
       │  └── it should revert
       └── when the caller is the current recipient
-         ├── when the NFT has been burned
+         ├── given the NFT has been burned
          │  └── it should revert
-         └── when the NFT has not been burned
+         └── given the NFT has not been burned
             ├── when the withdrawable amount is zero
```
PaulRBerg, line 14: Should be "given when the withdrawable amount is zero" since this is a background property.

test/integration/concrete/nft-descriptor/safe-asset-decimals/safeAssetDecimals.tree
```diff
@@ -1,5 +1,8 @@
...
-│  └── it should return 0
-└── when the call does not revert
-   └── it should return the correct value
+├── when asset's decimal is not defined
+│   ├── when asset is an EOA
+│   │   └── it should return 0
+│   └── when asset is a non ERC20 compliant token
+│       └── it should return 0
+└── when asset's decimal is defined
+    └── it should return the correct decimal value
```
PaulRBerg, line 8: Unneeded extra spaces

test/integration/concrete/nft-descriptor/safe-asset-symbol/safeAssetSymbol.tree
```diff
@@ -1,8 +1,8 @@
...
-   │  └── it should return a hard-coded value
-   └── when the symbol is not long
-      └── it should return the correct value
+├── when a non ERC20 compliant asset
+│   └── it should return a hard-coded value: ERC20
+└── when an ERC20 asset
+    ├── when the symbol is longer than 30 characters
+    │   └── it should return a hard-coded value: Long Symbol
+    └── when the symbol is not longer than 30 characters
+        └── it should return the correct symbol value
```
PaulRBerg, line 8: Unneeded extra spaces

PaulRBerg (changes requested):
Thanks, @smol-ninja. This is fantastic work. I've left several specific comments below, but here's some additional feedback:

- All "when stream's status is X" branches should be turned into "given stream's status is X". The status of the stream is a background property.
- All `whenNotNull` modifiers should be turned into `givenNotNull`, and same for their associated branches. This is because the value that an id points to is in storage, so it's something "prepared" before the contract call.
- Any branches that use the phraseology "in the future" or "in the past" should be refactored to use "given" instead of "when" because time is not something in the user's control. E.g. `streamedAmountOf.tree` was left unchanged.

PaulRBerg:
I have accidentally merged this PR.

In any case, if there's any change I've done that you don't like, feel free to make another PR, @smol-ninja.

smol-ninja:
No problem. But it was puzzling to see getting the PR merged while I was still reviewing it @PaulRBerg. Nevertheless, I am reviewing your changes and leaving comments / opening a new PR to resolve them.

PaulRBerg:
Thanks @smol-ninja. Sorry for that.

test/integration/concrete/lockup/withdraw-max-and-transfer/withdrawMaxAndTransfer.tree
smol-ninja: Right. Changed the modifier name as well: https://github.com/sablier-labs/v2-core/pull/657

smol-ninja:
> Any branches that use the phraseology "in the future" or "in the past" should be refactored to use "given" instead of "when" because time is not something in the user's control.

Agree that time is not something in the user's control but shouldn't it still to be seen as a parameter to be picked from outside the system and not something from its previous state? I think of time as a hidden parameter that you can pick at runtime during the mode of execution. 

And very thanks for making the corrections. There is one change though. You changed from `When` to `Given` in case of withdrawal amount but modifier name remains to be using `When`. Please check PR https://github.com/sablier-labs/v2-core/pull/657 for this.

PaulRBerg:
> shouldn't it still to be seen as a parameter to be picked from outside the system and not something from its previous state?

Good question. On second thoughts, I am not so confident in my choice.

Time is fuzzy because there are two ways it can be interpreted:

1. Background state since `block.timestamp` can be manipulated by the Foundry dev
2. User-provided parameter, e.g., `startTime`

But I would be inclined to go with the 2nd approach because of the order of operations: the parameter is provided after the background state is configured.

Therefore, we might want to switch back to "when". I will create a PR.

> And very thanks for making the corrections

I thank you!

> You changed from When to Given in case of withdrawal amount but modifier name remains to be using When

Good catch.

PR #664, opened 2023-08-23
perf: dry "isCold" and "isWarm"

PaulRBerg:
Closes https://github.com/sablier-labs/v2-core/issues/610


PaulRBerg:
@smol-ninja because only `isWarm` and `isCold` can be moved upstream in the abstract contract.

Most other functions, including `isDepleted`, interact with the `LockupLinear.Stream` and `LockupDynamic.Stream` structs. And this means that the implementations cannot be de-duplicated.

Related: [Solidity does not have anything similar to Rust traits](https://twitter.com/PaulRBerg/status/1531652974161838082).

PR #678, opened 2023-09-04
test: use "given" instead of "when"

PaulRBerg:
Reference: https://github.com/sablier-labs/v2-periphery/pull/178#issuecomment-1705151676

### DIFF
--- a/test/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.t.sol
+++ b/test/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.t.sol
@@ -91,7 +91,7 @@ contract StreamedAmountOf_LockupDynamic_Integration_Concrete_Test is
         assertEq(actualStreamedAmount, expectedStreamedAmount, "streamedAmount");
     }
 
-    modifier whenCurrentMilestoneNot1st() {
+    modifier givenCurrentMilestoneNot1st() {
         _;
     }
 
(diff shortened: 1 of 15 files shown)


andreivladbrg (changes requested):
There are a few places where you've missed to change to "given". I will address it my next commit.

PaulRBerg:
Thanks @andreivladbrg.

when you push the fix commit yourself, you can mark the PR as accepted, to avoid the back-and-forth.

PaulRBerg:
> There are a few places where you've missed to change to "given". I will address it my next commit.

I did not miss those. Please read my comment above and delete your latest commit.

> Lmk if the changes made are correct

they are not

andreivladbrg:
> The rationale here is that "block.timestamp" is not something user-provided.

Yes, ofc it isn't, but `startTime` `cliffTime` and `stopTime` are user provided which are contract state variables.

Please tell me that if I understand it correctly: 
- state variable -> given
- user provided variables -> when

You've commited these [changes](https://github.com/sablier-labs/v2-core/pull/678/commits/56544e0d997f1a04f9f5352182ed3405571f542c): 
image

I've commited these [changes](https://github.com/sablier-labs/v2-core/pull/678/commits/8886dc500faf9790e83c4c3eb65767cf346f4fc9):
image


I fail to understand why you would use "given" for the `streamedAmountOf` function but not for `withdraw` function, which calls the `streamedAmountOf` internally. They are exactly the same to me and if you decide to use "given" for one of them you should for the other.

Please compare the changes made by you and the changes made by me and please tell me if there are any differences between them.

PaulRBerg:
Apologies @andreivladbrg, you are absolutely right.

I thought that the changes you applied were for the create function. But they are for the withdraw functions.

> Please tell me that if I understand it correctly:
>
> - state variable -> given
> - user provided variables -> when

As a heuristic, this is okay.

PR #679, opened 2023-09-05
feat: reverse mapping to retrieve stream Ids of a recipient

smol-ninja:
Addresses https://github.com/sablier-labs/v2-core/issues/644. @PaulRBerg lmk if you need any changes to the PR. I'll address them myself.

### DIFF
--- a/src/SablierV2LockupDynamic.sol
+++ b/src/SablierV2LockupDynamic.sol
@@ -62,6 +62,9 @@ contract SablierV2LockupDynamic is
     /// @dev Sablier V2 Lockup Dynamic streams mapped by unsigned integer ids.
     mapping(uint256 id => LockupDynamic.Stream stream) private _streams;
 
+    /// @dev Sablier V2 Lockup Dynamic streamIds mapped by recipient addresses.
+    mapping(address recipient => uint256[] streamIds) private _streamsByUser;
+
     /*//////////////////////////////////////////////////////////////////////////
                                      CONSTRUCTOR
     //////////////////////////////////////////////////////////////////////////*/
@@ -166,6 +169,11 @@ contract SablierV2LockupDynamic is
         if (_statusOf(streamId) == Lockup.Status.SETTLED) {
             stream.isCancelable = false;
         }
+    }
+
+    /// @inheritdoc ISablierV2LockupDynamic
+    function getStreamsByUser(address recipient) external view returns (uint256[] memory) {
+        return _streamsByUser[recipient];
     }
 
     /// @inheritdoc ISablierV2Lockup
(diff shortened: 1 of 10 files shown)


src/SablierV2LockupDynamic.sol
```diff
@@ -62,6 +62,9 @@ contract SablierV2LockupDynamic is
     /// @dev Sablier V2 Lockup Dynamic streams mapped by unsigned integer ids.
     mapping(uint256 id => LockupDynamic.Stream stream) private _streams;
 
+    /// @dev Sablier V2 Lockup Dynamic streamIds mapped by recipient addresses.
+    mapping(address recipient => uint256[] streamIds) private _streamsByUser;
```
andreivladbrg, line 66: Let's rename to `_recipientStreamIds`

src/SablierV2LockupDynamic.sol
```diff
@@ -168,6 +171,11 @@ contract SablierV2LockupDynamic is
         }
     }
 
+    /// @inheritdoc ISablierV2LockupDynamic
+    function getStreamsByUser(address recipient) external view returns (uint256[] memory) {
```
andreivladbrg, line 175: Let's rename to `getRecipientStreamIds`

test/integration/concrete/lockup-linear/get-streams-by-user/getStreamsByUser.tree
```diff
@@ -0,0 +1,5 @@
+ getStreamsByUser.t.sol
+ ├── when the recipient has no registered stream
+ │   └── it should return an empty array
+ └── when the recipient has atleast one registered stream
+     └── it should return an array of stream Ids belonging to that recipient
```
andreivladbrg, line 5: Let's say "it should return the correct stream ids array"

PaulRBerg:
Hey @smol-ninja, apologies for merging this quickly without waiting for your follow-up. We should have kept working on this PR.

@andreivladbrg - can you please explain to @smol-ninja what the issue is so that he can make another PR against our new branch `feat/reverse-mapping` in which he fixes it?

src/SablierV2LockupDynamic.sol
PaulRBerg: I prefer `_streamsByUser` - it's clearer because it shows that it's the same stream ids, but filtered by the user address.

The `recipient` key should be renamed to `user`.

src/SablierV2LockupDynamic.sol
PaulRBerg: See https://github.com/sablier-labs/v2-core/issues/644#issuecomment-1706320511

PR #682, opened 2023-09-07
refactor: disable sender to withdraw

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/674

Note: before merging this PR into 2.1 we should merge https://github.com/sablier-labs/v2-core/pull/683 into `refactor/disable-sender-to-withdraw`.


PaulRBerg:
> before merging this PR into 2.1 we should merge https://github.com/sablier-labs/v2-core/pull/683 into refactor/disable-sender-to-withdraw.

Actually, I will first review this PR, and then we can change the base branch of #683 to `main`.

andreivladbrg:
I started that branch from this one, merging that one into main would mean to merge all 2.1 changes.

PaulRBerg:
Yes, that's why I said we will change the base branch to `main` *after* reviewing and merging this PR.

PaulRBerg:
Let's put this PR on hold until we figure out the best way to handle the integration with NFTfi:

- https://github.com/sablier-labs/v2-core/discussions/684
- https://github.com/sablier-labs/v2-core/discussions/685

andreivladbrg:
@PaulRBerg Any updates?

PaulRBerg:
I'm still on the fence about this change.

My current thinking is that if we remove the sender's ability to withdraw, then we should definitely consider implementing a [relayer system](https://github.com/sablier-labs/v2-core/discussions/155#discussioncomment-6877117). Could you share your thoughts on that suggestion, @andreivladbrg?

andreivladbrg:
> hen we should definitely consider implementing a https://github.com/sablier-labs/v2-core/discussions/155#discussioncomment-6877117

The complexity might not be worth it. I suggest that we make a statistic on how many times the sender has called the `withdraw` function in the v1 and v2 contracts. If the percentage is too low, we shouldn't proceed. If it's not too low, we should reconsider.

PaulRBerg:
> The complexity might not be worth it

Can you please have a look at the conversation history in that other thread and provide some more concrete explanations?

> I suggest that we make a statistic on how many times the sender has called the withdraw function

agree

PaulRBerg:
I have just realized that retaining the sender's ability to withdraw is crucial for airstreams.

The creator of the airstream campaign can't know what kind of addresses there will be among the recipients. Some of those addresses might be contracts that are not able to call `withdraw`. And in these cases, the sender can come to the rescue and perform the airdrop on behalf of those recipients (presumably via `withdrawAll` to withdraw for multiple recipients).

Coupled with the recent [realization](https://github.com/sablier-labs/v2-core/discussions/684#discussioncomment-7056765) that not all NFT lending protocols are liable to the same issues as NFTfi, it should be fine to close this PR for the time being.

PR #686, opened 2023-09-08
feat: emit metadata in `_afterTokenTransfer` hook

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/670


src/abstracts/SablierV2Lockup.sol
```diff
@@ -324,6 +324,20 @@ abstract contract SablierV2Lockup is
                              INTERNAL CONSTANT FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @notice Overrides the internal ERC-721 transfer function to update the NFT SVG on external platforms by emitting
+    /// an event upon transfer.
+    /// @dev This event is also emitted when the NFT is minted or burned.
+    function _afterTokenTransfer(
+        address,
+        address,
```
PaulRBerg, line 332: Let's add comments after the commas to make it clear what these parameters stand for, e.g. like we do in the [proxy plugin](https://github.com/sablier-labs/v2-periphery/blob/9cf7a9a2d2c7c60a96efe0b02a155a243fefc819/src/SablierV2ProxyPlugin.sol#L77-L82).

PaulRBerg (changes requested):
Generally LGTM but we also need to write a test that shows that the metadata event gets emitted on `transferFrom`.

src/abstracts/SablierV2Lockup.sol
andreivladbrg: 1. I followed your `_beforeTokenTransfer` implementation and didn't add comments for the unused function parameters
 https://github.com/sablier-labs/v2-core/blob/a31dec3f43af2386c337c44f944c1cc5a52fa6a9/src/abstracts/SablierV2Lockup.sol#L331
2. The name of the ID is not the same as `{ERC721._afterTokenTransfer}`: `firstTokenId` vs `streamId`

Given the suggestion, we should provide commented names for both hooks. Should the comments mirror the actual hook parameter names?

andreivladbrg:
> but we also need to write a test that shows that the metadata event gets emitted on transferFrom.

We should add for both before and after hooks tests, then.

src/abstracts/SablierV2Lockup.sol
PaulRBerg: 1. I made a mistake there. You should always feel free to criticize and poke holes in my implementations.
2. Using `streamId` should be fine

PaulRBerg:
> We should add for both before and after hooks tests

Yes.

andreivladbrg:
@PaulRBerg pushed 2 more commits. Lmk if there is anything left

PaulRBerg (approved):
LGTM, let's wait for the CI to finish and then you can squash and merge, @andreivladbrg.

PR #688, opened 2023-09-15
build: upgrade solidity version to 0.8.21

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/483


PaulRBerg (changes requested):
We don't need to bump the pragma to `>=0.8.21`. We can keep it to `>=0.8.19` to enable compatibility with as many external projects as possible.

The reason why we have to use `0.8.19` is because this is what is used by PRBMath (which uses user-defined operators).

andreivladbrg:
>  We can keep it to >=0.8.19 to enable compatibility with as many external projects as possible

Makes sense, going to push a commit

andreivladbrg:
Lmk if there is anything else. If not, can you approve the PR?

PaulRBerg (approved):
I've pushed one commit to add the new `show_unsupported` and `show_proved_safe` STMChecker settings (see the release notes for [0.8.20](https://github.com/ethereum/solidity/releases/tag/v0.8.20)).

And I've also re-generated the gas snapshot.

PR looks good now. Good job.

PR #692, opened 2023-09-20
refactor: change the streamed amount to total amount in NFT SVG

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/638


PaulRBerg:
@andreivladbrg can you please share an example of an SVG generated with the new code in this PR?

shell/generate-svg.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Notes:
-# - There are four input arguments: progress, status, streamed amount, and duration
+# - There are four input arguments: progress, status, total, and duration
```
PaulRBerg, line 4: Fine to say "total amount" here

src/libraries/NFTSVG.sol
```diff
@@ -58,9 +58,9 @@ library NFTSVG {
         (vars.statusWidth, vars.statusCard) =
             SVGElements.card({ cardType: SVGElements.CardType.STATUS, content: params.status });
 
-        // Generate the streamed card.
-        (vars.streamedWidth, vars.streamedCard) =
-            SVGElements.card({ cardType: SVGElements.CardType.STREAMED, content: params.streamed });
+        // Generate the total card.
```
PaulRBerg, line 61: "total amount"

PaulRBerg (approved):
Generally LGTM. Paging @razgraf so he can double-check that the content looks good to him, too.

PaulRBerg:
Seeing the image now makes me thing that "Amount" might have been clearer ..

razgraf:
> Seeing the image now makes me thing that "Amount" might have been clearer ..

Looks good.
Q1: Although "Amount" can refer to a couple of different amounts (total, streamed, withdrawn) I think it's ok to leave it at "Amount" - seems clearer to me
Q2: Any value in adding the streamed amount on the side?

PaulRBerg:
> I think it's ok to leave it at "Amount" - seems clearer to me

Agree

> Should we change the label?

Yes. Sorry @andreivladbrg 🙈

> Any value in adding the streamed amount on the side?

No. Nobody will notice it (most folks in this space have ADHD).

Let's wait until a user requests the streamed amount.

andreivladbrg:
Updated the card, now it looks like this:

image

shell/generate-svg.sh
```diff
@@ -12,7 +12,7 @@ set -euo pipefail
 # Load the arguments while using default values
 arg_progress=${1:-4235}
 arg_status=${2:-"Streaming"}
-arg_streamed=${3:-"1.23M"}
+arg_amoung=${3:-"1.23M"}
```
PaulRBerg, line 15: Typo

PaulRBerg (approved):
✔️

PR #694, opened 2023-09-26
build: remove oz peer deps

PaulRBerg:
Closes https://github.com/sablier-labs/v2-core/issues/635.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,6 @@
     "token-streaming"
   ],
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


andreivladbrg:
Let's also add "^" for prb math: `"@prb/math": "^4.0.1"`

PaulRBerg:
Nope. Future versions of PRBMath might contain different logic, incompatible with the logic used by Sablier V2.0.

Let me have a think about that before we change it

PR #700, opened 2023-10-02
refactor: capitalize immutable variables 

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/699

Note: being a small change I think this shouldn't be included in the `CHANGELOG` .

### DIFF
--- a/test/fork/Fork.t.sol
+++ b/test/fork/Fork.t.sol
@@ -12,8 +12,8 @@ abstract contract Fork_Test is Base_Test {
                                      CONSTANTS
     //////////////////////////////////////////////////////////////////////////*/
 
-    IERC20 internal asset;
-    address internal holder;
+    IERC20 internal immutable ASSET;
+    address internal immutable HOLDER;
 
     /*//////////////////////////////////////////////////////////////////////////
                                      VARIABLES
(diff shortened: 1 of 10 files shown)


test/fork/Fork.t.sol
```diff
@@ -12,8 +12,8 @@ abstract contract Fork_Test is Base_Test {
                                      CONSTANTS
     //////////////////////////////////////////////////////////////////////////*/
 
-    IERC20 internal immutable asset;
-    address internal immutable holder;
+    IERC20 internal asset;
+    address internal holder;
```
PaulRBerg, line 16: Why have you removed `immutable` here? Let's make them uppercase

PaulRBerg (changes requested):
I agree that we should not include this in the changelog. This PR has nothing do with the public API.

test/fork/Fork.t.sol
andreivladbrg: It was easier to remove the state, it is not relevant here to use `immutable` but as you wish.

PaulRBerg: It is relevant to have these variables remain immutable - the goal is to prevent accidental modifications.

PaulRBerg (approved):
now it looks good

PR #718, opened 2023-10-26
refactor: emit event before calling hooks

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/717


PaulRBerg (approved):
perfect

PR #720, opened 2023-10-26
build: remove goerli support

andreivladbrg:
Remove goerli support in `foundry.toml` file and use sepolia as default chain in deployment workflows.


PaulRBerg (approved):
LGTM

PR #727, opened 2023-11-20
Update metadata before hook in cancel function

andreivladbrg:
The idea behind this is to prevent DOS attacks when the recipient is a malicious contract and they could force excessive gas costs on the sender.


PaulRBerg:
This PR addresses the following audit finding:

https://cantina.xyz/ai/6d6a47ff-5279-4502-8705-d837b159a31e/findings/c50d4d6c-e9fe-4bad-b05b-393b1eecbdb4

andreivladbrg:
@PaulRBerg does it look good to you?

PR #729, opened 2023-11-28
refactor: remove double r

PaulRBerg:
I noticed that the word `transferable` was written with a double `r` in some cases.


andreivladbrg (approved):
lgtm

PR #734, opened 2023-12-04
Install dependencies with Node.js instead of git modules

andreivladbrg:
Addresses https://github.com/sablier-labs/v2-core/discussions/508 and updates `prb-math` to `4.0.2`.

Since this project is not going to have a `lib` directory anymore, it no longer makes sense to keep the re-exported types. This is because if the project is installed with `forge install sablier-labs/v2-core` it won't have recursive modules to install, and node modules can't be installed. And if the project is installed with `yarn add @sablier/v2-core`, it would automatically install those deps (prb-math and openzeppelin).


PaulRBerg:
> should we temporarily remove this job and keep it only here https://github.com/sablier-labs/v2-core/pull/726?

Not gonna review this today anyway, so let's keep the job there for now.

Turns out that there's a bug in Foundry: https://github.com/foundry-rs/foundry/issues/6519

andreivladbrg:
I've made sure to mention that the recommended method to install `v2-core` project is via Node.js:

image

However, I realized that for integrators who still use `forge install`, they would also need to manually install `prb-math` and `openzeppelin`. This is because there are no longer recursive modules. Should we mention this in the `README` file?  @PaulRBerg

.github/workflows/ci-deep.yml
```diff
@@ -68,12 +68,24 @@ jobs:
...
+
+      - name: "Install Node.js"
+        uses: "actions/setup-node@v3"
+        with:
+          cache: "pnpm"
+          node-version: "lts/*"
+
+      - name: "Install the Node.js dependencies"
+        run: "pnpm install"
+
```
PaulRBerg, line 88: We need to find a way to shorten this syntax - it's too verbose.

Related: https://github.com/sablier-labs/v2-periphery/issues/152

.github/workflows/ci-deep.yml
```diff
@@ -91,7 +103,7 @@ jobs:
             cache
             out
             out-optimized
-
+
```
PaulRBerg, line 106: what's up with this space

.github/workflows/ci-deep.yml
```diff
@@ -152,7 +188,7 @@ jobs:
...
```
PaulRBerg, line 191: more spaces

.github/workflows/ci-deep.yml
```diff
@@ -185,7 +234,7 @@ jobs:
...
```
PaulRBerg, line 237: more spaces, wow

README.md
```diff
@@ -26,7 +26,15 @@ of tokens deposited.
...
-### Foundry
+### Node.js (recommended)
+
+Sablier V2 Core is available as a Node.js package:
+
+```shell
+yarn add @sablier/v2-core
+```
+
+### Foundry (slow)
```
PaulRBerg, line 37: Let's remove "slow" - can be misunderstood to mean that Foundry is slow. The "recommended" note above is sufficient.

package.json
```diff
@@ -46,7 +52,7 @@
     "web3"
   ],
   "peerDependencies": {
-    "@prb/math": "4.0.1"
+    "@prb/math": "4.0.2"
```
PaulRBerg, line 55: Shouldn't this be `4.x`? Or `4.0.x`?

Just thinking out loud

remappings.txt
```diff
@@ -1,6 +1,6 @@
...
-@prb/test/=lib/prb-test/src/
-forge-std/=lib/forge-std/src/
-solady/=lib/solady/src/
-solarray/=lib/solarray/src/
+@openzeppelin/contracts/=node_modules/@openzeppelin/contracts/
+@prb/math/=node_modules/@prb/math/
+@prb/test/=node_modules/@prb/test/
+forge-std/=node_modules/forge-std/
+solady/=node_modules/solady/
+solarray/=node_modules/solarray/
```
PaulRBerg, line 6: Will these work with pnpm node modules system, @andreivladbrg? See what I asked you [here}(https://github.com/sablier-labs/v2-core/discussions/508#discussioncomment-7751667)

Can you please perform some tests?

PaulRBerg (changes requested):
feedback below

PaulRBerg:
> it no longer makes sense to keep the re-exported types.

Those types have always been problematic, actually. See [the ](https://github.com/simplemachine92/JBSips/issues/10).

So it's good that we are getting rid of them.

> they would also need to manually install prb-math and openzeppelin. This is because there are no longer recursive modules. Should we mention this in the README file?

Yes, we should. Also, please test that it works properly!

.github/workflows/ci-deep.yml
andreivladbrg: dunno what's up with github, maybe a bug on them (similar to `.sol`) 

if you actually got to look at the file, it looks good

https://github.com/sablier-labs/v2-core/blob/96f89cfabea3dc6638f0b64a30cfb44066a95944/.github/workflows/ci.yml#L93-L107

.github/workflows/ci-deep.yml
andreivladbrg: answered here https://github.com/sablier-labs/v2-core/pull/734#discussion_r1427360409

.github/workflows/ci-deep.yml
andreivladbrg: answered here https://github.com/sablier-labs/v2-core/pull/734#discussion_r1427360409

remappings.txt
andreivladbrg: I think you got my point wrong, as also said in that discussion, pnpm places only the third party deps under the `node_modules/.pnpm`.

If you clone the repo, you can run `pnpm install` and its deps would be place under `node_modules/`, so we don't have to worry about the remappings for the projects itself. 

What I mean when I say that we should recommend `yarn add`  instead of `pnpm add` is when an integrator (as allo was) should add to their project our package.  For them, when adding our package, the third party deps are `openzeppelin/contracts/` and `@prb/math/`:

https://github.com/sablier-labs/v2-core/blob/96f89cfabea3dc6638f0b64a30cfb44066a95944/package.json#L13-L16

More explantions about CLIs: 
- `pnpm add` : Installs a package and any packages that it depends on. https://pnpm.io/cli/install
- `pnpm install` : is used to install all dependencies for a project. https://pnpm.io/cli/add


> Can you please perform some tests? 

Did several tests when I've written my answer [here](https://github.com/sablier-labs/v2-core/discussions/508#discussioncomment-7374482).

andreivladbrg:
> Those types have always been problematic, actually. See https://github.com/simplemachine92/JBSips/issues/10.

So it's good that we are getting rid of them.

great

> Yes, we should. Also, please test that it works properly!

I did. You can test it yourself using this [branch](https://github.com/sablier-labs/v2-periphery/tree/build/multi-chain-script) from periphery. Everything works there.

package.json
andreivladbrg: not sure what you mean:

image

.github/workflows/ci-deep.yml
andreivladbrg: I've managed to cache the `node_modules` dir and it seems to work expect the coverage, which i don't understand

PaulRBerg: I wasn't referring to the `run: "pnpm install"` step. I was referring to all of these separate steps - it would be great if we could have a shared job. See the linked issue.

> I've managed to cache the node_modules dir and re-use it

That's not what that `cache:` key does. It only caches global packages. See:

- https://github.com/actions/setup-node?tab=readme-ov-file#caching-global-packages-data
- https://github.com/actions/setup-node/issues/612

I will bring back that key

PaulRBerg: Oh nevermind I looked at the code now and I see what you did.

Great idea to cache the `node_modules`!

.github/workflows/ci-deep.yml
PaulRBerg: @andreivladbrg there's no bug in GitHub. There were actual extra spaces there that you added in this PR. I have removed them via my latest commit.

.github/workflows/ci.yml
```diff
@@ -296,8 +316,20 @@ jobs:
...
+
+      - name: "Install Node.js"
+        uses: "actions/setup-node@v3"
         with:
-          submodules: "recursive"
+          cache: "pnpm"
+          node-version: "lts/*"
+
+      - name: "Install the Node.js dependencies"
+        run: "pnpm install"
```
PaulRBerg, line 332: This step is now redundant given that we are caching `node_modules`.

package.json
PaulRBerg: There's a difference between `4.0.2` and `4.x` and `4.0.x` in the context of peer dependencies.

https://chat.openai.com/share/cc9100d2-7294-42bc-b579-a3ef5c79e194

Anyway, we can leave it as `4.0.2` for now; this was more so a reminder for me to have a look later on.

remappings.txt
PaulRBerg: > I think you got my point wrong

I didn't; I was just unclear in my question.

> If you clone the repo, you can run pnpm install and its deps would be place under node_modules/, so we don't have to worry about the remappings for the projects itself.

I know. I was referring to third-party developers installing V2 Core. Would the remappings still work recursively?

Sorry for not being clear earlier.

> What I mean when I say that we should recommend `yarn add` instead of `pnpm add` is when an integrator

This is equivalent to asking the integrators to switch over to Pnpm instead of Yarn. And, as I've asked you in the discussion: `Surely there must be a way to force Pnpm to not put the dependencies under `node_modules/.pnpm` without asking users to manually install those packages?`

> Did several tests when I've written my answer https://github.com/sablier-labs/v2-core/discussions/508#discussioncomment-7374482.

Thanks but you missed one of my questions. Let's follow up there.

PaulRBerg:
> I did. You can test it yourself using this [branch](https://github.com/sablier-labs/v2-periphery/tree/build/multi-chain-script) from periphery. Everything works there.

Fantastic

.github/workflows/ci.yml
andreivladbrg: the `slither-analyze` job is going to be removed from `ci.yml` and we will need the `pnpm install`in its own file, see #726

PaulRBerg: Fine. In this case, it doesn't really matter what we do here - can you keep my latest commit as is, please?

PR #748, opened 2023-12-17
build: multi-chain deployment script

PaulRBerg:
great job @andreivladbrg, I like the script. Let me know what you think about my changes.


foundry.toml
```diff
@@ -79,16 +79,6 @@
...
-[etherscan]
-  arbitrum = { key = "${API_KEY_ARBISCAN}" }
-  avalanche = { key = "${API_KEY_SNOWTRACE" }
-  bnb_smart_chain = { key = "${API_KEY_BSCSCAN}" }
-  gnosis_chain = { key = "${API_KEY_GNOSISSCAN}" }
-  mainnet = { key = "${API_KEY_ETHERSCAN}" }
-  optimism = { key = "${API_KEY_OPTIMISTIC_ETHERSCAN}" }
-  polygon = { key = "${API_KEY_POLYGONSCAN}" }
-  sepolia = { key = "${API_KEY_ETHERSCAN}" }
-
```
PaulRBerg, line 91: @andreivladbrg why did you remove these? Was it because of the env var name mismatch? If yes, I fixed this with the separation between the base env and the deployment env.

foundry.toml
```diff
@@ -100,7 +90,7 @@
   wrap_comments = true
 
 [rpc_endpoints]
-  arbitrum = "https://arbitrum-mainnet.infura.io/v3/${API_KEY_INFURA}"
+  arbitrum_one = "https://arbitrum-mainnet.infura.io/v3/${API_KEY_INFURA}"
```
smol-ninja, line 93: Shouldn't this be renamed to "arbitrum" as well?

PaulRBerg: Yep, it should. I will fix later today (have some staging changes locally.)

shell/deploy-multi-chain.sh
```diff
@@ -0,0 +1,232 @@
...
+    deployment_command="";
+
+    # Choose the script based on the flag
+    if [[ $DETERMINISTIC_DEPLOYMENT == true ]]; then
+        echo -e "\n${IC}Deploying deterministic contracts to $chain...${NC}"
+        # Construct the command
+        deployment_command="forge script script/DeployDeterministicCore3.s.sol \
+        --rpc-url $rpc_url \
+        --sig run(string,address,address,uint256) \
+        \'ChainID $chain_id, Version 1.1.1\' \
```
smol-ninja, line 178: This is not correct. It generates '\'\''ChainID' 11155111, Version '1.1.0\'\''' and now 'ChainID $chain_id, Version 1.1.1'

I've fixed it here https://github.com/sablier-labs/v2-core/pull/753

foundry.toml
andreivladbrg: They are problematic and I always end up in passing `--etherscan-api-key` flag, thus I decided to remove them and I still consider to be the better choice

PaulRBerg: I see. I agree that it would be helpful to remove the API keys from `foundry.toml`. They add maintenance cost.

> i find redundant and unnecessary to have 2 different .env files so i'd say to keep one

1. Not redundant because of security reasons; it's much much much safer not to load deployment keys all the time during testing and development
    - FYI: this separation between development and production env keys is super common in traditional software development
2. Reminder that necessary/ unnecessary is a bad criterion; helpful/ unhelpful is 1000x better

shell/deploy-multi-chain.sh
PaulRBerg: gotcha, feel free to push directly @smol-ninja.

foundry.toml
andreivladbrg: > Reminder that necessary/ unnecessary is a bad criterion; helpful/ unhelpful is 1000x better

just wording, reformulating then:

i find redundant and unhelpful to have 2 different .env files so i'd say to keep one

PaulRBerg:
It looks like `forge script` [supports a `--multi` parameter](https://book.getfoundry.sh/reference/cli/forge/script?highlight=multi%20chain%20deployment#forge-script):

> If present, `--resume` or `--verify` will be assumed to be a multi chain deployment

Are you aware of this, @smol-ninja, @andreivladbrg? Might be worth digging this up.

smol-ninja:
Yes. I have used it before. In order to use it, we will have to explicitly specify it in the script files. Since we will still need an option to select chains to deploy on (could be because of higher gas fee on one chain at the time of deployment), the script will be required to take list of chains as input. 

This works without using `--multi` flag as its part of the solidity script code. What `--multi` flag does is make `--verify` to work alongside multi chain deployment. 

IMO it does not add any additional benefit other than handing over the multi deployment logic to the script files.

```solidity
function run() {
    vm.createSelectFork(POLYGON_RPC_URL);
    vm.startBroadcast();
    
    vm.stopBroadcast();
    
     vm.createSelectFork(ETHEREUM_RPC_URL);
    vm.startBroadcast();
    
    vm.stopBroadcast();
}
```

.env.deployment.example
```diff
@@ -0,0 +1,42 @@
...
+export ARBITRUM_SEPOLIA_ADMIN="YOUR_ADMIN_ADDRESS"
+export AVALANCHE_ADMIN="YOUR_ADMIN_ADDRESS"
+export BASE_ADMIN="YOUR_ADMIN_ADDRESS"
+export BSC_ADMIN="YOUR_ADMIN_ADDRESS"
+export GNOSIS_ADMIN="YOUR_ADMIN_ADDRESS"
+export MAINNET_ADMIN="YOUR_ADMIN_ADDRESS"
+export OPTIMISM_ADMIN="YOUR_ADMIN_ADDRESS"
+export POLYGON_ADMIN="YOUR_ADMIN_ADDRESS"
+export SCROLL_ADMIN="YOUR_ADMIN_ADDRESS"
+export SEPOLIA_ADMIN="YOUR_ADMIN_ADDRESS"
```
PaulRBerg, line 42: We can keep these values directly in version control because the admin addresses are public:

https://docs.sablier.com/concepts/governance#multisig-addresses

PaulRBerg (commented):
LGTM except for the admin addresses that can be hard-coded directly in the script. I've pushed a commit myself.

Good to merge, but as I've said [here](https://github.com/sablier-labs/v2-core/pull/752#issuecomment-1892554259), we should do some more testing when we get to deploy V2.2.

PaulRBerg:
Feel free to accept the PR and merge it @smol-ninja @andreivladbrg.

smol-ninja (approved):

PaulRBerg:
@andreivladbrg - you have the honor to squash and merge.

PR #781, opened 2024-01-05
Add sender hook and rename recipient's hook functions

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/742 and https://github.com/sablier-labs/v2-core/issues/736

Depends on https://github.com/sablier-labs/v2-core/pull/779 to be merged on `staging` first.

Note: called `getSender` before the internal `_withdraw` function is called for security reason, in case the sender is changed when the assets are transffered.


src/abstracts/SablierV2Lockup.sol
```diff
@@ -277,14 +281,29 @@ abstract contract SablierV2Lockup is
             revert Errors.SablierV2Lockup_Overdraw(streamId, amount, withdrawableAmount);
         }
 
+        // Retrieve the sender from storage.
+        address sender = getSender(streamId);
```
smol-ninja, line 285: > `getSender` before the internal `_withdraw` function is called for security reason, in case the sender is changed when the assets are transfered

Do we allow changing sender after stream has been created?

-------
Instead of loading sender from storage 2 times: 
1. `bool isCallerStreamSender = _isCallerStreamSender(streamId);`
2. `address sender = getSender(streamId);`

we can just load it once and re-use it from stack. It would save 100 gas.

src/abstracts/SablierV2Lockup.sol
```diff
@@ -277,14 +281,29 @@ abstract contract SablierV2Lockup is
...
+                caller: msg.sender,
+                to: to,
+                amount: amount
+            }) { } catch { }
+        }
+
+        // Interactions: if `msg.sender` is not the sender and the sender is a contract, try to invoke the
+        // withdraw hook on it without reverting if the hook is not implemented, and also without bubbling up
+        // any potential revert.
+        if (!isCallerStreamSender && sender.code.length > 0) {
```
smol-ninja, line 305: Since `isCallerStreamSender` will no longer be needed after this [PR: make withdraw callable by any account](https://github.com/sablier-labs/v2-core/pull/785/files#diff-aa886876392e91852d49ed143fbd48c2fd40b518b994039251f0da369c3cbeb1L249), I suggest we should use `msg.sender != sender` here.

test/integration/concrete/lockup/withdraw/withdraw.t.sol
```diff
@@ -144,19 +145,153 @@ abstract contract Withdraw_Integration_Concrete_Test is Integration_Test, Withdr
...
+        whenCallerRecipient
     {
+        test_Withdraw_CallerRecipient(defaultStreamId, users.sender);
+    }
+
+    modifier givenSenderContract() {
+        _;
+    }
+
+    function test_Withdraw_SenderContract()
```
smol-ninja, line 157: this should be `test_Withdraw_SenderDoesNotImplementHook`, since sender is `noop` which does not implement hook. And there has already been a modifier `givenSenderContract` used.

smol-ninja (changes requested):

src/abstracts/SablierV2Lockup.sol
andreivladbrg: correct, i agree

test/integration/concrete/lockup/withdraw/withdraw.t.sol
andreivladbrg: correct, missed it

src/abstracts/SablierV2Lockup.sol
andreivladbrg: > Do we allow changing sender after stream has been created?

no, we don't, it was just a precaution in case some unwanted behaviour happens (better safe than sorry)

> we can just load it once and re-use it from stack. It would save 100 gas.

correct, but i suggest we leave it as it is, since, as you mentioned below, it is going to be removed anyway in your PR

smol-ninja: > it is just a precaution in case some unwanted behaviour happens

I understand your point and I think it's a good practice to take extra precautions. But I also think this is an overkill since there is no way to change sender after stream has been created. We don't have proxies so there is also no risk of overlapping storage slots. wdyt?

Nevertheless, in this case, we can leave it as it is because we, anyways, have to fetch `getSender(streamId)` after `_withdraw`.

> but i suggest we leave it as it is, since, as you mentioned below

OK. Agree.

smol-ninja (approved):
LGTM!

src/abstracts/SablierV2Lockup.sol
PaulRBerg: Am I correct that you ended undoing this behavior (i.e. the call to `getSender`)?

I agree with @smol-ninja that it's overkill to do this in production. The sender is unchangeable.

It might be worth adding a precautionary check like this in the tests, though.

smol-ninja: > Am I correct that you ended undoing this behavior (i.e. the call to getSender) in another PR

We no longer load sender from storage twice after #785 .

PaulRBerg: got it

PR #783, opened 2024-01-06
perf: remove unchecked in for loop

### LINKED ISSUE #757: Remove "unchecked" in for loop incrementors
Thanks to Solidity v0.8.22, we can say goodbye to the ugly "unchecked { ++i; }" syntax when using for loops.

Context:

- https://github.com/ethereum/solidity/releases/tag/v0.8.22
- https://twitter.com/PaulRBerg/status/1717148805282296017

smol-ninja:
Closes #757.

### DIFF
--- a/src/libraries/SVGElements.sol
+++ b/src/libraries/SVGElements.sol
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // solhint-disable max-line-length,quotes
-pragma solidity >=0.8.19;
+pragma solidity >=0.8.22;
 
 import { Math } from "@openzeppelin/contracts/utils/math/Math.sol";
 import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
--- a/src/libraries/SVGElements.sol
+++ b/src/libraries/SVGElements.sol
@@ -239,12 +239,11 @@ library SVGElements {
         unchecked {
             uint256 charWidth = largeFont ? 16 : 13;
             uint256 semicolonIndex;
-            for (uint256 i = 0; i < length;) {
+            for (uint256 i = 0; i < length; ++i) {
                 if (bytes(text)[i] == ";") {
                     semicolonIndex = i;
                 }
                 width += charWidth;
-                i += 1;
             }
 
             // Account for escaped characters (such as &#8805;).
--- a/test/utils/Fuzzers.sol
+++ b/test/utils/Fuzzers.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity >=0.8.19;
+pragma solidity >=0.8.22;
 
 import { PRBMathCastingUint128 as CastingUint128 } from "@prb/math/src/casting/Uint128.sol";
 import { UD60x18, ud, uUNIT } from "@prb/math/src/UD60x18.sol";
--- a/.github/workflows/multibuild.yml
+++ b/.github/workflows/multibuild.yml
@@ -21,6 +21,6 @@ jobs:
       - name: "Check that V2 Core can be built with multiple Solidity versions"
         uses: "PaulRBerg/foundry-multibuild@v1"
         with:
-          min: "0.8.19"
+          min: "0.8.22"
           max: "0.8.23"
           skip-test: "true"
--- a/.solhint.json
+++ b/.solhint.json
@@ -3,7 +3,7 @@
   "rules": {
     "avoid-low-level-calls": "off",
     "code-complexity": ["error", 9],
-    "compiler-version": ["error", ">=0.8.19"],
+    "compiler-version": ["error", ">=0.8.22"],
     "contract-name-camelcase": "off",
     "const-name-snakecase": "off",
     "custom-errors": "off",
--- a/script/Base.s.sol
+++ b/script/Base.s.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
-pragma solidity >=0.8.19 <0.9.0;
+pragma solidity >=0.8.22 <0.9.0;
 
 import { Script } from "forge-std/src/Script.sol";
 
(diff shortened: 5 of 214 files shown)


andreivladbrg (changes requested):
The pragma solidity version must updated to: `pragma solidity >=0.8.22;`

andreivladbrg:
Don't mind the `test-utils` fail, let's not update the [precompiles](https://github.com/sablier-labs/v2-core/blob/a2fe652380b6eadbff17e9791a75c12cbc2e6531/test/utils/Precompiles.sol#L27-L34) on each PR, to prevent git conflicts.

test/utils/Fuzzers.sol
```diff
@@ -87,12 +87,10 @@ abstract contract Fuzzers is Constants, Utils {
         uint128 estimatedDepositAmount = segments[0].amount;
 
         // Fuzz the other segment amounts by bounding from 0.
-        unchecked {
```
andreivladbrg, line 90: this shouldn't have been removed

test/utils/Fuzzers.sol
```diff
@@ -117,16 +115,14 @@ abstract contract Fuzzers is Constants, Utils {
 
     /// @dev Fuzzes the deltas.
     function fuzzSegmentDeltas(LockupDynamic.SegmentWithDelta[] memory segments) internal pure {
-        unchecked {
```
andreivladbrg, line 120: this shouldn't have been removed

src/libraries/SVGElements.sol
```diff
@@ -236,16 +236,18 @@ library SVGElements {
...
-                if (bytes(text)[i] == ";") {
-                    semicolonIndex = i;
-                }
+        uint256 charWidth = largeFont ? 16 : 13;
+        uint256 semicolonIndex;
+        for (uint256 i = 0; i < length; ++i) {
+            if (bytes(text)[i] == ";") {
+                semicolonIndex = i;
+            }
+            unchecked {
```
andreivladbrg, line 245: why did you make this change? it should have remained the same

andreivladbrg (commented):
same for the `gas-snapshot`, we shouldn't updated it on each PR to prevent conflicts

i think you can remove your last commit, you can use these commands:

```bash
git reset --hard HEAD~1
git push -f
```

test/utils/Fuzzers.sol
smol-ninja: Thanks for the comment, Andrei. I considered not removing it but I couldn't understand why we have `unchecked` in this place. This is the code required for testing so what is the rationale behind optimising for gas in here?

src/libraries/SVGElements.sol
smol-ninja: In the original code, the `unchecked` block encompassed the arithmetic operations, ternary statement and the loop that checks for the semicolon character in the text. It might imply that the `unchecked` keyword is also optimizing the other expressions which are not arithmetics, which is not the case anymore. 

So by moving it closer to the arithmetic operations, my goal was to make it clear that it only applies to these operations and does not affect other operations.

Even though this change resulted in two `unchecked` statements instead of one. However, IMO the increased clarity and explicitness of the code outweighs the minor redundancy introduced by the additional `unchecked` statement.

What do you think?

test/utils/Fuzzers.sol
smol-ninja: Same comment as above.

test/utils/Fuzzers.sol
andreivladbrg: i remember making the same comment back when this file was initially written. @PaulRBerg, you pointed out that using the `unchecked` block would also slightly decrease compilation time. do i remember that correctly, Paul? and is that still the case?

src/libraries/SVGElements.sol
andreivladbrg: > keyword is also optimizing the other expressions which are not arithmetics, which is not the case anymore

i am not sure i understand what you mean, are you referring to the if statement? or to the loop counter?

> my goal was to make it clear that it only applies to these operations and does not affect other operations.

what other operations are "affected"?

the loop counters, are "affected" by default by an `unchecked` block

> IMO the increased clarity and explicitness of the code outweighs the minor redundancy introduced by the additional unchecked statement

i see. personally, i believe that having multiple `unchecked` blocks decreases clarity when reading the code, as it results in the code being nested within two blocks instead of one. imo, the redundancy outweighs the benefits.

smol-ninja: The point that I wanted to make is `unchecked` only impacts `width += charWidth` and `width -= charWidth * semicolonIndex` and not every expression that it covers. That's why I changed this code so it appears to be covering only these two expressions. 

> i believe that having multiple unchecked blocks decreases clarity when reading the code

I agree with you now. Having two unchecked statements in such a short block goes against DRY principle. I will update it now.

test/utils/Fuzzers.sol
PaulRBerg: it's not about compilation times - but about run times.

Looping over fuzzed arrays is a computationally heavy operation. I haven't benchmarked this, but I'm pretty sure that the `unchecked` blocks help in this circumstance - it's basically ~100 gas saved per loop, which can add to a quite high consumption for large fuzzing campaigns (e.g., in the CI Deep workflow).

src/libraries/SVGElements.sol
PaulRBerg: Agree with @andreivladbrg.

PaulRBerg (changes requested):
Generally LGTM except for that change in `Fuzzerz` that should be undone (as per @andreivladbrg's feedback above).

It is also worth noting that this PR introduces a breaking change, i.e., changing the compiler version pragma from `>=0.8.19` to `>=0.8.22`.

test/utils/Fuzzers.sol
smol-ninja: Got it. Thanks for explaining Paul. I will revert them now.

PaulRBerg:
Is this PR the only reason why we changed the pragma to `>=0.8.22` in Sablier V2.2?

smol-ninja:
Yes.

PR #784, opened 2024-01-06
refactor: consistency in calldata related struct parameters

### LINKED ISSUE #728: Inconsistency in the order of parameters in structs between LockupLinear and LockupDynamic
We should have a consistent order of parameter structs:

- Sender
- Recipient
- Total Amount
- Asset
- Cancelable
- Transferable
- Time-related variables
- Broker

LD:

https://github.com/sablier-labs/v2-core/blob/e1a53b4c839912573996c313d447aeeed6eb434c/src/types/DataTypes.sol#L111-L121

LL:

https://github.com/sablier-labs/v2-core/blob/e1a53b4c839912573996c313d447aeeed6eb434c/src/types/DataTypes.sol#L223-L232

smol-ninja:
Closes #728.

### DIFF
--- a/src/SablierV2LockupDynamic.sol
+++ b/src/SablierV2LockupDynamic.sol
@@ -273,15 +273,15 @@ contract SablierV2LockupDynamic is
         // Checks, Effects and Interactions: create the stream.
         streamId = _createWithMilestones(
             LockupDynamic.CreateWithMilestones({
+                sender: params.sender,
+                recipient: params.recipient,
+                totalAmount: params.totalAmount,
                 asset: params.asset,
-                broker: params.broker,
                 cancelable: params.cancelable,
                 transferable: params.transferable,
-                recipient: params.recipient,
-                segments: segments,
-                sender: params.sender,
                 startTime: uint40(block.timestamp),
-                totalAmount: params.totalAmount
+                segments: segments,
+                broker: params.broker
             })
         );
     }
--- a/src/SablierV2LockupLinear.sol
+++ b/src/SablierV2LockupLinear.sol
@@ -264,14 +264,14 @@ contract SablierV2LockupLinear is
         // Checks, Effects and Interactions: create the stream.
         streamId = _createWithRange(
             LockupLinear.CreateWithRange({
+                sender: params.sender,
+                recipient: params.recipient,
+                totalAmount: params.totalAmount,
                 asset: params.asset,
-                broker: params.broker,
                 cancelable: params.cancelable,
                 transferable: params.transferable,
                 range: range,
-                recipient: params.recipient,
-                sender: params.sender,
-                totalAmount: params.totalAmount
+                broker: params.broker
             })
         );
     }
--- a/test/fork/LockupDynamic.t.sol
+++ b/test/fork/LockupDynamic.t.sol
@@ -34,15 +34,15 @@ abstract contract LockupDynamic_Fork_Test is Fork_Test {
     //////////////////////////////////////////////////////////////////////////*/
 
     struct Params {
-        Broker broker;
-        UD60x18 protocolFee;
-        address recipient;
         address sender;
+        address recipient;
+        uint128 withdrawAmount;
+        bool transferable;
+        UD60x18 protocolFee;
         uint40 startTime;
         uint40 warpTimestamp;
         LockupDynamic.Segment[] segments;
-        uint128 withdrawAmount;
-        bool transferable;
+        Broker broker;
     }
 
     struct Vars {
@@ -179,15 +179,15 @@ abstract contract LockupDynamic_Fork_Test is Fork_Test {
         // Create the stream.
         lockupDynamic.createWithMilestones(
             LockupDynamic.CreateWithMilestones({
+                sender: params.sender,
+                recipient: params.recipient,
+                totalAmount: vars.totalAmount,
                 asset: ASSET,
-                broker: params.broker,
                 cancelable: true,
                 transferable: params.transferable,
-                recipient: params.recipient,
-                segments: params.segments,
-                sender: params.sender,
                 startTime: params.startTime,
-                totalAmount: vars.totalAmount
+                segments: params.segments,
+                broker: params.broker
             })
         );
 
--- a/test/fork/LockupLinear.t.sol
+++ b/test/fork/LockupLinear.t.sol
@@ -34,15 +34,15 @@ abstract contract LockupLinear_Fork_Test is Fork_Test {
     //////////////////////////////////////////////////////////////////////////*/
 
     struct Params {
-        Broker broker;
-        UD60x18 protocolFee;
-        LockupLinear.Range range;
-        address recipient;
         address sender;
+        address recipient;
         uint128 totalAmount;
-        uint40 warpTimestamp;
         uint128 withdrawAmount;
         bool transferable;
+        UD60x18 protocolFee;
+        uint40 warpTimestamp;
+        LockupLinear.Range range;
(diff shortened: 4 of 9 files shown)


src/SablierV2LockupDynamic.sol
```diff
@@ -273,15 +273,15 @@ contract SablierV2LockupDynamic is
         // Checks, Effects and Interactions: create the stream.
         streamId = _createWithMilestones(
             LockupDynamic.CreateWithMilestones({
+                sender: params.sender,
```
andreivladbrg, line 276: the order here should have remained the same, it was alphabetically ordered

src/SablierV2LockupLinear.sol
```diff
@@ -264,14 +264,14 @@ contract SablierV2LockupLinear is
         // Checks, Effects and Interactions: create the stream.
         streamId = _createWithRange(
             LockupLinear.CreateWithRange({
+                sender: params.sender,
+                recipient: params.recipient,
+                totalAmount: params.totalAmount,
```
andreivladbrg, line 269: the order here should have remained the same, it was alphabetically ordered

test/fork/LockupDynamic.t.sol
```diff
@@ -179,15 +179,15 @@ abstract contract LockupDynamic_Fork_Test is Fork_Test {
...
-                broker: params.broker,
                 cancelable: true,
                 transferable: params.transferable,
-                recipient: params.recipient,
-                segments: params.segments,
-                sender: params.sender,
                 startTime: params.startTime,
-                totalAmount: vars.totalAmount
+                segments: params.segments,
+                broker: params.broker
```
andreivladbrg, line 190: it was alphabetically ordered

test/fork/LockupLinear.t.sol
```diff
@@ -181,14 +181,14 @@ abstract contract LockupLinear_Fork_Test is Fork_Test {
...
+                totalAmount: params.totalAmount,
                 asset: ASSET,
-                broker: params.broker,
                 cancelable: true,
                 transferable: params.transferable,
                 range: params.range,
-                recipient: params.recipient,
-                sender: params.sender,
-                totalAmount: params.totalAmount
+                broker: params.broker
```
andreivladbrg, line 191: it was alphabetically ordered

test/integration/fuzz/lockup-linear/createWithRange.t.sol
```diff
@@ -198,14 +198,14 @@ contract CreateWithRange_LockupLinear_Integration_Fuzz_Test is
...
                 asset: dai,
-                broker: params.broker,
                 cancelable: params.cancelable,
-                range: params.range,
-                recipient: params.recipient,
                 transferable: params.transferable,
-                sender: params.sender,
-                totalAmount: params.totalAmount
+                range: params.range,
+                broker: params.broker
```
andreivladbrg, line 208: it was alphabetically ordered

test/utils/Defaults.sol
```diff
@@ -184,54 +184,54 @@ contract Defaults is Constants {
...
+            totalAmount: TOTAL_AMOUNT,
             asset: asset,
-            broker: broker(),
             cancelable: true,
             transferable: true,
-            recipient: users.recipient,
             segments: segmentsWithDeltas(),
-            sender: users.sender,
-            totalAmount: TOTAL_AMOUNT
+            broker: broker()
```
andreivladbrg, line 194: it was alphabetically ordered

test/utils/Defaults.sol
```diff
@@ -184,54 +184,54 @@ contract Defaults is Constants {
...
-            broker: broker(),
             cancelable: true,
             transferable: true,
-            recipient: users.recipient,
-            segments: segments(),
-            sender: users.sender,
             startTime: START_TIME,
-            totalAmount: TOTAL_AMOUNT
+            segments: segments(),
+            broker: broker()
```
andreivladbrg, line 221: it was alphabetically ordered

andreivladbrg (changes requested):
i think you've missed the best practice for how we pass arguments when calling a function with a struct as a param

thus, only the order from `src/types/DataTypes.sol` should have been changed

smol-ninja:
Thanks, @andreivladbrg for the feedback. Please find my comments below.

> I think you've missed the best practice for how we pass arguments when calling a function with a struct as a param

I was aware that we generally follow an alphabetically ordered pattern. So I looked at the [Sablier guidelines](https://sablier.notion.site/Sablier-Coding-Practices-Contracts-2f5920cf497646b4a2ea982c1bacb5a8) before making these changes. However, in the original code, it does not appear to be alphabetically ordered (please see below and correct me if I misunderstood it - a b c t r s s s t).

```solidity
 streamId = _createWithMilestones(
            LockupDynamic.CreateWithMilestones({
                asset: params.asset,
                broker: params.broker,
                cancelable: params.cancelable,
                transferable: params.transferable,
                recipient: params.recipient,
                segments: segments,
                sender: params.sender,
                startTime: uint40(block.timestamp),
                totalAmount: params.totalAmount
            })
        );
```

Now my rationale behind making these changes was **code consistency**. By following the same order, as declaring structs in the `DataTypes.sol`, throughout the codebase, it appears to make code easier to understand at a glance and reduce the likelihood of missing out on any struct variable since they are now in the same order.

However, I acknowledge your point that we should adhere to best practices for function parameters. If the alphabetical ordering of the arguments is a standard that we want to maintain, I am willing to revert the changes and keep it alphabetically ordered in the function.

andreivladbrg:
Thanks @smol-ninja for your detailed response

> in the original code, it does not appear to be alphabetically ordered

indeed, it was not alphabetically ordered, it was a mistake on our end, `transferable` should have been the last one

> Now my rationale behind making these changes was code consistency. By following the same order, as declaring structs in the DataTypes.sol, throughout the codebase, it appears to make code easier to understand at a glance and reduce the likelihood of missing out on any struct variable since they are now in the same order.

i see, i understand your rationale, and tbh, i'm starting to like it and am leaning towards it as being the better choice. @PaulRBerg i'm curious about your opinion on this

smol-ninja:
@PaulRBerg we'd love your opinion on this.

PaulRBerg:
This is an exemplary case of **conflicting heuristics** (which are inevitable in life).

- On the one hand, we want to order the struct fields in a non-alphabetic way to optimize the gas consumption.
- On the other hand, we want to order the inputs alphabetically in the tests (to adhere to our best practice).

So the dilemma is: should we not apply the alphabetical order heuristic in this case (the tests)?

While I can't think of any decisive argument, I'm slightly more inclined to go with the non-alphabetic approach (the one that @smol-ninja wishes to implement). The reason for this is that it would (ironically) create more consistency among our create function calls - across the production code in `src`, the test code in `test`, and our [examples](https://github.com/sablier-labs/examples).

andreivladbrg:
> I'm slightly more inclined to go with the non-alphabetic approach (the one that @smol-ninja wishes to implement). The reason for this is that it would (ironically) create more consistency among our create function calls - across the production code in src, the test code in test, and our [examples](https://github.com/sablier-labs/examples).

yeah, completely relatable

smol-ninja:
> On the one hand, we want to order the struct fields in a non-alphabetic way to optimize the gas consumption

As discussed [here](https://github.com/sablier-labs/v2-core/issues/728#issuecomment-1877065940), the order of parameters does not affect gas usage in this case.

However, even then, I completely agree that it creates more consistency. Since, we all agree on this, I will merge this PR.

PR #786, opened 2024-01-09
chore: remove etherscan API keys

PaulRBerg:
As realized in https://github.com/sablier-labs/v2-core/pull/748, keeping the Etherscan API keys in version control is not worth the maintenance cost.

We only need these keys at deployment time.


andreivladbrg (approved):
Are we planning to close that PR? Trying to understand why we have separate PR for this specific thing

.env.example
```diff
@@ -1,11 +1,4 @@
-export API_KEY_ARBISCAN="YOUR_API_KEY_ARBISCAN"
-export API_KEY_BSCSCAN="YOUR_API_KEY_BSCSCAN"
-export API_KEY_ETHERSCAN="YOUR_API_KEY_ETHERSCAN"
-export API_KEY_GNOSISSCAN="YOUR_API_KEY_GNOSISSCAN"
 export API_KEY_INFURA="YOUR_API_KEY_INFURA"
-export API_KEY_OPTIMISTIC_ETHERSCAN="YOUR_API_KEY_OPTIMISTIC_ETHERSCAN"
-export API_KEY_POLYGONSCAN="YOUR_API_KEY_POLYGONSCAN"
-export API_KEY_SNOWTRACE="YOUR_API_KEY_SNOWTRACE"
 export RPC_URL_MAINNET="YOUR_RPC_URL_MAINNET"
```
smol-ninja, line 2: I think we can also remove `RPC_URL_MAINNET` and just keep the `API_KEY_INFURA`. Then in [foundry.toml](https://github.com/sablier-labs/v2-core/blob/54ba6f5298eeb8f21ed8c66d5c1ff71f4612eb19/foundry.toml#L97), we can use `mainnet = "https://mainnet.infura.io/v3/${API_KEY_INFURA}"`. 

Is there a different reason why do we have separate `RPC_URL_MAINNET`?

smol-ninja (changes requested):

PaulRBerg:
> Are we planning to close https://github.com/sablier-labs/v2-core/pull/748?

Eventually, yes. Just never found the time to review it.

> Trying to understand why we have separate PR for this specific thing

I need it for another PR.

.env.example
PaulRBerg: We need `RPC_URL_MAINNET` for fork testing purposes because our Infura developer plan doesn't offer an archive node.

smol-ninja (approved):

PR #791, opened 2024-01-12
refactor: withdraw related notes

smol-ninja:


src/interfaces/ISablierV2Lockup.sol
```diff
@@ -244,11 +244,14 @@ interface ISablierV2Lockup is
...
     /// and `msg.sender` is not the recipient.
+    /// - This function attempts to invoke a hook on the stream's sender, provided that the sender is a contract
+    /// and `msg.sender` is not the sender.
     ///
     /// Requirements:
     /// - Must not be delegate called.
     /// - `streamId` must not reference a null or depleted stream.
-    /// - `to` must be the recipient if `msg.sender` is not the stream's recipient or an approved third party.
+    /// - if `msg.sender` is not the stream's recipient, nor an approved third party, the `to`
+    /// must be the recipient.
```
andreivladbrg, line 254: let's undo this change

src/interfaces/ISablierV2Lockup.sol
```diff
@@ -294,7 +297,7 @@ interface ISablierV2Lockup is
     /// @dev Emits multiple {Transfer}, {WithdrawFromLockupStream}, and {MetadataUpdate} events.
     ///
     /// Notes:
-    /// - This function attempts to call a hook on the recipient of each stream, unless `msg.sender` is the recipient.
+    /// - Refer to the notes in {withdraw}.
```
andreivladbrg, line 300: Let's say:
```
/// - This function attempts to call a hook on the recipient of each stream, unless `msg.sender` is the recipient.
/// - This function attempts to call a hook on the sender of each stream, unless `msg.sender` is the sender.
```

andreivladbrg (changes requested):

andreivladbrg (approved):

PR #796, opened 2024-01-16
Make NFT descriptor public

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/763

### DIFF
--- a/src/abstracts/SablierV2Lockup.sol
+++ b/src/abstracts/SablierV2Lockup.sol
@@ -29,12 +29,8 @@ abstract contract SablierV2Lockup is
     /// @inheritdoc ISablierV2Lockup
     uint256 public override nextStreamId;
 
-    /*//////////////////////////////////////////////////////////////////////////
-                                  INTERNAL STORAGE
-    //////////////////////////////////////////////////////////////////////////*/
-
-    /// @dev Contract that generates the non-fungible token URI.
-    ISablierV2NFTDescriptor internal _nftDescriptor;
+    /// @inheritdoc ISablierV2Lockup
+    ISablierV2NFTDescriptor public override nftDescriptor;
 
     /*//////////////////////////////////////////////////////////////////////////
                                      CONSTRUCTOR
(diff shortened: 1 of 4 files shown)


src/abstracts/SablierV2Lockup.sol
```diff
@@ -29,12 +29,8 @@ abstract contract SablierV2Lockup is
     /// @inheritdoc ISablierV2Lockup
     uint256 public override nextStreamId;
 
-    /*//////////////////////////////////////////////////////////////////////////
-                                  INTERNAL STORAGE
-    //////////////////////////////////////////////////////////////////////////*/
-
     /// @dev Contract that generates the non-fungible token URI.
```
smol-ninja, line 32: It should be 
```soldiity
/// @inheritdoc ISablierV2Lockup
```

src/interfaces/ISablierV2Lockup.sol
```diff
@@ -134,6 +134,9 @@ interface ISablierV2Lockup is
     /// @notice Counter for stream ids, used in the create functions.
     function nextStreamId() external view returns (uint256);
 
+    /// @notice Contract that generates the non-fungible token URI.
```
smol-ninja, line 137: How about this?

```solidity
/// @notice Retrieves the address of NFT descriptor contract that generates the non-fungible token URI.
```

smol-ninja (changes requested):

src/abstracts/SablierV2Lockup.sol
andreivladbrg: correct

src/interfaces/ISablierV2Lockup.sol
andreivladbrg: it is better like it is now

we usually use "Retrieves" word for getters that implement the function within a scope. here it is not the case, as it is implemented automatically by the language when:

```solidity
ISablierV2NFTDescriptor public override nftDescriptor;
```

PaulRBerg (approved):
lgtm

PR #798, opened 2024-01-21
Rename create functions

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/797

### DIFF
--- a/test/integration/shared/lockup-dynamic/LockupDynamic.t.sol
+++ b/test/integration/shared/lockup-dynamic/LockupDynamic.t.sol
@@ -39,10 +39,10 @@ abstract contract LockupDynamic_Integration_Shared_Test is Lockup_Integration_Sh
         _params.createWithTimestamps.broker = defaults.broker();
 
         // See https://github.com/ethereum/solidity/issues/12783
-        LockupDynamic.SegmentWithDelta[] memory segmentsWithDeltas = defaults.segmentsWithDeltas();
+        LockupDynamic.SegmentWithDuration[] memory segmentsWithDurations = defaults.segmentsWithDurations();
         LockupDynamic.Segment[] memory segments = defaults.segments();
         for (uint256 i = 0; i < defaults.SEGMENT_COUNT(); ++i) {
-            _params.createWithDurations.segments.push(segmentsWithDeltas[i]);
+            _params.createWithDurations.segments.push(segmentsWithDurations[i]);
             _params.createWithTimestamps.segments.push(segments[i]);
         }
     }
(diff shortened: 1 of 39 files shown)


test/integration/shared/lockup-linear/LockupLinear.t.sol
```diff
@@ -60,60 +60,60 @@ abstract contract LockupLinear_Integration_Shared_Test is Lockup_Integration_Sha
...
+        LockupLinear.CreateWithTimestamps memory params = _params.createWithTimestamps;
         params.range.end = endTime;
-        streamId = lockupLinear.createWithRange(params);
+        streamId = lockupLinear.createWithTimestamps(params);
     }
 
-    /// @dev Creates the default stream with the provided createWithRange.
-    function createDefaultStreamWithRange(LockupLinear.Range memory createWithRange)
+    /// @dev Creates the default stream with the provided createWithTimestamps.
+    function createDefaultStreamWithRange(LockupLinear.Range memory createWithTimestamps)
```
smol-ninja, line 83: This appears to be inconsistent with the datatype used here.

Instead of `LockupLinear.Range memory createWithTimestamps`, how about `LockupLinear.Range memory range`?

On line 88, we can say `params.range = range`.

test/integration/shared/lockup-dynamic/LockupDynamic.t.sol
```diff
@@ -21,130 +21,130 @@ abstract contract LockupDynamic_Integration_Shared_Test is Lockup_Integration_Sh
...
     function createDefaultStreamWithBroker(Broker memory broker) internal override returns (uint256 streamId) {
-        LockupDynamic.CreateWithMilestones memory params = _params.createWithMilestones;
+        LockupDynamic.CreateWithTimestamps memory params = _params.createWithTimestamps;
         params.broker = broker;
-        streamId = lockupDynamic.createWithMilestones(params);
+        streamId = lockupDynamic.createWithTimestamps(params);
     }
 
     /// @dev Creates the default stream with deltas.
     function createDefaultStreamWithDeltas() internal returns (uint256 streamId) {
```
smol-ninja, line 70: Wouldn't it sound better to be called function `createDefaultStreamWithDurations`? This function calls function `lockupDynamic.createWithDurations` with `_params.createWithDurations` so the implementation does not appear to be consistent with the function name.

smol-ninja (changes requested):

test/integration/shared/lockup-linear/LockupLinear.t.sol
PaulRBerg: Yeah, this is indeed a mistake. And it was a mistake even before this PR. Good catch.

andreivladbrg:
@smol-ninja as said [here](https://github.com/sablier-labs/v2-core/issues/797#issuecomment-1903864668)

i have also renamed milestone to timestamp and delta to duration

lmk if i missed something

smol-ninja (approved):
looks good to me now.

andreivladbrg:
> Looks good to me now. But we should merge https://github.com/sablier-labs/v2-core/pull/785 first, right?

yeah, rebased now, after the ci will merge this

PR #803, opened 2024-01-24
Merge headers

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/801

IMO it looks much cleaner now


src/abstracts/SablierV2Base.sol
```diff
@@ -22,18 +22,14 @@ abstract contract SablierV2Base is
     using SafeERC20 for IERC20;
 
     /*//////////////////////////////////////////////////////////////////////////
-                                  PUBLIC CONSTANTS
+                                  STATE VARIABLES
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierV2Base
```
smol-ninja, line 28: I think we should keep this order:
1. Constants and immutable
2. Mutable variables

The order of mutable variables plays a key role in storage layout so we should keep them together. Technically speaking, inserting constants in between mutable variables does not have any impact as long as the order of mutable variables is optimized for storage. But from a readability point of view, I think having immutable variables followed by mutable variables looks good.

test/invariant/handlers/BaseHandler.sol
```diff
@@ -12,25 +12,21 @@ import { TimestampStore } from "../stores/TimestampStore.sol";
...
-    /// @dev The virtual address of the Foundry VM.
-    address internal constant VM_ADDRESS = address(uint160(uint256(keccak256("hevm cheat code"))));
-
-    /*//////////////////////////////////////////////////////////////////////////
-                                     VARIABLES
+                                    STATE-VARIABLES
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Maps function names to the number of times they have been called.
     mapping(string func => uint256 calls) public calls;
```
smol-ninja, line 19: same as above.

smol-ninja (changes requested):

src/abstracts/SablierV2Base.sol
andreivladbrg: > But from a readability point of view,  I think having immutable variables followed by mutable variables looks good.

I agree.

smol-ninja (approved):
I am good going forward with this approach.

andreivladbrg:
@PaulRBerg both of us like this change, wdyt?

PaulRBerg:
I'm against this PR as it currently is implemented.

My brain is hurt when I see "State Variables" and right below that, there are constants.

A constant is the opposite of a variable.

Can we please introduce a new category, "CONSTANTS" (and include immutables under that)?

PaulRBerg (changes requested):
I'm against this PR as it currently is implemented.

My brain is hurt when I see "State Variables" and right below that, there are constants.

A constant is the opposite of a variable.

Can we please introduce a new category, "CONSTANTS" (and include immutables under that)?

andreivladbrg:
> My brain is hurt when I see "State Variables" and right below that, there are constants.

even the Solidity docs refers to constants as state variables: https://docs.soliditylang.org/en/v0.8.23/contracts.html#constant-and-immutable-state-variables

I really don't see the problem. Have you read my message from [here](https://github.com/sablier-labs/v2-core/discussions/800#discussioncomment-8226258)?

State(for contracts) = code(from bytecode) + storage

constants are included in the bytecode, thus it makes to refer to them as "state"

> Can we please introduce a new category, "CONSTANTS" (and include immutables under that)?

would it be better if we change to the initial version i suggested? it removes the "variable" word which is problematic _for you_

```solidity
                                 CONTRACT STATE
```

PaulRBerg:
I did a little research, and you're right @andreivladbrg. The community refers to constants as state variables:

https://www.oreilly.com/library/view/mastering-blockchain-programming/9781839218262/7f1861df-c39b-4e76-b177-9ec0ddb10bfe.xhtml

Let's keep "STATE VARIABLES" and merge the PR as is.

PaulRBerg (approved):
lgtm now

andreivladbrg:
great 🙌

PR #814, opened 2024-02-05
feat: add a util function to construct the create2 salt

andreivladbrg:
Addresses https://github.com/sablier-labs/v2-core/discussions/810

I decided to log the string salt to visually see what is being passed (also testing it, in case something is not working).

@smol-ninja, can you check, please, if I missed anything in the multi-chain script?

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -71,6 +71,7 @@
 
 # Test the optimized contracts without re-compiling them
 [profile.test-optimized]
+  ffi = true
   src = "test"
 
 [doc]
--- a/script/Base.s.sol
+++ b/script/Base.s.sol
@@ -1,4 +1,5 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
+// solhint-disable no-console
 pragma solidity >=0.8.19 <0.9.0;
 
 import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
--- a/script/Base.s.sol
+++ b/script/Base.s.sol
@@ -1,9 +1,14 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity >=0.8.19 <0.9.0;
 
+import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
+
+import { console2 } from "forge-std/src/console2.sol";
 import { Script } from "forge-std/src/Script.sol";
 
 abstract contract BaseScript is Script {
+    using Strings for uint256;
+
     /// @dev Included to enable compilation of the script without a $MNEMONIC environment variable.
     string internal constant TEST_MNEMONIC = "test test test test test test test test test test test junk";
 
@@ -37,5 +42,26 @@ abstract contract BaseScript is Script {
         vm.startBroadcast(broadcaster);
         _;
         vm.stopBroadcast();
+    }
+
+    /// @dev The presence of the salt instructs Forge to deploy contracts via this deterministic CREATE2 factory:
+    /// https://github.com/Arachnid/deterministic-deployment-proxy
+    ///
+    /// Notes:
+    /// - The salt format is "ChainID <chainid>, Version <version>".
+    /// - The version is obtained from the `package.json` file via --ffi cheatcode:
+    /// https://book.getfoundry.sh/cheatcodes/ffi
+    function _constructCreate2Salt() internal returns (bytes32) {
+        string memory chainId = block.chainid.toString();
+        string[] memory inputs = new string[](4);
+        inputs[0] = "jq";
+        inputs[1] = "-r";
+        inputs[2] = ".version";
+        inputs[3] = "./package.json";
+        bytes memory result = vm.ffi(inputs);
+        string memory version = string(result);
+        string memory create2Salt = string.concat("ChainID ", chainId, ", Version ", version);
+        console2.log(create2Salt);
+        return bytes32(abi.encodePacked(create2Salt));
     }
 }
--- a/script/Base.s.sol
+++ b/script/Base.s.sol
@@ -7,7 +7,7 @@ import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
 import { console2 } from "forge-std/src/console2.sol";
 import { Script } from "forge-std/src/Script.sol";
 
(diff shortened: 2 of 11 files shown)


smol-ninja (dismissed):
LGTM! I haven't tested the scripts locally though.

PaulRBerg:
Looks like the fork test fails are due to QuickNode?

SCR-20240205-qebb

script/Base.s.sol
```diff
@@ -38,4 +44,25 @@ abstract contract BaseScript is Script {
...
         vm.stopBroadcast();
     }
+
+    /// @dev The presence of the salt instructs Forge to deploy contracts via this deterministic CREATE2 factory:
+    /// https://github.com/Arachnid/deterministic-deployment-proxy
+    ///
+    /// Notes:
+    /// - The salt format is "ChainID <chainid>, Version <version>".
+    /// - The version is obtained from the `package.json` file via --ffi cheatcode:
+    /// https://book.getfoundry.sh/cheatcodes/ffi
```
PaulRBerg, line 54: Let's add another note before this that says that the user needs to have the `jq` CLI tool installed.

script/Base.s.sol
```diff
@@ -38,4 +44,25 @@ abstract contract BaseScript is Script {
...
+        string memory chainId = block.chainid.toString();
+        string[] memory inputs = new string[](4);
+        inputs[0] = "jq";
+        inputs[1] = "-r";
+        inputs[2] = ".version";
+        inputs[3] = "./package.json";
+        bytes memory result = vm.ffi(inputs);
+        string memory version = string(result);
+        string memory create2Salt = string.concat("ChainID ", chainId, ", Version ", version);
+        console2.log(create2Salt);
```
PaulRBerg, line 65:
```suggestion
console2.log("The CREATE2 salt is \"%s\"", create2Salt);
```

script/Base.s.sol
```diff
@@ -38,4 +44,25 @@ abstract contract BaseScript is Script {
...
     }
+
+    /// @dev The presence of the salt instructs Forge to deploy contracts via this deterministic CREATE2 factory:
+    /// https://github.com/Arachnid/deterministic-deployment-proxy
+    ///
+    /// Notes:
+    /// - The salt format is "ChainID <chainid>, Version <version>".
+    /// - The version is obtained from the `package.json` file via --ffi cheatcode:
+    /// https://book.getfoundry.sh/cheatcodes/ffi
+    function _constructCreate2Salt() internal returns (bytes32) {
```
PaulRBerg, line 55: We're not in production contracts, let's drop the underscore.

PaulRBerg (changes requested):
I like this idea a lot - but we should absolutely test this new function in `test/utils`.

andreivladbrg:
I don't think is related to QuickNode, since in periphery worked: https://github.com/sablier-labs/v2-periphery/pull/280

smol-ninja:
That looks like rate limiting imposed by [Quicknode](https://www.quicknode.com/pricing) (2500 credits / sec).

PaulRBerg:
Bummer. Let's try another provider - I will switch to ChainNodes.

andreivladbrg:
> but we should absolutely test this new function in test/utils.

note: it would require to include `--ffi` flag in the test commands (locally) otherwise the tests would fail. e.g. `forge test 

do you have any solution on how to fix this in the CI? how to include a flag in the command here:

https://github.com/sablier-labs/v2-core/blob/2b65c464a40fe0924ef1c3ce0aa24a0d90b3abf8/.github/workflows/ci.yml#L39-L45

PaulRBerg:
> do you have any solution on how to fix this in the CI?

Yes. We can add `ffi = true` in the Foundry config file.

test/utils/BaseScript.t.sol
```diff
@@ -0,0 +1,23 @@
...
+    BaseScript internal baseScript = new BaseScript();
+
+    function test_ConstructCreate2Salt() public {
+        string memory chainId = block.chainid.toString();
+        string memory version = "1.1.2";
+        string memory salt = string.concat("ChainID ", chainId, ", Version ", version);
+
+        bytes32 actualSalt = baseScript.constructCreate2Salt();
+        bytes32 expectedSalt = bytes32(abi.encodePacked(salt));
+        assertEq(actualSalt, expectedSalt, "create2 salt mismatch");
```
PaulRBerg, line 21: Let's capitalize CREATE2

script/Base.s.sol
```diff
@@ -7,7 +7,7 @@ import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
 import { console2 } from "forge-std/src/console2.sol";
 import { Script } from "forge-std/src/Script.sol";
 
-abstract contract BaseScript is Script {
+contract BaseScript is Script {
```
PaulRBerg, line 10: Why is this not an abstract contract anymore?

andreivladbrg: you can't initialize an abstract contract: 

https://github.com/sablier-labs/v2-core/blob/c6fcc322ff237a88302b38ec31a9b812d6e73277/test/utils/BaseScript.t.sol#L12

PaulRBerg: Can we please make a mock for `BaseScript` in the testts instead of marking it as non-abstract?

We should not let users accidentally interact with it.

foundry.toml
```diff
@@ -72,6 +72,7 @@
 # Test the optimized contracts without re-compiling them
 [profile.test-optimized]
   src = "test"
+  ffi = true
```
PaulRBerg, line 75: Missing alphabetical ordering

PaulRBerg (changes requested):
See below

script/Base.s.sol
PaulRBerg: Ok, makes sense, let's keep it as non-abstract.

andreivladbrg:
@PaulRBerg does it look good now? can you approve the PR? 

will merge it once the CI is over

PaulRBerg:
This PR continues to be amazingly useful - @andreivladbrg you should tweet about it.

PR #821, opened 2024-02-10
refactor: change disclaimer when the nft is non-transferable

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/818

### DIFF
--- a/src/SablierV2NFTDescriptor.sol
+++ b/src/SablierV2NFTDescriptor.sol
@@ -90,9 +90,10 @@ contract SablierV2NFTDescriptor is ISablierV2NFTDescriptor {
             generateDescription({
                 streamingModel: vars.streamingModel,
                 assetSymbol: vars.assetSymbol,
-                streamId: streamId.toString(),
                 sablierAddress: vars.sablierAddress,
-                assetAddress: vars.asset.toHexString()
+                assetAddress: vars.asset.toHexString(),
+                streamId: streamId.toString(),
+                isTransferable: vars.sablier.isTransferable(streamId)
             }),
             '","external_url":"https://sablier.com","name":"',
             generateName({ streamingModel: vars.streamingModel, streamId: streamId.toString() }),
@@ -250,14 +251,21 @@ contract SablierV2NFTDescriptor is ISablierV2NFTDescriptor {
     function generateDescription(
         string memory streamingModel,
         string memory assetSymbol,
-        string memory streamId,
         string memory sablierAddress,
-        string memory assetAddress
+        string memory assetAddress,
+        string memory streamId,
+        bool isTransferable
     )
         internal
         pure
         returns (string memory)
     {
+        // Depending on the transferability of the NFT, declare the relevant information.
+        string memory info = isTransferable
+            ?
+            unicode"⚠️ WARNING: Transferring the NFT makes the new owner the recipient of the stream. The funds are not automatically withdrawn for the previous recipient."
+            : unicode"❕INFO: This NFT is non-transferable. It cannot be sold or transferred to another account.";
+
         return string.concat(
             "This NFT represents a payment stream in a Sablier V2 ",
             streamingModel,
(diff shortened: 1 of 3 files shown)


smol-ninja (approved):
LGTM.

src/SablierV2NFTDescriptor.sol
```diff
@@ -258,6 +260,13 @@ contract SablierV2NFTDescriptor is ISablierV2NFTDescriptor {
         pure
         returns (string memory)
     {
+        // Depending on the transferability of the NFT, declare the relevant disclaimer.
+        string memory disclaimer = isTransferable
+            ?
+            unicode"⚠️ WARNING: Transferring the NFT makes the new owner the recipient of the stream. The funds are not automatically withdrawn for the previous recipient."
+            :
+            unicode"🔒 IMPORTANT: This NFT is non-transferable. It may be listed on marketplaces but cannot be sold or transferred. Ownership is permanent.";
```
PaulRBerg, line 268: Can we change the text to the following:

> ❕INFO: This NFT is non-transferable. It cannot be sold or transferred to another account.

PaulRBerg (changes requested):
Thanks @andreivladbrg, left some requests

test/unit/concrete/nft-descriptor/generateDescription.t.sol
```diff
@@ -5,11 +5,13 @@ pragma solidity >=0.8.22 <0.9.0;
 import { NFTDescriptor_Unit_Concrete_Test } from "./NFTDescriptor.t.sol";
 
 contract GenerateDescription_Unit_Concrete_Test is NFTDescriptor_Unit_Concrete_Test {
-    string internal constant DISCLAIMER =
+    string internal constant INFO_TRANSFERABLE =
         unicode"⚠️ WARNING: Transferring the NFT makes the new owner the recipient of the stream. The funds are not automatically withdrawn for the previous recipient.";
+    string internal constant INFO_NON_TRANSFERABLE =
```
PaulRBerg, line 10: Not alphabetically ordered

src/SablierV2NFTDescriptor.sol
```diff
@@ -248,6 +249,7 @@ contract SablierV2NFTDescriptor is ISablierV2NFTDescriptor {
 
     /// @notice Generates a string with the NFT's JSON metadata description, which provides a high-level overview.
     function generateDescription(
+        bool isTransferable,
```
PaulRBerg, line 252: Could we please define this parameter last? After `streamId`

andreivladbrg: after `streamId` doesn't mean last:

https://github.com/sablier-labs/v2-core/blob/e8c604733c40cb70accf0ec410cca0d785b6cbd7/src/SablierV2NFTDescriptor.sol#L250-L258

PaulRBerg: Let's move `streamId` after `assetAddress`, and `isTransferable` after `assetAddress`

andreivladbrg: ok 😅

andreivladbrg:
@PaulRBerg all good now?

PaulRBerg (approved):
looks good now

src/SablierV2NFTDescriptor.sol
smol-ninja: Out of curiosity, why didn't you go with alphabetical order in this case?

PaulRBerg: @smol-ninja cuz there is a natural hierarchy: contract variables first, followed by stream data.

In general: alphabetical is the default unless there's another more specific rule that makes sense in a particular case.

But this is more art than science; if you prefer, I would be open to reordering the params alphabetically.

smol-ninja: Let's be artists. Gtg.

PR #825, opened 2024-02-16
feat: declare maxCount in BaseScript

andreivladbrg:
I have named it `maxCount` because it is going to be used for both segments and tranches.

@PaulRBerg fyi i will update the warp drive

### DIFF
--- a/script/Base.s.sol
+++ b/script/Base.s.sol
@@ -10,6 +10,9 @@ import { Script } from "forge-std/src/Script.sol";
 contract BaseScript is Script {
     using Strings for uint256;
 
+    /// @dev The Avalanche chain ID.
+    uint256 internal constant AVALANCHE_CHAIN_ID = 43_114;
+
     /// @dev Included to enable compilation of the script without a $MNEMONIC environment variable.
     string internal constant TEST_MNEMONIC = "test test test test test test test test test test test junk";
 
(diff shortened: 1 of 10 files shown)


script/Base.s.sol
```diff
@@ -22,6 +25,10 @@ contract BaseScript is Script {
     /// @dev Used to derive the broadcaster's address if $ETH_FROM is not defined.
     string internal mnemonic;
 
+    /// @dev The upper limit on the size of specific data structures that are placed in the contract storage to ensure
+    /// transactions stay within the block gas limit.
+    uint256 internal maxCount;
```
smol-ninja, line 30: should be above `mnemonic` (alphabetically)

smol-ninja (changes requested):
LGTM.

script/Base.s.sol
```diff
@@ -22,6 +25,10 @@ contract BaseScript is Script {
     /// @dev Used to derive the broadcaster's address if $ETH_FROM is not defined.
     string internal mnemonic;
 
+    /// @dev The upper limit on the size of specific data structures that are placed in the contract storage to ensure
+    /// transactions stay within the block gas limit.
```
PaulRBerg, line 29: Let's change to:

> The upper limit on the length of data structures to ensure that transactions stay within the block gas limit

script/Base.s.sol
```diff
@@ -37,6 +44,13 @@ contract BaseScript is Script {
             mnemonic = vm.envOr({ name: "MNEMONIC", defaultValue: TEST_MNEMONIC });
             (broadcaster,) = deriveRememberKey({ mnemonic: mnemonic, index: 0 });
         }
+
+        // Sets `maxCount` to 300 for Avalanche, or 500 for all other chains.
```
PaulRBerg, line 48: `and` instead of `or`.

PaulRBerg (changes requested):
feedback below

andreivladbrg:
addressed your suggestions, does it look good now?

smol-ninja (approved):

PR #834, opened 2024-03-05
Use absolute paths to import src contracts in test files

smol-ninja:
It improves readability.

### DIFF
--- a/test/Base.t.sol
+++ b/test/Base.t.sol
@@ -3,14 +3,14 @@ pragma solidity >=0.8.22 <0.9.0;
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
-import { ISablierV2Comptroller } from "../src/interfaces/ISablierV2Comptroller.sol";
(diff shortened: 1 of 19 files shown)


andreivladbrg:
it might improve readability, but the reason why we don't use absolute paths to import src contracts in specific test files is because these might be imported in other repos as well (e.g., [periphery](https://github.com/sablier-labs/v2-periphery/blob/cc0e660e9b8d1277bb8cb0cedd440630d1379c48/test/Base.t.sol#L13-L14)) - and if we make this change, it would revert due to not finding the file

thus, we should not pursue with this change

smol-ninja:
Right. But isn't that only true for `utils` folder since that is part of the npm package?

andreivladbrg:
the repo can be installed as:

```
 "@sablier/v2-core": "github:sablier-labs/v2-core#staging"
```

or as gitmodule:
```
forge install sablier-labs/v2-core
```

we don't know what a person would need from our test files, so we want to offer as much flexibility as possible.

smol-ninja:
I agree. Thanks for your comment.

PR #839, opened 2024-03-07
Remove protocol fee

andreivladbrg:
As discussed in https://github.com/sablier-labs/company-discussions/discussions/2, this PR removes the protocol fees.

With this change, it was possible to remove:

- `SablierV2Comptroller` contract
- `SablierV2Base` contract

PR #841, opened 2024-03-08
Move precompiles to precompiles directory

### LINKED ISSUE #789: Move the precompiles to `script`
As discussed [here](https://github.com/sablier-labs/v2-periphery/issues/238), it would be helpful to ship the precompiles so that we can write a protocol deployment script that spans V2 Core and V2 Periphery.

Task:

- [ ] Move the precompiles in the `script` directory
- [ ] Include the `script/Precompiles.s.sol` file in the npm package (via `files`)

Food for thought: or maybe we shouldn't include the precompiles in `script`? How about another directory, e.g., `precompiles`?

I will let you @andreivladbrg and @smol-ninja self-organize.

smol-ninja:
Resolves #789.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,6 +5,7 @@ coverage:
   status:
     patch: off
 ignore:
+  - "precompiles"
   - "script"
   - "src/libraries/NFTSVG.sol"
   - "src/libraries/SVGElements.sol"
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
   },
   "files": [
     "artifacts",
+    "precompiles",
     "src",
     "test/utils",
     "CHANGELOG.md",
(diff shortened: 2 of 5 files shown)


PaulRBerg (approved):
looks good so far

andreivladbrg (changes requested):
Let's ignore the precompiles dir in the `.codecove.yaml` file because coverage has significantly decreased even though the file is tested.

https://github.com/sablier-labs/v2-core/blob/refactor/precompiles-directory/codecov.yml

Let's first merge #839 and then rebase from `refactor/remove-protocol-fee` to remove comptroller precompiles.

smol-ninja:
Andrei, I have included your suggestions. lmk if it looks good now.

andreivladbrg (approved):
Looks good now, thank you

PR #845, opened 2024-03-12
Remove duplicated script

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/844


precompiles/Precompiles.sol
```diff
@@ -184,10 +184,9 @@ contract Precompiles {
     /// @notice Deploys all V2 Core contracts in the following order:
     ///
     /// 1. {SablierV2NFTDescriptor}
-    /// 2. {SablierV2LockupDynamic}
-    /// 3. {SablierV2LockupLinear}
-    /// 4. {SablierV2LockupTranched}
-    /// 5. {SablierV2NFTDescriptor}
+    /// 2. {SablierV2LockupLinear}
```
smol-ninja, line 187: Since this is the order of the contracts in which they are deployed. It should be 

```solidity
SablierV2LockupDynamic
SablierV2LockupLinear
SablierV2LockupTranched
```

smol-ninja (changes requested):
LGTM except the following comment.

smol-ninja (approved):

PR #848, opened 2024-03-13
refactor: rename functions in Helpers

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/847

### DIFF
--- a/src/SablierV2LockupTranched.sol
+++ b/src/SablierV2LockupTranched.sol
@@ -145,8 +145,8 @@ contract SablierV2LockupTranched is
         noDelegateCall
         returns (uint256 streamId)
     {
-        // Checks: check the durations and generate the canonical tranches.
-        LockupTranched.Tranche[] memory tranches = Helpers.checkDurationsAndCalculateTimestamps(params.tranches);
+        // Generate the canonical tranches.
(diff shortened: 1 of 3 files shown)


src/SablierV2LockupTranched.sol
```diff
@@ -190,18 +190,17 @@ contract SablierV2LockupTranched is
             return 0;
         }
 
-        // If the end time is not in the future, return the deposited amount.
```
PaulRBerg, line 193: The "not" should not have been removed.

An `endTime` lower than `currentTime` is an end time that IS in the past, which means that it IS NOT in the future.

src/SablierV2LockupTranched.sol
```diff
@@ -190,18 +190,17 @@ contract SablierV2LockupTranched is
...
+        // If the end time is in the future, return the deposited amount.
         if (_streams[streamId].endTime <= currentTime) {
             return _streams[streamId].amounts.deposited;
         }
 
         // Sum the amounts in all tranches that precede the current time.
+        // Using unchecked arithmetic is safe because the sum of the tranche amounts are equal to the total amount
```
PaulRBerg, line 199: "is equal"

PaulRBerg (changes requested):
looks good overall, just a few glitches

PR #852, opened 2024-03-15
feat: disallow zero startTime in dynamic and tranched

andreivladbrg:
Addresses https://github.com/sablier-labs/v2-core/discussions/851


test/invariant/Lockup.t.sol
```diff
@@ -95,6 +95,15 @@ abstract contract Lockup_Invariant_Test is Invariant_Test {
         }
     }
 
+    function invariant_DepositedAmountNotZero() external useCurrentTimestamp {
+        uint256 lastStreamId = lockupStore.lastStreamId();
+        for (uint256 i = 0; i < lastStreamId; ++i) {
+            uint256 streamId = lockupStore.streamIds(i);
+            uint128 depositAmount = lockup.getDepositedAmount(streamId);
+            assertNotEq(depositAmount, 0, "Invariant violated: stream non-null, deposited amount zero");
```
smol-ninja, line 103: A thought:

The `depositAmount` and `startTime` are set at the time of creating streams. They are never modified throughout the lifetime of the stream. I don't think they need to be tested as invariants 🤔. Of course we can compare them against zero, but is that a good invariant? If yes, then why not other variables such as `segmentCount` etc.?

smol-ninja (changes requested):
Thanks for the PR Andrei. I have reviewed it and left some comments below:

src/interfaces/ISablierV2LockupDynamic.sol
```diff
@@ -118,6 +118,7 @@ interface ISablierV2LockupDynamic is ISablierV2Lockup {
     /// - Must not be delegate called.
     /// - `params.totalAmount` must be greater than zero.
     /// - If set, `params.broker.fee` must not be greater than `MAX_BROKER_FEE`.
+    /// - `params.startTime` must be greater than zero.
```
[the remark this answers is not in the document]
PaulRBerg: sounds good

test/invariant/Lockup.t.sol
PaulRBerg: We absolutely should keep these tests.

> They are never modified throughout the lifetime of the stream

That is an invariant - the point of an invariant test is to prove that the invariant holds.

It all boils down to fallibilism - an unknown bug could set the deposit amount to zero, or the start time to zero, during the lifecycle of the stream.

> then why not other variables such as segmentCount etc.?

Maybe we should test other variables, too!

But not `segmentCount` because that's not part of contract storage. It's a variable dynamically generated at runtime, and it's associated with a specific function.

PaulRBerg (approved):
good job

test/invariant/Lockup.t.sol
smol-ninja: Fair and agree. Thanks for your comment.

PaulRBerg: No prob

andreivladbrg:
@smol-ninja I have updated the branch with your feedback, does it look good?

smol-ninja (approved):
Perfect. Looks good.

PR #855, opened 2024-03-18
test for `MAX_BROKER_FEE`

smol-ninja:
As discussed in https://github.com/sablier-labs/v2-core/discussions/854#discussioncomment-8826059.


PaulRBerg (approved):
good job

PR #860, opened 2024-03-20
perf: optimising `_calculateStreamedAmount()` of Lockup Tranched

### LINKED ISSUE #858: Gas optimization in `_calculateStreamedAmount()` of Lockup Tranched
Found this while reviewing contracts:
https://github.com/sablier-labs/v2-core/blob/1b2ee6df156b156799a4e3bd680b726079b018fb/src/SablierV2LockupTranched.sol#L203-L211

The above code can be optimised to:
```solidity
uint256 index = 1;
unchecked {
    while (tranches[index].timestamp <= currentTime) {
        streamedAmount += tranches[index].amount;
        index += 1;
    }
}
```

Looping over 100 tranches, it saves ~800 gas.

@andreivladbrg what do you think?

smol-ninja:
closes #858.


andreivladbrg (approved):
Nice finding.

The only downside is that the readability has decreased, but I find it worth it.

smol-ninja:
> The only downside is that the readability has decreased

may be, but for someone who can write recursion code, it should be easy to understand.

PR #864, opened 2024-03-21
Keep streamedAmountOf function only in SablierV2Lockup abstract contract

andreivladbrg:
Addresses https://github.com/sablier-labs/v2-core/discussions/856


src/SablierV2LockupDynamic.sol
```diff
@@ -186,6 +176,18 @@ contract SablierV2LockupDynamic is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc SablierV2Lockup
+    /// @dev The streaming function is:
```
smol-ninja, line 179: Should we add notice tag here as well? Half information is in `SablierV2Lockup` and half here.

@notice Calculates the streamed amount of the stream without looking up the stream's status. This function is declared in `SablierV2Lockup`.

src/SablierV2LockupTranched.sol
```diff
@@ -181,9 +171,17 @@ contract SablierV2LockupTranched is
...
     /// @inheritdoc SablierV2Lockup
+    /// @dev The streaming function is:
+    ///
+    /// $$
+    /// f(x) = \Sigma(esa)
+    /// $$
+    ///
+    /// Where:
+    ///
+    /// - $\Sigma(esa)$ is the sum of all elapsed tranches' amounts.
```
smol-ninja, line 182: I think we should call it `eta` - elapsed tranches' amounts.

src/SablierV2LockupDynamic.sol
andreivladbrg: we have `/// @inheritdoc SablierV2Lockup` and the `override` specifier, so i think that is enough

src/SablierV2LockupTranched.sol
andreivladbrg: correct

src/SablierV2LockupDynamic.sol
smol-ninja: Yes I agree and precisely the reason why I pointed that out because the reader has to go to the abstract just to read the notice tag.

smol-ninja:
Andrei, I've also added Base endpoints in this PR.

src/SablierV2LockupDynamic.sol
andreivladbrg: >because the reader has to go to the abstract just to read the notice tag.

that's pretty much the same for every function that uses inheritdoc to an interfaces, e.g. `createWithDurations` `createWithTimestamps`

andreivladbrg:
> Andrei, I've also added Base endpoints in this PR.

i am sorry that i have not mentioned, but i have pushed it directly to `staging` https://github.com/sablier-labs/v2-core/commit/6775fec1070dc6625db7d4ae59f399d194fb1aa8

src/SablierV2LockupDynamic.sol
smol-ninja: OK. I agree.

smol-ninja:
> i am sorry that i have not mentioned, but i have pushed it directly to staging

no problem. I have reverted it back.

andreivladbrg:
if it looks good, can you approve?

smol-ninja (approved):

PR #874, opened 2024-04-02
Reorder conditions for if-else in LD to increase branch coverage

smol-ninja:


src/SablierV2LockupDynamic.sol
```diff
@@ -240,13 +240,13 @@ contract SablierV2LockupDynamic is
             currentSegmentTimestamp = segments[index].timestamp;
 
             uint40 previousTimestamp;
-            if (index > 0) {
+            if (index == 0) {
+                // Otherwise, the current segment is the first, so use the start time as the previous timestamp.
```
PaulRBerg, line 244: Wording needs to be changed here - "Otherwise" should be moved under the `else`.

PaulRBerg (changes requested):
see comment below. We also need to update precompiles.

PaulRBerg (approved):
lgtm

PR #892, opened 2024-04-10
perf: use unchecked to calculate sender amount

smol-ninja:
As discussed in https://github.com/sablier-labs/v2-core/discussions/882.


PaulRBerg (approved):
gud

PR #895, opened 2024-04-12
feat: add deployment to tenderly virtual testnets

nvitorovic:
This PR adds a script that deploys contracts to a Tenderly Virtual TestNet.
3 pieces of information are needed (`TODO` in the script):

- Tenderly Access Key
- Unlocked RPC URL
- Admin address

Note: the script temporarily modifies `foundry.toml` by adding an `etherscan` config. It cleans it up as soon as the deployment is done.


smol-ninja:
Thank you for the PR. However, as discussed on Telegram, we would like to come back to this in the future when we need it.

PR #907, opened 2024-04-23
build: replace ffi with stdJson for salt creation

smol-ninja:
Related discussion: https://github.com/sablier-labs/v2-core/discussions/904


PaulRBerg (approved):
noice

PR #913, opened 2024-04-30
refactor: `Range` struct to `Timestamp` struct

### LINKED ISSUE #911: Rename "range" data types to "timestamps"
As discussed in https://github.com/sablier-labs/v2-core/discussions/902.

smol-ninja:
Closes #911.


PaulRBerg (changes requested):
The singular `timestamp` should be `timestamps`.

src/types/DataTypes.sol
```diff
@@ -192,6 +184,14 @@ library LockupDynamic {
         Lockup.Amounts amounts;
         Segment[] segments;
     }
+
+    /// @notice Struct encapsulating the LockupDynamic timestamps.
+    /// @param start The Unix timestamp indicating the stream's start.
+    /// @param end The Unix timestamp indicating the stream's end.
+    struct Timestamp {
```
andreivladbrg, line 191: Timestamps

src/types/DataTypes.sol
```diff
@@ -277,6 +267,16 @@ library LockupLinear {
         Lockup.Amounts amounts;
         uint40 cliffTime;
     }
+
+    /// @notice Struct encapsulating the LockupLinear timestamps.
+    /// @param start The Unix timestamp for the stream's start.
+    /// @param cliff The Unix timestamp for the cliff period's end. A value of zero means there is no cliff.
+    /// @param end The Unix timestamp for the stream's end.
+    struct Timestamp {
```
andreivladbrg, line 275: Timestamps

src/types/DataTypes.sol
```diff
@@ -355,6 +347,14 @@ library LockupTranched {
         Tranche[] tranches;
     }
 
+    /// @notice Struct encapsulating the LockupTranched timestamps.
+    /// @param start The Unix timestamp indicating the stream's start.
+    /// @param end The Unix timestamp indicating the stream's end.
+    struct Timestamp {
```
andreivladbrg, line 353: Timestamps

andreivladbrg (changes requested):

andreivladbrg:
i've pushed a new commit @smol-ninja 

lmk if it looks good, otherwise we can revert it

src/types/DataTypes.sol
smol-ninja: The original idea was to rename `Range` to Timestamp` and use plural, `timestamps` for variable name. Ref: https://github.com/sablier-labs/v2-core/discussions/902#discussioncomment-9118755

andreivladbrg: I am little bit confused by these:

1.
image

2. 
image

IMO, it does not make sense to have `Timestamp` over `Timestamps` since there are multiple time variables within the structs, and the variable name is `timestamps`. And one other argument is that we have the function`createWithTimestamps`.

smol-ninja: I agree with your arguments. I would prefer `Timestamps` over `Timestamp`.

andreivladbrg: @PaulRBerg could you please confirm, which version you suggested?

PaulRBerg: It is true that originally I/ we said `Timestamp`, but looking at the PR code I realized that `Timestamps` would work much better.

smol-ninja:
@andreivladbrg since we all have agreed to use `Timestamp`, can we merge it now?

src/types/DataTypes.sol
smol-ninja: Thank you for agreeing.

andreivladbrg (approved):

andreivladbrg:
looking good now, approved

PR #931, opened 2024-05-17
Update count script improvements

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/929

### DIFF
--- a/script/Base.s.sol
+++ b/script/Base.s.sol
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // solhint-disable no-console
+// forgefmt: disable-start
 pragma solidity >=0.8.22 <0.9.0;
 
 import { Strings } from "@openzeppelin/contracts/utils/Strings.sol";
@@ -85,12 +86,12 @@ contract BaseScript is Script {
     /// @dev Populates the segment & tranche count map. Values can be updated using the `update-counts.sh` script.
     function populateSegmentAndTrancheCountMap() internal {
         // Arbitrum chain ID
-        segmentCountMap[42_161] = 1170;
-        trancheCountMap[42_161] = 1210;
+        segmentCountMap[42161] = 1170;
+        trancheCountMap[42161] = 1210;
 
         // Avalanche chain ID.
-        segmentCountMap[43_114] = 530;
-        trancheCountMap[43_114] = 540;
+        segmentCountMap[43114] = 530;
+        trancheCountMap[43114] = 540;
 
         // Base chain ID.
         segmentCountMap[8453] = 2200;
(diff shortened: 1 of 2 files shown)


script/Base.s.sol
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 // solhint-disable no-console
+// forgefmt: disable-start
```
smol-ninja, line 3: Instead of putting this at the top of the file, we can put it inside the function.

```solidity
function populateSegmentAndTrancheCountMap() internal {
  // forgefmt: disable-start

  // --- snip ---

  // forgefmt: disable-end
}
```

So it could just disable formatting for the block and not the whole file.

shell/update-counts.sh
```diff
@@ -52,7 +45,4 @@ update_counts() {
 update_counts "Segments" "segmentCountMap"
 update_counts "Tranches" "trancheCountMap"
 
-# Reformat the code with Forge
-forge fmt $BASE_SCRIPT
```
smol-ninja, line 56: In that case, we can keep it.

smol-ninja (changes requested):

script/Base.s.sol
PaulRBerg: agree with Shub

shell/update-counts.sh
```diff
@@ -28,20 +22,19 @@ update_counts() {
...
         local chain_id=$(echo $line | awk '{print $8}')
-        local formatted_chain_id=$(format_chain_id $chain_id)
 
         # Add the data to the table
-        table+="\n$map_name,$formatted_chain_id,$count"
+        table+="\n$map_name,$chain_id,$count"
 
         # Update the map for each chain ID using sd
-        sd "$map_name\[$formatted_chain_id\] = [0-9_]+;" "$map_name[$formatted_chain_id] = $count;" $BASE_SCRIPT
+        sd "$map_name\[$chain_id\] = [0-9_]+;" "$map_name[$chain_id] = $count;" $BASE_SCRIPT
```
PaulRBerg, line 37: do we need the underscore in this regex?

`[0-9_]+`

shell/update-counts.sh
```diff
@@ -28,20 +22,19 @@ update_counts() {
     local output=$(FOUNDRY_PROFILE=benchmark forge t --mt "test_Estimate${test_name}" -vv)
     echo -e "\nParsing output for $test_name..."
 
-    # Define a table with headers
+    # Define a table with headers. This table is not put in the Solidity script file.
```
PaulRBerg, line 25: Let's explain what the table is used for

shell/update-counts.sh
andreivladbrg: you are correct, we no longer need underscore since we have removed `format_chain_ids`

script/Base.s.sol
andreivladbrg: > So it could just disable formatting for the block and not the whole file.

unfortunately, i tested this before and it doesn’t work as expected (idk if it’s intended or a bug in forge fmt). adding `// forgefmt: disable-start` will be applied globally to the file. that is, if you set it within a function scope, it won’t be applied to the specified range 

so i decided to put it at the start of the file

PaulRBerg: sad. if this is a bug in Forge please report it in the Foundry repository

smol-ninja: Addressed in the latest commit: https://github.com/sablier-labs/v2-core/pull/931/commits/2a3b509fe359ca20ee3a0a88cbea066b0bd60d67

shell/update-counts.sh
smol-ninja: Addressed: https://github.com/sablier-labs/v2-core/pull/931/commits/2a3b509fe359ca20ee3a0a88cbea066b0bd60d67

smol-ninja (approved):

PR #936, opened 2024-05-28
Build/npm ignore file

andreivladbrg:
- Closes https://github.com/sablier-labs/v2-core/issues/935

@PaulRBerg I've also included your commit


smol-ninja (approved):
LGTM.

PaulRBerg:
Thanks @andreivladbrg, just FYI in the future, you can also add me as a [co-author](https://stackoverflow.com/q/7442112/3873510).

PaulRBerg (changes requested):
plesae do not merge this

we shuold not use a `.npmignore` file because it adds very complicated logic when `files` are used at the same time.

I am no longer able to find the reference for this but I left a comment in a Medium post about this issue many years ago

andreivladbrg:
> just FYI in the future, you can also add me as a [co-author](https://stackoverflow.com/q/7442112/3873510).

which commit are you referring to? i used `git cherry-pick` and you are the author of the commit:

image


> we should not use a `.npmignore` file because it adds very complicated logic when `files` are used at the same time.
> 

sheesh, didn't know about this

> Let's not merge this as is.

do you have any suggestion?

PaulRBerg:
All I meant is that instead of cherrypicking, you could have added my as a co-author. Judt FYI. Cherrypicking was fine in this case.

re sheesh - just use `files`, as I have suggested initially. You can put negative globs like this:

```json
{
  "files": ["!test/utils/*.t.sol"]
}
```

Also, why not ignore all `*.t.sol` files? not just those in `test/utils`.

andreivladbrg:
> just use files, as I have suggested initially. You can put negative globs like this:


@PaulRBerg I've added this in `package.json` but it doesn't seem to work, as seen here:

https://app.warp.dev/block/CRhLQnKpcZGWmIdLnGpveL

what if we include this CLI in the [prepack](https://github.com/sablier-labs/v2-core/blob/78d8c5ac9294f0628102834dbff0aff977a32f80/package.json#L68): 

```.json
"find . -name '*.t.sol' -delete",
```

PaulRBerg:
It appears that I was wrong - [negative globs are not supported in the `files` field](https://chatgpt.com/share/47021f48-1c2e-4c48-a51c-3fa98d5844ea).

Therefore, it's impossible to ignore the `*.t.sol` files without using an `.npmignore` file.

I pushed a commit to add an `.npmignore` file in the `test` directory. It has to be added there because otherwise npm will ignore the ignore file itself.

It should be working now - please double-check before merging @andreivladbrg @smol-ninja. Then, squash and merge, please.

PaulRBerg (approved):
lgtm now

smol-ninja:
Thanks @PaulRBerg for the fix. It seems to be working now: https://app.warp.dev/block/8z3gVbenfdyKFq3RsDoFKj

andreivladbrg:
great, ty to both🚀

PR #937, opened 2024-06-02
docs: update security assumptions

### LINKED ISSUE #920: Exclude clickjacking in security disclosure
Here:

https://github.com/sablier-labs/v2-core/blob/main/SECURITY.md

### LINKED ISSUE #921: Add assumption about the max value of `exponent`
A Codehawk auditor asked a really good question about the max value of `UD2x18`.

Since its max value is capped by `type(uint64).max`, the exponent cannot represent a value bigger than `18.446744073709551615`. Thus, we should add this in the [Assumptions section](https://github.com/sablier-labs/v2-core/blob/staging/SECURITY.md#assumptions) of SECURITY.md:

**In Lockup dynamic, the fixed-point representation of a segment's `exponent` does not exceed `18.446744073709551615`.**

PaulRBerg:
### Linked issues
- Closes #920
- Closes #921.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -31,8 +31,8 @@ The Program does NOT cover the following:
 
 Vulnerabilities contingent upon the occurrence of any of the following also are outside the scope of this Program:
 
-- Front-end bugs
-- DDOS attacks
+- Front-end bugs (clickjacking etc.)
+- DDoS attacks
 - Spamming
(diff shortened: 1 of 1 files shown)


smol-ninja (dismissed):
LGTM.

smol-ninja:
Why has the coverage go down so much?

PR #940, opened 2024-06-07
perf: emit events before hook calls

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/939.

### DIFF
--- a/src/abstracts/SablierV2Lockup.sol
+++ b/src/abstracts/SablierV2Lockup.sol
@@ -281,7 +281,7 @@ abstract contract SablierV2Lockup is
     }
 
     /// @inheritdoc ISablierV2Lockup
-    function renounce(uint256 streamId) external override noDelegateCall notNull(streamId) updateMetadata(streamId) {
+    function renounce(uint256 streamId) external override noDelegateCall notNull(streamId) {
         // Check: the stream is not cold.
         Lockup.Status status = _statusOf(streamId);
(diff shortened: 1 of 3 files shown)


andreivladbrg:
@smol-ninja let's not udpate yet the precompiles, will do it after we address all the findings

andreivladbrg (changes requested):
Since the only place where the `updateMetadata` modifier is used is in the `_update` hook, why not remove it completely and emit the event individually in the function? This approach is also more gas efficient

smol-ninja:
Agree. Changing it right away.

smol-ninja:
All changes done. You can review it now @andreivladbrg.

src/abstracts/SablierV2Lockup.sol
```diff
@@ -376,6 +363,9 @@ abstract contract SablierV2Lockup is
         // Effects and Interactions: make the withdrawal.
         _withdraw(streamId, to, amount);
 
+        // Emits an ERC-4906 event to trigger an update of the NFT metadata.
```
PaulRBerg, line 366: Let's say just `Emit` here.

PaulRBerg (approved):
just one comment, otherwise LGTM

smol-ninja:
@andreivladbrg looks good to merge?

andreivladbrg (approved):
yes ser

image

PR #941, opened 2024-06-07
Fix/check start time

andreivladbrg:
Depends on #940

Closes https://github.com/sablier-labs/v2-core/issues/938


test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -1,8 +1,11 @@
 streamedAmountOf.t.sol
+├── given the stream's status is "PENDING"
+│  └── given the cliff time is zero
```
smol-ninja, line 3: this should be "given the start time is in the future"

smol-ninja (changes requested):

test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -17,7 +19,28 @@ contract StreamedAmountOf_LockupLinear_Integration_Concrete_Test is
         StreamedAmountOf_Integration_Concrete_Test.setUp();
     }
 
-    function test_StreamedAmountOf_CliffTimeInThePast()
+    modifier givenStatusPendind() {
+        _;
+    }
+
+    function test_StreamedAmountOf_CliffTimeZero()
```
smol-ninja, line 26:
```suggestion
    function test_StreamedAmountOf_StartTimeInTheFuture()
```

PaulRBerg (changes requested):
test trees must be updated to reflect the new execution paths

side note: we really need to start using Bulloak here

test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree
andreivladbrg: this "given the start time is in the future" branch is already tested in the common `streamedAmountOf` tests here:

https://github.com/sablier-labs/v2-core/blob/8a3121013e2b199dcbac0a83db1c079843f012c3/test/integration/concrete/lockup/streamed-amount-of/streamedAmountOf.tree#L11-L12

we need to have both conditions met, `start time is in the future & cliff time = 0`, and since the start time in the future is already tested, my reasoning was that `given the cliff time is zero` is enough. (also, saying just "given the start time is in the future" is not enough) 

lmk if you have a better suggestion

test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
andreivladbrg: let's first decide on the above comment

andreivladbrg:
> test trees must be updated to reflect the new execution paths

can you please provide more details? (it is a tricky situation here that because we have common tests in `lockup` for this function, and individual tests in each model - linear/dynamic/tranched)

> side note: we really need to start using Bulloak here

yes, it would be helpful, tho, imo it is a low priority atm

test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree
smol-ninja: Isn't it to test `_calculateStreamedAmount` which has the following execution flow:
1. If start time is in the future, i.e. a PENDING stream, it must return 0.
2. If start time is in present/past, i.e. a STREAMING stream, if cliff time is in the future, it must return 0.
3. Otherwise, it must return the correct streamed amount.

There is no need to test "given the stream's status is "PENDING", given the cliff time is zero" because if the stream status is in future, it should return 0 regardless of cliff time.

> start time is in the future & cliff time = 0

Its an `OR` and not `AND`.

> since the start time in the future is already tested

Is it tested in other test contract? If yes, I still it should be tested here because its the test of streamed amount. Previously, because there was only `cliffTime > blockTimestamp` condition, it was not necessary, but now it is.

> given the cliff time is zero is enough

No its not enough. If the start time is in the future, then doesn't matter what the value of cliff time is, it should return 0.

andreivladbrg: > Isn't it to test `_calculateStreamedAmount` which has the following execution flow:
> 
> 1. If start time is in the future, i.e. a PENDING stream, it must return 0.
> 2. If start time is in present/past, i.e. a STREAMING stream, if cliff time is in the future, it must return 0.
> 3. Otherwise, it must return the correct streamed amount.

the problem here is that, the first condition checked in the `if`  statement is `cliffTime > blockTimestamp` and if condition evaluates to true, the second condition  (`startTime > blockTimestamp` ) is not checked (short-circuit)

with the invariants in mind (cliff can be zero):  
- $\text{if cliff} \gt 0 \implies \text{cliff}  \gt \text{start}$ 
- $\text{if cliff = }  0  \implies \text{cliff}  \lt \text{start}$  

we have these possible scenarios:

1. `0 = cliff time` and  `startTime < blockTimestamp` $\Rightarrow$ start time _is checked_ $\Rightarrow$ "PENDING"
2. `0 = cliff time` and  `startTime >= blockTimestamp` $\Rightarrow$ start time _is checked_ $\Rightarrow$ "STREAMING"
3. `0 < blockTimestamp < cliff time` and  `startTime > blockTimestamp` $\Rightarrow$ start time _is checked_ $\Rightarrow$ "PENDING"
4. `0 < blockTimestamp < cliff time` and  `startTime < blockTimestamp` $\Rightarrow$ start time _is not checked_ $\Rightarrow$ "STREAMING"
5. `0 < cliff time < blockTimestamp` and  `startTime < blockTimestamp` $\Rightarrow$ start time _is not checked_ $\Rightarrow$ "STREAMING"


and the problem here is that we have two kind of tests, common (between all models) and individuals, and here, for the linear case we have an unique variable (cliff), and it has specific cases that we don't check the start time. if we were to modify the `_calculateStreamedAmount` function like this:

```solidity
if (startTime > blockTimestamp) {
    return 0;
}

if (cliffTime > blockTimestamp) {
    return 0;
}
``` 

it would have been easier, but we are using a short circuit to be more gas efficient

> Its an OR and not AND.

yes, we use `OR` in the if block, but i was referring to the fact that both conditions must be checked

> There is no need to test "given the stream's status is "PENDING", given the cliff time is zero" because if the stream status is in future, it should return 0 regardless of cliff time

that's true but the condition is going to be checked only if `cliff > 0`

> No its not enough. If the start time is in the future, then doesn't matter what the value of cliff time is, it should return 0

as said in my previous message, the case when the start time is in the future is already tested in the common `lockup` tests, but in the default setup, the cliff is greater than start time, so with this test branch we want to test the first case in the possible scenarios from above 

---

lmk if i made my self clear

smol-ninja: > the problem here is that, the first condition checked in the if statement is cliffTime > blockTimestamp and if condition evaluates to true, the second condition (startTime > blockTimestamp ) is not checked (short-circuit)

This is an optimized implementation and should be done in this manner. My reasoning is in favour of test-driven approach. Instead of having test-later approach, in which the tests follow the implementation, we could also write tests branches in a different order (test driven approach). So that it covers all possible execution branches independently.

> if we were to modify the _calculateStreamedAmount function like this

I also disagree with modifying the `_calculateStreamedAmount` function. The current implementation is well optimized.

> lmk if i made my self clear

Yes you have. The debate seems to be whether tests should follow the execution flow (test later approach) or cover all scenarios in a different order than the implementation itself (test driven approach). In this case, the other approach seems better (because of the reasons you have raised above).

The issue with the proposed changes in this PR is that the test of `streamedAmountOf` missed out that specific case where "startTime is in the future" and it causes selection bias. Even though, its tested in common lockup, `_calculateStreamedAmount` is a part of Lockup Linear so it should be tested here too. Otherwise from a developer pov, it looks like an incomplete tests.

I would recommend this reading: https://about.codecov.io/blog/test-driven-vs-test-later-development-when-should-you-write-your-tests

andreivladbrg: > The issue with the proposed changes in this PR is that the test of `streamedAmountOf` missed out that specific case where "startTime is in the future" and it causes bias. Even though, its tested in common lockup, `_calculateStreamedAmount` is a part of Lockup Linear so it should be tested here too

once again, there are two edge cases (one _is_ tested in common lockup, and one _is not_)
1. `cliff time > 0 and startTime > blockTimestamp` - tested
2. `cliff time = 0 and startTime > blockTimestamp` - not tested

the whole discussion is simple imo, it's just an edge case that must be covered whether we use `test driven` or `test later`

> I would recommend this reading: https://about.codecov.io/blog/test-driven-vs-test-later-development-when-should-you-write-your-tests

thanks for sharing, will have a read (tho, i have my own opinions about this approach, here is not the place to debate them)

--- 

can you please tell again how would you like to call the test after you understood my point? it is a small thing that consumes a lot of time, for me it's important just to cover the edge case

smol-ninja: Fair enough. In that case lets move forward with your approach. When we plan to refactor them using Bulloak, that's when we can debate about it.

> it is a small thing that consumes a lot of time

Yes I agree.

smol-ninja (approved):

test/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -17,7 +19,28 @@ contract StreamedAmountOf_LockupLinear_Integration_Concrete_Test is
         StreamedAmountOf_Integration_Concrete_Test.setUp();
     }
 
-    function test_StreamedAmountOf_CliffTimeInThePast()
+    modifier givenStatusPendind() {
```
IaroslavMazur, line 22: Typo (should be "pendin**g**")

smol-ninja: Thanks for discovering it @IaroslavMazur. @andreivladbrg please feel free to push a commit directly onto staging branch to fix this.

andreivladbrg: e8b9041d79153b00406843ff4a1461868f5507d3

PR #944, opened 2024-06-14
build: bump solc to v0.8.26

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/943


PaulRBerg:
Doesn't this invalidate the recent audits?

PaulRBerg (approved):
lgtm but let's double check that the audit reports are not invalidated

smol-ninja:
> lgtm but let's double check that the audit reports are not invalidated

None of the finding depends on solc version. So, this should not invalidate the audit findings. cc @andreivladbrg for comments, if any.

PaulRBerg:
OK, then, let's merge it

andreivladbrg (approved):
no comments on my side

lgtm👍

PR #945, opened 2024-06-17
refactor: disallow non alphanumeric symbols

andreivladbrg:
We still need to make a decision, but this PR solves this finding: https://www.codehawks.com/report/clvb9njmy00012dqjyaavpl44#M-01


PaulRBerg (approved):
Thanks for the lift-off, @andreivladbrg. I have made several additions and changes:

- [x] Simplified the implementation using OR operators instead of AND
- [x] Added thorough concrete units
- [x] Added fuzz tests

PaulRBerg:
Separately, I opened feature requests in OpenZeppelin and Solady for implementing alphanumeric check functions:

- https://github.com/OpenZeppelin/openzeppelin-contracts/issues/5087
- https://github.com/Vectorized/solady/issues/962

smol-ninja:
Should we allow `-` as well?

From Egis audit report:

> Consider adding - (2D) as supported char in for token sybmol, because it is not a threat for JSON injection and there are tokens such as LP, which may use this char. Example token Recom- mendation:Addbool isDash = char == 2Dcheck

Also, even Sablier NFT contains `-`.

PaulRBerg:
I agree with adding support for `-`.

PR #946, opened 2024-06-17
feat: eip-4906 interface

smol-ninja:
The below PR only includes the test for eip-4906 interface to see whether it fails or succeeds.


andreivladbrg (approved):
LGTM👍

src/abstracts/SablierV2Lockup.sol
```diff
@@ -199,6 +200,11 @@ abstract contract SablierV2Lockup is
         streamedAmount = _streamedAmountOf(streamId);
     }
 
+    /// @inheritdoc ERC721
+    function supportsInterface(bytes4 interfaceId) public view override(ERC721, IERC165) returns (bool) {
+        return interfaceId == 0x49064906 || super.supportsInterface(interfaceId);
```
PaulRBerg, line 205: Someone reading this without context wouldn't understand much. I'll add a comment to explain what `0x49064906` means.

PaulRBerg: Better yet: rewrite it as `type(IERC4906).interfaceId`

src/abstracts/SablierV2Lockup.sol
```diff
@@ -199,6 +200,11 @@ abstract contract SablierV2Lockup is
         streamedAmount = _streamedAmountOf(streamId);
     }
 
+    /// @inheritdoc ERC721
+    function supportsInterface(bytes4 interfaceId) public view override(ERC721, IERC165) returns (bool) {
```
PaulRBerg, line 204: Components with fewer inherited components should be inherited first.

i.e. `IERC165` should come before `ERC721`.

PaulRBerg:
We can actually put it in the interface by inheriting `IERC4906` in `ISablierV2Lockup`.

This is how it should have always been done.

src/abstracts/SablierV2Lockup.sol
smol-ninja: `SablierV2Lockup` does not inherit `IERC165`/`ERC165`. `ERC721` inherits them. So, I added `ERC721` before `IERC165` to make it follow the inheritance tree (Lockup inherits ERC721 inherits ERC165). Do you still think `IERC165` should come before `ERC721`?

src/abstracts/SablierV2Lockup.sol
smol-ninja: Good point.

smol-ninja:
Any reason why [`IERC4906`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/IERC4906.sol) and not just [`IERC165`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/introspection/IERC165.sol)? Its ERC-165 that defines the standard interface.

PaulRBerg:
Because `IERC4906` interface also contains the `MetadataUpdate` and the `BatchMetadataUpdate` events, which we need.

PR #948, opened 2024-06-18
refactor: inherit IERC-4906 in abstract

PaulRBerg:
Refinement of #946

### DIFF
--- a/src/abstracts/SablierV2Lockup.sol
+++ b/src/abstracts/SablierV2Lockup.sol
@@ -2,7 +2,6 @@
 pragma solidity >=0.8.22;
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { ERC721 } from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
 import { IERC721Metadata } from "@openzeppelin/contracts/token/ERC721/extensions/IERC721Metadata.sol";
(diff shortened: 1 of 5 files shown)


src/abstracts/SablierV2Lockup.sol
```diff
@@ -1,8 +1,8 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.8.22;
 
-import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
```
smol-ninja, line 5: This should be above `IERC20` because path has higher priority over the imported contract name.

src/interfaces/ISablierV2Lockup.sol
```diff
@@ -2,6 +2,7 @@
 pragma solidity >=0.8.22;
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
```
smol-ninja, line 5: Same here.

smol-ninja (changes requested):
LGTM except the order of import.

src/abstracts/SablierV2Lockup.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/948/commits/65511d140a078b7f33cf2cd1308d74b4886d023c

src/interfaces/ISablierV2Lockup.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/948/commits/65511d140a078b7f33cf2cd1308d74b4886d023c

smol-ninja (approved):

PR #957, opened 2024-06-29
test: keep gas usage below foundry 1B limit

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/v2-core/issues/956
- Replaced `test_RevertWhen_LoopCalculationOverflowsBlockGasLimit` with `test_RevertWhen_SegmentCountTooHigh` since the function reverts if the count exceeds the max count defined in the contract. https://github.com/sablier-labs/v2-core/blob/3ee63dbeb570b133eb4095e211a5e736704c144c/src/libraries/Helpers.sol#L138

So we may not need the loop test anymore.


andreivladbrg (approved):
image

PR #960, opened 2024-06-29
implementing suggestions from Egis audit review

smol-ninja:
### Changelog
1. Index `admin` in `AllowToHook` event.

> Egis: This way if the admin is changed, it would be easy to query all addresses that were whitelisted by given admin address, etc.

2. Include dash in alphanumeric check

> Egis: Consider adding - (2D) as supported char in for token sybmol, because it is not a threat for JSON injection and there are tokens such as LP, which may use this char. An [example token](https://etherscan.io/token/0xa3ee21c306a700e682abcdfe9baa6a08f3820419) with a market cap of $271M

@andreivladbrg and @PaulRBerg tagging you for feedback on the above changes.


PaulRBerg (changes requested):
A dash `-` is not an alphanumeric character.

Thus, let's rename the `isAlphanumericWithSpaces` function to `isAllowedCharacter`.

smol-ninja:
Will rename it. I agree with your point but couldn't think of a good name. Will go with `isAllowedCharacter`

PaulRBerg (approved):
looks good now, great job

PaulRBerg:
Let's wait for @andreivladbrg's review.

andreivladbrg (approved):
LGTM 👍

PR #961, opened 2024-07-01
refactor: return the withdrawn amount

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/955


test/integration/concrete/lockup/withdraw-max-and-transfer/withdrawMaxAndTransfer.t.sol
```diff
@@ -112,11 +112,11 @@ abstract contract WithdrawMaxAndTransfer_Integration_Concrete_Test is
         emit Transfer({ from: users.recipient, to: users.alice, tokenId: defaultStreamId });
 
         // Make the max withdrawal and transfer the NFT.
-        lockup.withdrawMaxAndTransfer({ streamId: defaultStreamId, newRecipient: users.alice });
+        uint128 expectedWithdrawnAmount =
+            lockup.withdrawMaxAndTransfer({ streamId: defaultStreamId, newRecipient: users.alice });
 
         // Assert that the withdrawn amount has been updated.
         uint128 actualWithdrawnAmount = lockup.getWithdrawnAmount(defaultStreamId);
-        uint128 expectedWithdrawnAmount = withdrawAmount;
```
smol-ninja, line 119: Since this assertion is to verify that the "`getWithdrawnAmount` amount has been correctly updated", it should be be compared against `lockup.withdrawableAmountOf(defaultStreamId)`.

And then we should have a separate assertion to verify that the "returned" value is indeed same as `lockup.withdrawableAmountOf(defaultStreamId)` since thats what it is expected to return.

test/integration/concrete/lockup/withdraw-max-and-transfer/withdrawMaxAndTransfer.tree
```diff
@@ -22,4 +22,5 @@ withdrawMaxAndTransfer.t.sol
                   ├── it should transfer the NFT
                   ├── it should emit a {WithdrawFromLockupStream} event
                   ├── it should emit a {Transfer} event
-                  └── it should emit a {MetadataUpdate} event
+                  ├── it should emit a {MetadataUpdate} event
+                  └── it should return the withdrawn amount
```
smol-ninja, line 26: In the source code, it returns `withdrawableAmountOf` so it should `it should return the withdrawable amount`.

smol-ninja (changes requested):
I have left a few comments below. The same comments also apply to `withdrawMax`.

PaulRBerg (approved):

smol-ninja (approved):

PR #968, opened 2024-07-07
ci: use `fromJSON` to convert string into digits

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/965.

This fixes the issue as can be seen in [this manual run](https://github.com/sablier-labs/v2-core/actions/runs/9828167040). Solution found in [this discussion](https://github.com/actions/runner/issues/2206#issuecomment-1532246677).


PaulRBerg (approved):
Interesting, so it does seem as if GitHub changed the way they parse input values.

Anyway, great sleuthing @smol-ninja!

PR #973, opened 2024-07-15
Add more testnet endpoints

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/972

Notes:
- i've also renamed the admin variable for sepolia(mainnet) to `TESTNET_ADMIN` since it is used for all testnets
- regarding `blast`, i will push directly to its branch once this PR is approved

---

Regarding the CI failure, this is because we’ve renamed `RPC_URL_MAINNET` to `MAINNET_RPC_URL`, causing a mismatch between the expected environment variable in the reusable workflow and the organization secret. I don’t have access to add organization secrets (@PaulRBerg needs to do this).


foundry.toml
```diff
@@ -110,3 +110,6 @@
   optimism = "https://optimism-mainnet.infura.io/v3/${API_KEY_INFURA}"
   polygon = "https://polygon-mainnet.infura.io/v3/${API_KEY_INFURA}"
   sepolia = "https://sepolia.infura.io/v3/${API_KEY_INFURA}"
+  arbitrum_sepolia = "https://arbitrum-sepolia.blockpi.network/v1/rpc/public"
```
smol-ninja, line 113: It would be a good idea to order them alphabetically so that the both of the versions of the same chain are put together.

shell/deploy-multi-chain.sh
```diff
@@ -104,15 +103,17 @@ declare -A chains
 # define function to initialize all configurations
 function initialize {
     chains["arbitrum"]="$ARBITRUM_RPC_URL $ARBISCAN_API_KEY $ARBITRUM_ADMIN"
-    chains["arbitrum_sepolia"]="$ARBITRUM_SEPOLIA_RPC_URL $ARBISCAN_API_KEY $ARBITRUM_SEPOLIA_ADMIN"
+    chains["arbitrum_sepolia"]="arbitrum_sepolia $ARBISCAN_API_KEY $TESTNET_ADMIN"
```
smol-ninja, line 106: it should be `$ARBITRUM_SEPOLIA_RPC_URL` and not `arbitrum_sepolia` in this script. We use [.env.deployment.example](https://github.com/sablier-labs/v2-core/blob/main/.env.deployment.example) to load custom RPCs as public ones may have rate limiting and may not work at the time of deployment.

smol-ninja (changes requested):

shell/deploy-multi-chain.sh
andreivladbrg: my idea with using `<chain>_sepolia` is to read the value from `foundry.toml`, it is similar to:

```bash
FOUNDRY_PROFILE=optimized \
forge script script/DeployDeterministicCore.s.sol \
--broadcast \
--verify \
--etherscan-api-key ARBISCAN_API_KEY \
--rpc-url arbitrum_sepolia \
--sig "run(address)" \
0xb1bEF51ebCA01EB12001a639bDBbFF6eEcA12B9F \
-vvv
```

andreivladbrg: we don't need to re-declare it in `.env.deployment`

smol-ninja: Good point. Then, why don't we remove [.env.deployment](https://github.com/sablier-labs/v2-core/blob/main/.env.deployment) file entirely and only use `foundry.toml` for all the chains? Main network endpoints require `API_KEY_INFURA` in `foundry.toml` s so having `.env.deployment` is just redundant and more efforts to manage.

andreivladbrg: i agree to use only the `foundry.toml` file, but instead of “forcing” someone to use `infura,` we should change it to `<CHAIN_NAME>_RPC_URL`. this way, you can use any RPC provider, such as alchemy

smol-ninja: In that case, they can also change it directly in `foundry.toml` file. My point was to have consistency in how we load those variables. `<CHAIN_NAME>_RPC_URL` works too but does that mean that these RPC URLs must be added to `.env` file? If yes, then you may want to update `.env.example` as well.

andreivladbrg: > CHAIN_NAME>_RPC_URL works too but does that mean that these RPC URLs must be added  to .env file? If yes, then you may want to update .env.example as well:

yes, ofc

smol-ninja: Oh. I was suggesting to completely remove `.env.deployment.example` and only use `.env.example`. During fork test, there is no need to populate all rpc variables. So that deployment script loads from `foundry.toml` which loads from `.env`

foundry.toml
```diff
@@ -99,14 +99,17 @@
...
+  base_sepolia = "https://sepolia.base.org"
   bnb = "https://bsc-dataseed.binance.org"
   gnosis = "https://rpc.gnosischain.com"
   localhost = "http://localhost:8545"
-  ethereum = "${RPC_URL_MAINNET}"
-  mainnet = "${RPC_URL_MAINNET}"
-  optimism = "https://optimism-mainnet.infura.io/v3/${API_KEY_INFURA}"
-  polygon = "https://polygon-mainnet.infura.io/v3/${API_KEY_INFURA}"
-  sepolia = "https://sepolia.infura.io/v3/${API_KEY_INFURA}"
+  mainnet = "${MAINNET_RPC_URL}"
```
smol-ninja, line 110: I still think it is confusing to be subjective. For example, we are using an URL directly here for `base` and `scroll` while for others, we have RPC URL loading from `.env.deployment`. How about the following:

- Keep all variables in `.env.deployment.example`
- Keep only what is required for development in `.env.example`
- Have the same env variables names
- We can even set values to public URL directly in the environment example files.

wdyt?

shell/deploy-multi-chain.sh
smol-ninja: Agree to keep `.env.deployment.example`

foundry.toml
andreivladbrg: > we are using an URL directly here for base and scroll while for others

because the RPC urls used directly here are public by the chain org, [base](https://docs.base.org/docs/network-information/), [bnb](https://www.binance.com/en/square/post/3613468649713), [scroll](https://docs.scroll.io/en/developers/developer-quickstart/).
in case of mainnet (and the other) it requires to have a third party providere (e.g. infura) to use a rpc url

> Keep all variables in .env.deployment.example

we are doing this for the RPCs that are not used in devolpment

> Keep only what is required for development in .env.example

we are doing this, we keep only `MAINNET_RPC_URL` in `.env.example` (which is used for fork tests)

> Have the same env variables names

we are doing this, what var is not having the same name?

> We can even set values to public URL directly in the environment example files 

i don't think this is a good idea, the .env.example is ignored by the system, you would still need to create an .env file to place them

smol-ninja: Agree to all your points. 

> we are doing this, what var is not having the same name?

It was different, but you have made them all same in this PR which is what I was suggesting.

I feel I may have not been able to explain it properly. I created a commit to show what I have in my mind: https://github.com/sablier-labs/v2-core/commit/8a02d6bdb12d9afc4e24197792ec0ba7a62e0035

Let me know your thoughts.

shell/deploy-multi-chain.sh
```diff
@@ -103,17 +102,19 @@ declare -A chains
...
-    chains["optimism"]="$OPTIMISM_RPC_URL $OPTIMISTIC_API_KEY $OPTIMISM_ADMIN"
-    chains["polygon"]="$POLYGON_RPC_URL $POLYGONSCAN_API_KEY $POLYGON_ADMIN"
-    chains["sepolia"]="$SEPOLIA_RPC_URL $ETHERSCAN_API_KEY $SEPOLIA_ADMIN"
-    chains["scroll"]="$SCROLL_RPC_URL $SCROLLSCAN_API_KEY $SCROLL_ADMIN"
+    chains["arbitrum"]="arbitrum $ARBISCAN_API_KEY $ARBITRUM_ADMIN"
+    chains["arbitrum_sepolia"]="arbitrum_sepolia $ARBISCAN_API_KEY $TESTNET_ADMIN"
+    chains["avalanche"]="avalanche $SNOWTRACE_API_KEY $AVALANCHE_ADMIN"
+    chains["base"]="base $BASESCAN_API_KEY $BASE_ADMIN"
+    chains["base_sepolia"]="base_sepolia $BASESCAN_API_KEY $TESTNET_ADMIN"
+    chains["bnb_smart_chain"]="  $BSCSCAN_API_KEY $BNB_ADMIN"
```
smol-ninja, line 110: You missed `bnb` rpc url.

foundry.toml
andreivladbrg: > We can even set values to public URL directly in the environment example files

one more point against this is that you need to declare the RPC URL explicitly in `.env.deployment` file:

```
# foundry.toml file
base = "${BASE_RPC_URL}"
# .env file
BASE_RPC_URL="https://mainnet.base.org"
```

vs

```
# foundry.toml file
base = "https://mainnet.base.org"
```

smol-ninja: Yes thats what I was suggesting but since its a public endpoint, we can even include it in `.env.deployment.example`

shell/deploy-multi-chain.sh
andreivladbrg: correct, thanks

foundry.toml
andreivladbrg: is there any other reason to place them in `.env` file other than consistency? being consistent here means to repeat ourselves, see [comment](https://github.com/sablier-labs/v2-core/pull/973#discussion_r1681021072).

imo, storing actual public env vars in the `.env.example` loses the purpose of a `.env` file:

storing development-only, _non-sensitive_ versions of your service's environment variables

smol-ninja: As resolved over the slack, lets keep it you way.

smol-ninja (approved):
LGTM

Sorry for long discussion on this.

andreivladbrg:
@smol-ninja the PR on reusable workflow repo has been merged, and the ci her does not fail anymore

is it good to be merged now?

smol-ninja (approved):

PR #979, opened 2024-07-24
Include periphery in this repo

andreivladbrg:
This PR addresses the first point in package tethering issue https://github.com/sablier-labs/v2-core/issues/808

The idea here is to create multiple PRs, to make the review process easier (some of PRs will have a failing CI, because the fixes are done in the upcoming ones). 

> [!NOTE]
> This PR is meant to be merged into `staging` after #980 and #981 (and more to come) are merged into it (Merge order: last created –> second last –> … –> first created). The review would be easier if it is done in the order these PRs were created: first 979, second 980, third 981 and so on..

Each PR is going to be left as "draft" until the upcoming one is merged into it. 
last created (not draft), second last(draft) … first created(draft)

### Changes

- Add `core` dir in `src`
- Move all production contracts into `core`
- Add `periphery` dir in `src` 
- Move all [production periphery contracts](https://github.com/sablier-labs/v2-periphery/tree/main/src) here
- Update `script` import paths

---

Question:  I have thought about merging the `Errors` libraries into a single one and adding a dedicated `errors` directory in the `src` root. Also, I was considering whether we should move all interfaces (both core and periphery) to the root

```sh
current
├── core
│   └──everything the same
└── periphery
    └── everything the same

idea
├── core
│   └──everything the same but without errors and interfaces
├── errors
│   └── both core and periphery errors
├── interfaces
│   └── both core and periphery interfaces
└── periphery
    └── everything the same but without errors and interfaces
```

But, I consider that it is better to separate them for an easier repo structure to manage (current version), wdyt, do you have a better suggestion on how to structure the `src`?


andreivladbrg:
@PaulRBerg and @smol-ninja could you please provide your input on the question raised in the PR comment?

PaulRBerg:
This is not an easy choice. I need time to review this PR properly before sharing my input, and I don't yet know when I will be able to take a look at it. I'm sorry.

smol-ninja (approved):
Looks good. Just a note that when we merge these changes into the `staging` branch, we should make sure to add all the authors of v2-periphery as co-authors of the commit.

smol-ninja:
> do you have a better suggestion on how to structure the `src`?

@andreivladbrg I would suggest to keep `core` and `periphery` separate for these PRs and then start a separate discussion about the new structure that you have in your mind. Combining all the interfaces into a separate folder is a good topic to discuss.

andreivladbrg:
> I would rename the PR title to "Merge the Periphery contracts"

the idea is to merge each PR:
image

into this branch, and the merge to `staging`, as stated in the OP comment

so, "Include periphery in this repo" is better, since it will have all relevant files (src, tests, scripts, etc..)

smol-ninja:
https://github.com/sablier-labs/v2-core/pull/979/commits/c00cf23c1b2c788fb4f77495b16aaa89a9928f7d

- I have moved generic functions to Base test contract, such as `expectMultipleCallsToTransferFrom` as they relate to ERC20 transfers. 
- I've also created another section called `CALL EXPECTS - LOCKUP` (please have a look at the base file). `expectMultipleCallsToCreateWithDurationsLD` etc. also relate to the core function even though they are used in periphery.
- Periphery_Test only contains MERKLE-LOCKUP section which imo can also be moved to Base. I don't think it makes Base less readable.

---------------

We should use `super.setUp()` instead of `<PARENT_CONTRACT>.setUp()`. Why? So that, we can remove the setup functions which has no other logic than just calling the parent setup function.

```solidity
// This is not required if we use super.setUp().
function setUp() public virtual override {
    Base_Test.setUp();
}
```

You can argue that `<PARENT_CONTRACT>.setUp()` increases clarity. However, setup functions are always right below the inheritance statements. So its easy to look at the parent contract.

```
contract CHILD is PARENT {
function setUp() public virtual override {
    PARENT.setup();
}
```

@andreivladbrg thoughts?

.github/workflows/ci-deep.yml
```diff
@@ -70,7 +70,7 @@ jobs:
       MAINNET_RPC_URL: ${{ secrets.MAINNET_RPC_URL }}
     uses: "sablier-labs/reusable-workflows/.github/workflows/forge-test.yml@main"
     with:
-      foundry-fuzz-runs: ${{ fromJSON(inputs.forkFuzzRuns) || 1000 }}
+      foundry-fuzz-runs: ${{ fromJSON(inputs.forkFuzzRuns || '1000') }}
       foundry-profile: "test-optimized"
-      match-path: "test/fork/**/*.sol"
+      match-path: "test/{core,periphery}fork/**/*.sol"
```
smol-ninja, line 75: I think we should separate out core and periphery runs? For example 

```yaml
core-test-fork:
  needs: ["lint", "build"]
  secrets:
    MAINNET_RPC_URL: ${{ secrets.MAINNET_RPC_URL }}
  uses: "sablier-labs/reusable-workflows/.github/workflows/forge-test.yml@main"
  with:
    foundry-fuzz-runs: 20
    foundry-profile: "test-optimized"
    match-path: "test/core/fork/**/*.sol"
    name: "Core fork tests"

periphery-test-fork:
  needs: ["lint", "build"]
  secrets:
    MAINNET_RPC_URL: ${{ secrets.MAINNET_RPC_URL }}
  uses: "sablier-labs/reusable-workflows/.github/workflows/forge-test.yml@main"
  with:
    foundry-fuzz-runs: 20
    foundry-profile: "test-optimized"
    match-path: "test/periphery/fork/**/*.sol"
    name: "Periphery fork tests"
```

So when a test fails, it would be easy to debug.

andreivladbrg: Doing so it would be verbose, how can it improve debugging? You still can see the test fail/test path:

image

image

shell/deploy-multi-chain.sh
```diff
@@ -288,7 +288,7 @@ for chain in "${provided_chains[@]}"; do
     if [[ ${DETERMINISTIC_DEPLOYMENT} == true ]]; then
         echo -e "${SC}+${NC} Deterministic address"
         if [[ ${sol_script} == "" ]]; then
-            deployment_command+=("script" "script/DeployDeterministicCore.s.sol")
+            deployment_command+=("script" "script/DeployDeterministicProtocol.s.sol")
```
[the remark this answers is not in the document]
andreivladbrg: > We have discussed this previously that deploying the entire protocol in a single tx would cause gas to exceed block limit.

the problem before was not about gas, but about different optimizer runs between core/periphery which lead to a NFTDescriptor and Dynamic contract too big, now since we use the same foundry configuration it would work (tested):

https://github.com/sablier-labs/v2-core/blob/a34e5287bb77c3e20b8fb9d3ef349a4a80d990aa/foundry.toml#L18

https://github.com/sablier-labs/v2-periphery/blob/2d338a7ff6452b20a25e779f5a960f047a5d16af/foundry.toml#L20

> But lets create another issue for that

don't think we need one as per above reason

.github/workflows/ci-deep.yml
smol-ninja: Agree that it wont improve debugging. I actually wanted to make the flow looks like the following:

Screenshot 2024-07-30 at 10 40 53

where we have separate boxes for core and periphery. I thought it would look cool. But I couldn't figure out how to do it through yaml file. Marking it resolve as I agree with you.

shell/deploy-multi-chain.sh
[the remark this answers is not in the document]
smol-ninja: Oh. I tested it yesterday and it showed gas to be over 30M. How much gas requirement it shows to you?

smol-ninja:
> do as you wish

Then I would refactor it to use `super.setUp` in the next PR so that we can remove those redundant setup functions. They take up 7 lines.

PaulRBerg:
> I have decided to keep `<CONTRACT>.setUp().`

gud, me like it

PR #984, opened 2024-07-25
docs: status temperature

PaulRBerg:


src/types/DataTypes.sol
```diff
@@ -53,14 +53,19 @@ library Lockup {
     }
 
     /// @notice Enum representing the different statuses of a stream.
+    /// @dev The status can have a "temperature":
+    /// 1. Warm: Pending, Streaming. The passage of time alone can can change the status.
```
smol-ninja, line 57:
```suggestion
    /// 1. Warm: Pending, Streaming. The passage of time alone can change the status.
```

smol-ninja (changes requested):
LGTM otherwise.

PR #988, opened 2024-07-25
Include both core and periphery tests in CI

andreivladbrg:
Please refer to https://github.com/sablier-labs/v2-core/pull/979 notes.

### Changes
- run the tests for both core and periphery in CI
- lower the number of fuzz runs for integration tests
    - the reason for this is because it takes a long time to pass (more tests added) the CI, and the operation would be canceled by github.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,18 +24,18 @@ jobs:
     needs: ["lint", "build"]
     uses: "sablier-labs/reusable-workflows/.github/workflows/forge-test.yml@main"
     with:
-      foundry-fuzz-runs: 5000
+      foundry-fuzz-runs: 2000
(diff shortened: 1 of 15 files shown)


.github/workflows/ci.yml
```diff
@@ -24,26 +24,26 @@ jobs:
     needs: ["lint", "build"]
     uses: "sablier-labs/reusable-workflows/.github/workflows/forge-test.yml@main"
     with:
-      foundry-fuzz-runs: 5000
+      foundry-fuzz-runs: 2000
```
smol-ninja, line 27: Why did you lower it to 2000?

smol-ninja: nvm. Found the answer.

smol-ninja (approved):

PR #1016, opened 2024-08-13
test(refactor): dry'ify claim test contracts

smol-ninja:
@andreivladbrg This PR contains my suggestions for the following PR:
- https://github.com/sablier-labs/v2-core/pull/1011


test/periphery/integration/merkle-campaign/lt/claim/claim.t.sol
```diff
@@ -2,17 +2,20 @@
...
 
-contract Claim_MerkleLT_Integration_Test is Claim_Integration_Test, MerkleLT_Integration_Shared_Test {
-    function setUp() public override(Claim_Integration_Test, MerkleLT_Integration_Shared_Test) {
-        super.setUp();
+contract Claim_MerkleLT_Integration_Test is Claim_Integration_Test {
+    function setUp() public virtual override {
+        Claim_Integration_Test.setUp();
+
+        // Cast the {MerkleLT} contract as {ISablierMerkleBase}
+        merkleBase = ISablierMerkleBase(merkleLT);
```
andreivladbrg, line 18: as you said, we don't need this if we `MerkleLT_Integration_Shared_Test` contract

andreivladbrg (commented):
Thanks for the PR.

Sorry if I missunderstood your initial point, but TBH I don't think having actual tests in the `MerkleLL` and `MerkleInstant` contracts is a good idea. Having them placed in a dedicated `claim` dir it is better organized. 
Also, they are placed under "SHARED" header.

test/periphery/integration/merkle-campaign/lt/claim/claim.t.sol
smol-ninja: Yes but in that suggestion I wanted to move it to `MerkleLT.t.sol`.

smol-ninja:
> they are placed under "SHARED" header

Yes because most of the test functions in Claim are shared except the `test_Claim`. Its a Shared test with one exception.

> TBH I don't think having actual tests in the MerkleLL and MerkleInstant contracts is a good idea

No problem. Feel free to close this PR. I was not sure if I was able to convey my points through comments so I created this PR to help you understand. Now I guess you understood what I said and refute to accept these changes so thats fine by me.

smol-ninja:
> TBH I don't think having actual tests in the MerkleLL and MerkleInstant contracts is a good idea. Having them placed in a dedicated claim dir it is better organized.

Out of curiosity, why do you think its not a good idea? My rationale was that by doing so, we can keep all shared contracts in the same file even though claim requires overriding test_claim function.

andreivladbrg:
> I was not sure if I was able to convey my points through comments so I created this PR to help you understand. Now I guess you understood what I said and refute to accept these changes so thats fine by me 

Got it, thanks 🫡

> Out of curiosity, why do you think its not a good idea?

I don’t think it’s good because the actual tests require more maintenance, and that file should be changed, in my view, only if more shared features are added or removed.

PR #1018, opened 2024-08-16
feat: disallow sender to be zero address

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/v2-core/issues/1008
- Disallow sender to be zero address in all `_create` functions

### DIFF
--- a/src/core/SablierLockupDynamic.sol
+++ b/src/core/SablierLockupDynamic.sol
@@ -329,7 +329,9 @@ contract SablierLockupDynamic is
             Helpers.checkAndCalculateBrokerFee(params.totalAmount, params.broker.fee, MAX_BROKER_FEE);
 
         // Check: validate the user-provided parameters.
-        Helpers.checkCreateLockupDynamic(createAmounts.deposit, params.segments, MAX_SEGMENT_COUNT, params.startTime);
+        Helpers.checkCreateLockupDynamic(
+            params.sender, createAmounts.deposit, params.segments, MAX_SEGMENT_COUNT, params.startTime
+        );
 
         // Load the stream ID in a variable.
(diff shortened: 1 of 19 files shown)


andreivladbrg:
Before leaving a full review, is there a specific reason the `params.sender == address(0) `check is not in the `Helpers.checkCreate` function, but in the contracts themselves (LD, LL, LT)?

smol-ninja:
The goal was to revert as soon as possible if `sender` is zero without wasting any gas in `checkAndCalculateBrokerFee` call. Moving it into `Helpers.checkCreate` is technically same but it increases the revert gas cost.

andreivladbrg:
> The goal was to revert as soon as possible

the same argument can be applied to all other checks, but we place them in the library Helpers, as it is better structured, the gas implications are small, so i would suggest to move the check there
wdyt?

PaulRBerg:
Agree with @andreivladbrg, let's move the check in the `Helpers` library. The gas saving should be negligible.

smol-ninja:
> the same argument can be applied to all other checks

I don't think so. The other checks require a call to the `checkAndCalculateBrokerFee` function to update the amount values, which is then followed by the `checkCreateLockupXYZ` function. Whereas `sender` can be checked independently.

andreivladbrg:
> The other checks require a call to the `checkAndCalculateBrokerFee`

the timestamps don't require this

smol-ninja:
I've moved the sender check to `Helpers` library.

andreivladbrg (approved):
Screenshot 2024-07-22 at 3 18 14 PM

PaulRBerg:
love the LGTM memes

PR #1019, opened 2024-08-18
refactor: amounts type reference in Lockup.Stream struct

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/975


andreivladbrg (approved):
LGTM

PR #1021, opened 2024-08-19
Add deploy multi chain rust script

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/971

This PR re-implements the logic from the .sh [script](https://github.com/sablier-labs/v2-core/blob/main/shell/deploy-multi-chain.sh) into a Rust script to make it more robust and maintainable.

### New features not mentioned in the issue:
- Implement Solidity functions to write the deployed contracts in the [docs](https://docs.sablier.com/contracts/v2/deployments) format, allowing for easy copy-pasting.
- Move the admin multisig declarations in Soldity
- Automatically reads all configured chains from `foundry.toml`.
- Automates the process of moving the broadcast files to `v2-deployments`.

### Usage
1. `cd deploy-multi-chain`
2. Examples:
    - `cargo run` - Defaults to Sepolia and simulates the deployment.
    - `cargo run -- --all` - Simulates the deployment to all chains found in the `foundry.toml` file.
    - `cargo run -- --all --broadcast` - Deploys to all chains found in the `foundry.toml` file.
    - `cargo run -- --cp-bf` - Simulates on Sepolia and moves the broadcasted JSON files generated by Foundry to the `v2-deployments` directory at the path: `"../../v2-deployments/protocol/v{}/broadcasts/{}.json"`. If the dir doesn’t exist, it will be automatically created.

### Notes
- The [paths declared](https://github.com/sablier-labs/v2-core/blob/5d314f8604bb53144bc7aea281fe6fdcf52d6037/deploy-multi-chain/src/main.rs#L168) (e.g. `"../foundry.toml"`) in the Rust script are relative to the directory from which the script is run.
    - In the Usage section, I mentioned the first step is to use "cd", as it is a requirement to run `cargo run` from the root of the `deploy-multi-chain` dir
    - We can change the paths to run it as a CLI standalone with a command like `deploy-multi-chain --all`, but this will require installing it with `cargo install --path .` - this is up for debate.
- I have used the current timestamp to generate always an unique file name in case it already exists (ref point 3 in the issue)
- I have not added styling to the script output, as it may change.


PaulRBerg (commented):
Shouldn't this script live in its own dedicated repository?

1. We might want to use it in other Solidity repos, e.g. Flow
2. Rust code requires a dependency management solution (i.e. Cargo) and a formatter (RustFmt)

Check out my [rust-template](https://github.com/PaulRBerg-rust-template).

andreivladbrg:
> Shouldn't this script live in its own dedicated repository?

it doesn’t have the “bandwidth” to have its own dedicated repository. it’s built specifically to run the `DeployProtocol` or `DeployDeterministicProtocol` scripts.

creating a project of this nature requires more time and full focus. this is actually what sphinx (if you remember them) wanted to build, but they didn’t succeed.

also, i wouldn’t have built it in rust, as it isn’t the most user-friendly for these kinds of tasks :)) i would have chosen something lighter, like js or python

> Rust code requires a dependency management solution (i.e. Cargo) and a formatter (RustFmt)

lol, ik :))

andreivladbrg:
The PR is ready to be reviewed, @PaulRBerg @smol-ninja.

The code is more concise compared to the `.sh` script, but if you think the switch to Rust isn’t justified, I understand. I built it as a starting project to learn more about the language itself.

PaulRBerg:
> it doesn’t have the “bandwidth” to have its own dedicated repository.

I think it does. It seems like a bad practice to mix two programming paradigms (Solidity and Rust) in the same repository.

> creating a project of this nature requires more time and full focus

You can keep the code as is. All you would need to do in a separate repo is use git submodules to import the Solidity code.

> lol, ik :))

Sorry, I missed the `Cargo.toml`. However, this PR does not introduce a [`rustfmt.toml`](https://github.com/PaulRBerg/rust-template/blob/main/rustfmt.toml) file.

andreivladbrg:
> I think it does. It seems like a bad practice to mix two programming paradigms (Solidity and Rust) in the same repository.

i honestly don’t think it is bad practice to mix multiple programming languages in a single repo, especially since Solidity is a very specific language meant for implementing Ethereum smart contracts. as our codebase becomes more complex, it’s not feasible to use just one language for every task.

> You can keep the code as is. All you would need to do in a separate repo is use git submodules to import the Solidity code.

i think i have an idea - we don’t need to use git submodules, as we can just install it as a standalone CLI that is meant to be used in the root of this repo. please refer to my first note in the OP

PaulRBerg:
> honestly don’t think it is bad practice to mix multiple programming languages in a single repo

I'm still against it, but OK, let's see how it works with the Rust code in this repo.

> as we can just install it as a standalone CLI

In that case, I'd move it to a separate repo. That's where *standalone* stuff should be.

andreivladbrg:
Closing in favor of https://github.com/sablier-labs/v2-core/pull/1049 and the new repo created: https://github.com/sablier-labs/deploy-multi-chain

PaulRBerg:
Great decision

PR #1038, opened 2024-09-09
feat: sablier fee on Airstreams claim

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/v2-core/issues/1032
- Closes https://github.com/sablier-labs/v2-core/issues/1040
    - `SablierFee` struct
- Adds a new user: `campaignOwner` in `struct Users`
- Adds `BUSL-1.1` license to `SablierMerkleFactory`


src/periphery/abstracts/SablierMerkleBase.sol
```diff
@@ -31,12 +31,18 @@ abstract contract SablierMerkleBase is
...
+    address public immutable FACTORY;
+
     /// @inheritdoc ISablierMerkleBase
     bytes32 public immutable override MERKLE_ROOT;
 
     /// @dev The name of the campaign stored as bytes32.
     bytes32 internal immutable NAME;
 
+    /// @inheritdoc ISablierMerkleBase
+    uint256 public immutable SABLIER_FEE;
```
andreivladbrg, line 44: why don't we also remove the variable from the merkle contract and keep it only in factory?

```solidity
function claim(uint256 index, address recipient, uint128 amount, bytes32[] calldata merkleProof) external payable {
        // Check: `msg.value` is not less than the sablier fee.
        if (msg.value < FACTORY. SABLIER_FEE()) {
            revert Errors.SablierMerkleBase_InsufficientFeePayment(msg.value, SABLIER_FEE);
        }

        // --snip--
}
```

this approach might more expensive, as there is one more external call, but i believe this to be better as we have the flexibility to active all merkle campaigns, even after deployment

wdyt?

andreivladbrg (commented):
i've seen that you have modified the contract as per my suggestion

i do have one more question:

src/periphery/abstracts/SablierMerkleBase.sol
smol-ninja: > why don't we also remove the variable from the merkle contract and keep it only in factory?

Same reason as [given here](https://github.com/sablier-labs/v2-core/issues/1032#issuecomment-2337913704), so that Sablier admin cannot change the fee for active campaigns. It would be a bad idea to display one fee at the time of launching a campaign, and then update it later. Campaign creators may not like that.

PaulRBerg:
Thanks, @smol-ninja!

We need to implement some additional functionality, unfortunately.

https://github.com/sablier-labs/v2-core/issues/new

smol-ninja:
> We need to implement some additional functionality, unfortunately

No problem. How do we proceed @andreivladbrg? Should I close this PR then or should we review and merge it and then work on top of it?

smol-ninja:
I think reviewing and merging would be better as I have already worked on some of the changes that would be required for https://github.com/sablier-labs/v2-core/issues/1040 or the other PR can be merged into `feat/claim-fee-airstream`.

smol-ninja:
@andreivladbrg You can review this PR now. Its completed from my end.

smol-ninja:
Bumping this up @andreivladbrg so that you can add it to your to do task.

andreivladbrg:
I’m sorry, @smol-ninja 😞. I rebased on `staging` from `main`, and it caused a lot of conflicts in this PR. I didn’t realize the consequences. I will help you fixing them

PaulRBerg:
`git reflog`

smol-ninja:
> I’m sorry, @smol-ninja 😞. I rebased on staging from main, and it caused a lot of conflicts in this PR. I didn’t realize the consequences. I will help you fixing them

No problem @andreivladbrg. I am sure it wont take more than 2 mins to fix it. These ain't as scary as they look in Github UI.

src/periphery/interfaces/ISablierMerkleFactory.sol
```diff
@@ -135,4 +180,54 @@ interface ISablierMerkleFactory {
...
+    ///
+    /// Notes:
+    /// - This function transfers ETH to the provided address. If the receiver is a contract, it must be able to receive
+    /// ETH.
+    ///
+    /// Requirements:
+    /// - The caller must be the admin.
+    ///
+    /// @param to The address to receive the Sablier fees.
+    function withdrawFees(address payable to, ISablierMerkleBase merkleLockup) external;
```
andreivladbrg, line 232: merkls instant isn't considered a merkle "lockup", so, we should rename the param to something else like `sablierMerkle`

src/periphery/interfaces/ISablierMerkleBase.sol
```diff
@@ -81,4 +88,16 @@ interface ISablierMerkleBase is IAdminable {
...
     /// @param amount The amount of tokens to claw back.
     function clawback(address to, uint128 amount) external;
+
+    /// @notice Withdraws the Sablier fees accrued to the provided address.
+    ///
+    /// @dev This function transfers ETH to the provided address. If the receiver is a contract, it must be able to
+    /// receive ETH.
+    ///
+    /// Requirements:
+    /// - The caller must be the Factory contract.
```
andreivladbrg, line 98:
```
/// - `msg.sender` must be the `FACTORY` contract.
```

src/periphery/interfaces/ISablierMerkleFactory.sol
```diff
@@ -68,12 +84,30 @@ interface ISablierMerkleFactory {
         pure
         returns (bool result);
 
+    /// @notice Retrieves the default sablier fee required to claim an airstream.
+    /// @dev A minimum of this fee must be paid in ETH during `claim`.
+    function defaultSablierFee() external view returns (uint256);
+
+    /// @notice Retrieves the custom sablier fee struct for a specified campaign creator.
+    /// @dev It return two fields:
```
andreivladbrg, line 92: let's remove this `dev` natspec as the struct is already explained in `DataTypes`

src/periphery/interfaces/ISablierMerkleFactory.sol
andreivladbrg, line 232: should we  add `to` zero address check, in case we mistakenly pass it?

andreivladbrg (changes requested):
So sorry for the late review, we’ve been busy with Flow, and I’ve just gotten my focus back on Lockup.

---

I haven’t fully reviewed the tests yet, but I’ll leave my feedback on `src` below:

Also, why `campaignOwner` and not `campaignAdmin`?

smol-ninja:
Finally. Thanks for the partial review @andreivladbrg. Appreciate it!

> Also, why `campaignOwner` and not `campaignAdmin`?

No reason just https://github.com/sablier-labs/v2-core/issues/1032#issuecomment-2337907316: `campaignAdmin` didn't come to my mind when I was proposing this. Paul was fine with anything as long as it is prefixed with `campaign` and you didn't object as well.

src/periphery/SablierMerkleInstant.sol
```diff
@@ -21,7 +21,12 @@ contract SablierMerkleInstant is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Constructs the contract by initializing the immutable state variables.
-    constructor(MerkleBase.ConstructorParams memory baseParams) SablierMerkleBase(baseParams) { }
+    constructor(
+        MerkleBase.ConstructorParams memory baseParams,
+        uint256 sablierFee
```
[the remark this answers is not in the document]
smol-ninja: `baseParams` is passed as an input parameter to `createMerkle` functions and then it is used in `SablierMerkleBase` without any modification to the values, whereas `sablierFee` is calculated during the execution.

Therefore, `sablierFee` does not fit well into `ConstructorParams`.

src/periphery/SablierMerkleFactory.sol
```diff
@@ -1,23 +1,49 @@
...
+{
+    /*//////////////////////////////////////////////////////////////////////////
+                                  STATE VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierMerkleFactory
+    uint256 public defaultSablierFee;
+
+    /// @dev A mapping of custom Sablier fees by user.
+    mapping(address campaignCreator => MerkleFactory.SablierFee customFee) private _sablierFeeByUser;
```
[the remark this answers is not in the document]
smol-ninja: Agree. Will change it to  `_sablierFeeByUsers` and `MerkleFactory.SablierFeeByUser`.

Its declared `private` to signal that its not used in any of the child contracts i.e. its private to factory contract only. Why would you mark it to `internal`?

src/periphery/abstracts/SablierMerkleBase.sol
```diff
@@ -49,7 +55,7 @@ abstract contract SablierMerkleBase is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Constructs the contract by initializing the immutable state variables.
-    constructor(MerkleBase.ConstructorParams memory params) {
+    constructor(MerkleBase.ConstructorParams memory params, uint256 sablierFee) {
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038#discussion_r1797787350

src/periphery/SablierMerkleFactory.sol
```diff
@@ -99,16 +184,21 @@ contract SablierMerkleFactory is ISablierMerkleFactory {
             )
         );
 
+        // Fetch the Sablier fee for the user, or use the default fee.
+        uint256 sablierFee =
+            _sablierFeeByUser[msg.sender].enabled ? _sablierFeeByUser[msg.sender].fee : defaultSablierFee;
```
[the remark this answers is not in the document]
smol-ninja: `_computeSablierFeeForUser`?

src/periphery/SablierMerkleFactory.sol
```diff
@@ -36,17 +62,71 @@ contract SablierMerkleFactory is ISablierMerkleFactory {
...
+    {
+        return _sablierFeeByUser[campaignCreator];
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                         ADMIN-FACING NON-CONSTANT FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierMerkleFactory
+    function resetSablierFeeByUser(address campaignCreator) external override onlyAdmin {
```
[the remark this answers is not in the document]
smol-ninja: It was [suggested in Slack](https://sablier.slack.com/archives/C06BB90SY02/p1726060104041329?thread_ts=1726059679.845309&cid=C06BB90SY02) to have an explicit function such as `resetFee` for explicitness.

src/periphery/SablierMerkleFactory.sol
```diff
@@ -36,17 +62,71 @@ contract SablierMerkleFactory is ISablierMerkleFactory {
...
+
+        emit SetDefaultSablierFee(msg.sender, defaultFee);
+    }
+
+    /// @inheritdoc ISablierMerkleFactory
+    function setSablierFeeByUser(address campaignCreator, uint256 fee) external override onlyAdmin {
+        MerkleFactory.SablierFee storage feeByUser = _sablierFeeByUser[campaignCreator];
+
+        // If user does not belong to the custom fee list.
+        if (!feeByUser.enabled) feeByUser.enabled = true;
```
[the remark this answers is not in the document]
smol-ninja: But but I love "ternary operator" xD. Isn't one line much cleaner to read in this case? it looks so biutiful 🥹.

I will add `check`.

src/periphery/SablierMerkleFactory.sol
```diff
@@ -36,17 +62,71 @@ contract SablierMerkleFactory is ISablierMerkleFactory {
...
+    /// @inheritdoc ISablierMerkleFactory
+    function setDefaultSablierFee(uint256 defaultFee) external override onlyAdmin {
+        // Effect: update the default Sablier fee.
+        defaultSablierFee = defaultFee;
+
+        emit SetDefaultSablierFee(msg.sender, defaultFee);
+    }
+
+    /// @inheritdoc ISablierMerkleFactory
+    function setSablierFeeByUser(address campaignCreator, uint256 fee) external override onlyAdmin {
```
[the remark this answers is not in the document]
smol-ninja: Given that this function will mostly be called by the admin either directly through etherscan or a contract call, wouldn't it be a good UX if we allow simpler datatypes to be the input parameters instead of a struct? I suppose Sablier UI wont integrate these functions.

Even though it means we need two functions `setSablierFeeByUser` and `resetSablierFeeByUser` whereas with your approach, we would need only `setSablierFeeByUser`.

My only argument in favour of keeping `setSablierFeeByUser` and `resetSablierFeeByUser` is that they are more explicit to what they do, leaves no room for errors (for example, an admin cannot accidentally pass `{fee: some number, enabled: false}`).

src/periphery/interfaces/ISablierMerkleFactory.sol
```diff
@@ -135,4 +180,54 @@ interface ISablierMerkleFactory {
...
+    ///
+    /// Notes:
+    /// - The new fee will only be applied to the future campaigns.
+    ///
+    /// Requiurements:
+    /// - The caller must be the admin.
+    ///
+    /// @param campaignCreator The user for whom the fee is being set.
+    /// @param fee The new fee to be set.
+    function setSablierFeeByUser(address campaignCreator, uint256 fee) external;
```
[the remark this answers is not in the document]
smol-ninja: Lets settle on the above comment first.

src/periphery/interfaces/ISablierMerkleFactory.sol
```diff
@@ -135,4 +180,54 @@ interface ISablierMerkleFactory {
...
+
+    /// @notice Sets the default Sablier fee for claiming an airstream.
+    /// @dev Emits a {SetDefaultSablierFee} event.
+    ///
+    /// Notes:
+    /// - The new default fee will only be applied to the future campaigns.
```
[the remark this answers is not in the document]
smol-ninja: Your suggestion does not sound valid when default fee is changed. Did you mean

`Previously deployed campaigns are not affected by this new fee, only future ones will include the new fee`

src/periphery/interfaces/ISablierMerkleFactory.sol
smol-ninja: `merkleBase`? given the type is `ISablierMerkleBase`. merkle is a known keyword to represent merkle tree so calling it sablier merkle may not be appropriate.

src/periphery/interfaces/ISablierMerkleBase.sol
smol-ninja: Do you want to replace `caller` with `msg.sender`? if so, we have been using caller to represent `msg.sender` through the natspecs.

Agree to use `FACTORY` keyword.

src/periphery/interfaces/ISablierMerkleFactory.sol
smol-ninja: Fair point. I wanted to add a `@return` tag but couldn't figure out a way to I decided to include the `@dev` tag.

smol-ninja:
As per Slack conversation between @razgraf and @andreivladbrg, its beneficial for the UI to include `sablierFee` in the `CreateMerkle` events. However, to be able to include `sablierFee` in `CreateMerkleLT` event, I would either have to merge [`_deployMerkleLT` code](https://github.com/sablier-labs/v2-core/blob/952406952b2de58c9b0db77963006da16cf8fb2c/src/periphery/SablierMerkleFactory.sol#L228) into `createMerkleLT` or return `sablierFee` in `_deployMerkleLT` to then access it in the event log.

@andreivladbrg, which one would you go for? I’d like to go with the first approach as `_deployMerkleLT`  is anyways only called in `createMerkleLT` so why not include the code directly and avoid having returned value. But I do not remember why we decided to have a separate function `_deployMerkleLT`.

src/periphery/interfaces/ISablierMerkleBase.sol
andreivladbrg: > Do you want to replace `caller` with `msg.sender`? if so, we have been using caller to represent `msg.sender` through the natspecs.

yes, that's going to be consistent over core and periphery (e.g. [1](https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/src/core/interfaces/ISablierLockup.sol#L205), [2](https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/src/core/interfaces/ISablierLockup.sol#L220), [3](https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/src/core/interfaces/ISablierLockup.sol#L237))

smol-ninja: Sounds good. I will replace `caller` with `msg.sender` then.

smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038/commits/155401cfafa1e2e76c64f4b23a6ec53ede5b91d9

src/periphery/SablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: yeah, `_computeSablierFeeForUser` sounds good to me

src/periphery/SablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: i agree with you i like it too, i would let @PaulRBerg decide

src/periphery/SablierMerkleInstant.sol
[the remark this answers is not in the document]
andreivladbrg: ohh, yeaah, my bad, missed the fact that is passed in the function param

> sablierFee does not fit well into ConstructorParams

you are correct

src/periphery/SablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: i have missed this suggestion, sorry, i see now

src/periphery/SablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: > directly through etherscan or a contract call, wouldn't it be a good UX if we allow simpler datatypes to be the input parameters instead of a struct

correct, using a struct isn't the best experience, we could write it with direct params: `function setSablierFeeByUser(address campaignCreator, uint256 fee, bool enabled)` 

but as you said:
>> My only argument in favour of keeping setSablierFeeByUser and resetSablierFeeByUser is that they are more explicit to what they do, leaves no room for errors

which it makes sense and i agree with

src/periphery/interfaces/ISablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: yes

src/periphery/interfaces/ISablierMerkleFactory.sol
andreivladbrg: `markleBase` works as well for me

src/periphery/interfaces/ISablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: keeping "set" with "reset" makes sense now

src/core/abstracts/SablierLockup.sol
```diff
@@ -51,10 +51,8 @@ abstract contract SablierLockup is
     /// @dev Emits a {TransferAdmin} event.
     /// @param initialAdmin The address of the initial contract admin.
```
andreivladbrg, line 52: i am wondering if we still need this natspec comments if we already have it in `Adminable` contract?

src/periphery/SablierMerkleFactory.sol
```diff
@@ -1,23 +1,47 @@
...
+
+    /// @dev A mapping of custom Sablier fees by user.
+    mapping(address campaignCreator => MerkleFactory.SablierFeeByUser customFee) private _sablierFeeByUsers;
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                     CONSTRUCTOR
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Emits a {TransferAdmin} event.
+    /// @param initialAdmin The address of the initial contract admin.
```
andreivladbrg, line 42: same comment about natspec

src/periphery/SablierMerkleFactory.sol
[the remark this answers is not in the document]
andreivladbrg: > Its declared `private` to signal that its not used in any of the child contracts i.e. its private to factory contract only. Why would you mark it to `internal`?

the only argument is that we used to be consistent with `internal` over everything, but we can change this to no longer be. i agree with this.

> _deployMerkleLT as private instead of internal?

i agree with this as well

andreivladbrg (commented):
2 more comments re `src`

test/periphery/fork/merkle-campaign/MerkleLT.t.sol
```diff
@@ -171,7 +179,22 @@ abstract contract MerkleLT_Fork_Test is Fork_Test {
             vars.merkleProof = getProof(leaves.toBytes32(), vars.leafPos);
         }
 
-        vars.merkleLT.claim({
+        // Expect call to `claim` with `sablierFee` as msg.value on the merkleLL contract.
+        vm.expectCall(
```
andreivladbrg, line 183: the common function

andreivladbrg (approved):
I just reviewed the tests as well, and they generally look good. (pushed some small polishes cf0990ec2468368758f8e90e500b9e291ff83949)

There are a couple of things I would update (due to the new shared structure in periphery), but to avoid doing the same work twice after rebasing in [this PR](https://github.com/sablier-labs/v2-core/pull/1058), I’ll address them there.

---

Sorry again for the long time waited. We are finally here 🚀

src/periphery/SablierMerkleFactory.sol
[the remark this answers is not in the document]
smol-ninja: Not everything. [Helper contract has private functions](https://github.com/sablier-labs/v2-core/blob/main/src/libraries/Helpers.sol#L217). But may be thats an exception.

Regardless, since you also agree with this, sounds good to me. We can use `internal` for functions/variables that can be overriden by the child contract and `private` for functions/variables are not supposed to be overriden by the child contracts.

src/core/abstracts/SablierLockup.sol
smol-ninja: What do you mean? The `param` natspec is independent of that.

test/periphery/fork/merkle-campaign/MerkleLT.t.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038/commits/ce35add0e51a934348f5f917460d126f8645506a

test/periphery/fork/merkle-campaign/MerkleInstant.t.sol
```diff
@@ -157,13 +165,28 @@ abstract contract MerkleInstant_Fork_Test is Fork_Test {
             vars.merkleProof = getProof(leaves.toBytes32(), vars.leafPos);
         }
 
+        // Expect call to `claim` with `sablierFee` as msg.value on the merkleInstant contract.
+        vm.expectCall(
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038/commits/ce35add0e51a934348f5f917460d126f8645506a

src/periphery/SablierMerkleFactory.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038/commits/ce35add0e51a934348f5f917460d126f8645506a

smol-ninja: see above comment.

andreivladbrg:
> @andreivladbrg, which one would you go for? I’d like to go with the first approach as `_deployMerkleLT` is anyways only called in `createMerkleLT` so why not include the code directly and avoid having returned value. But I do not remember why we decided to have a separate function `_deployMerkleLT`.

@smol-ninja I have missed this one. 
Do what you think is best.

The reason I introduced the internal function for deployment was due to a "stack too deep" error. If you can manage to abstract the logic into a single function, feel free to go for it.

src/periphery/interfaces/ISablierMerkleFactory.sol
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038/commits/ce35add0e51a934348f5f917460d126f8645506a

test/Base.t.sol
```diff
@@ -218,6 +223,22 @@ abstract contract Base_Test is Assertions, Calculations, Constants, DeployOptimi
         });
     }
 
+    /*//////////////////////////////////////////////////////////////////////////
+                            CALL EXPECTS - MERKLE LOCKUP
```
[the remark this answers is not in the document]
andreivladbrg: > But then we should also move `CALL EXPECTS - LOCKUP` to `Periphery.t.sol `?

yeah, sure, sounds good

> . My view is that Base contracts works too.

the problem i see with is that the contract becomes extra populate and might be hard to follow. but technically it works there, yes


> never understood what dictates what should go in `Periphery.t.sol` because core does not have such a contract

core does not have it, as what could be placed there, it is needed in the periphery as well. also, in core we have `Integration.t.sol`, which doesn't exist in periphery dir

src/core/abstracts/SablierLockup.sol
andreivladbrg: the `param`natspec i agree, we should keep, but the `dev` one?

smol-ninja:
@andreivladbrg I have resolved all your comments, except https://github.com/sablier-labs/v2-core/pull/1038#discussion_r1811524241, https://github.com/sablier-labs/v2-core/pull/1038#discussion_r1811524614 and https://github.com/sablier-labs/v2-core/pull/1038#discussion_r1811508596 which requires your attention. Feel free to comment/resolve them if you agree. 

Else if all looks good, I'd appreciate your kind approval.

src/core/abstracts/SablierLockup.sol
smol-ninja: Agree with you on that.

smol-ninja: You commented over `param` line so I thought thats you meant. Resolved in https://github.com/sablier-labs/v2-core/pull/1038/commits/5b24e38af6dbbd3ce876bd1b776fb9e7bdd06cf4.

src/periphery/SablierMerkleFactory.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1038/commits/5b24e38af6dbbd3ce876bd1b776fb9e7bdd06cf4

PR #1046, opened 2024-09-25
ci: fix slack notify

PaulRBerg:
This is what [ChatGPT recommended](https://chatgpt.com/share/66f3f856-20c4-800e-8f9b-65bb7ab23f11).

However, I wanted to double-check the recommendation, so I posted a [Q&A on StackOverflow](https://stackoverflow.com/q/79022703/3873510). Keeping the PR in draft mode until I receive confirmation.


PaulRBerg:
Merging since I have not received an answer on StackOverflow so we should just test this in practice.

PR #1058, opened 2024-10-09
Dry fy shared logic and add util Modifiers contract

andreivladbrg:
### Changes

Closes https://github.com/sablier-labs/v2-core/issues/1028 and https://github.com/sablier-labs/v2-core/issues/1045

---

## Motivation

There are two things that this PR should solve:

1. Making refactoring the tests a **much** easier process
2. New readers will spend **much** less time understanding the test structure (which is currently very **convoluted**, as we’ll see below)

### About the refactor process

I will start with what this PR began with, which is adding the `Modifiers` utility contract. Currently, in the `stating`
branch, we have two possible places where we declare our modifiers:

1. In the test files themselves:

https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/test/core/integration/concrete/lockup/burn/burn.t.sol#L38

2. In the shared files between integration tests (1. concrete, 2. fuzz):

https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/test/core/integration/shared/lockup/createWithTimestamps.t.sol#L13

Some of the modifiers contained logic that assumed some setup beforehand, but now, since the `Modifier` contract is the
parent (in terms of inheritance), they can't have that logic anymore.

Also, after moving all the modifiers to the new contract mentioned above, I ended up with shared files that contained
only two lines of code, e.g. this one:

https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/test/core/integration/shared/lockup/streamedAmountOf.t.sol#L7-L11

Which made me question whether we really need these shared files. The purpose of these files is to have a unique setup
or specific helper logic between `<function_name>_Concrete_Test` and `<function_name>_Fuzz_Test` (which we’ll call 'Type of test'). However, this was not the case for the vast majority of them, as they only contained duplicated logic throughout
the test files. Faced with this problem, I asked myself where the best place to put this duplicated logic would be, and
the answer is `Integration_Test`. But once I started doing this, I realized there was another component in this
refactor: `Lockup_Integration_Shared_Test`, which complicates things further from an inheritance tree standpoint as the only purpose of it is to declare some virtual functions, which are implemented later ([LL](https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/test/core/integration/shared/lockup-linear/LockupLinear.t.sol#L11)/[LD](https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/test/core/integration/shared/lockup-dynamic/LockupDynamic.t.sol#L11)/[LT](https://github.com/sablier-labs/v2-core/blob/a322ba684812fefb6fbf41cf6fd12e88904df662/test/core/integration/shared/lockup-tranched/LockupTranched.t.sol#L11)).

There are a lot of pieces in this puzzle that don’t need to be divided into so many parts, as it makes 1) making changes
to the code and 2) understanding the code for new readers _more fragile_.

At this point, I really had to understand our inheritance tree of a lockup function, and I came up with this diagram:

<details><summary>Click to see diagram </summary>
<p>

```mermaid
flowchart TD
    FunctionName_LockupModel_Integration_TypeTest --> LockupModel_Integration_TypeTest
    FunctionName_LockupModel_Integration_TypeTest --> FunctionName_Integration_TypeTest
    LockupModel_Integration_TypeTest --> Integration_Test
    LockupModel_Integration_TypeTest --> LockupModel_Integration_Shared_Test
    FunctionName_Integration_TypeTest --> Integration_Test
    FunctionName_Integration_TypeTest --> FunctionName_Integration_Shared_Test
    LockupModel_Integration_Shared_Test --> Lockup_Integration_Shared_Test
    FunctionName_Integration_Shared_Test --> Lockup_Integration_Shared_Test
    Lockup_Integration_Shared_Test --> Base_Test
    Integration_Test --> Base_Test
    Base_Test --> Utils
```

</p>
</details> 

After seeing this, we can conclude some obvious things:

1. `Lockup_Integration_Shared_Test` and `Integration_Test` share the same depth, indicating that they can be abstracted into a single test.
2. Both `LockupModel_Integration_TypeTest` and `FunctionName_Integration_TypeTest` depend on `Lockup_Integration_Shared_Test` and `Integration_Test`, making the organization difficult to follow.

This structure adds unneeded complexity, it's like trying to reach your left ear with your right hand.

image

By abstracting the `Lockup_Integration_Shared_Test`'s logic into `Integration_Test`, we can remove `LockupModel_Integration_TypeTest`completely, as its only purpose is to call the setup functions from the left side (`LockupModel_Integration_Shared_Test`) and the right side (`Integration_Test`) of the diagram. The resulting inheritance diagram would look like this:

<details><summary>Click to see diagram</summary>
<p>

```mermaid
flowchart TD
    FunctionName_LockupModel_Integration_TypeTest --> LockupModel_Integration_Shared_Test
    FunctionName_LockupModel_Integration_TypeTest --> FunctionName_Integration_TypeTest
    LockupModel_Integration_Shared_Test --> Integration_Test
    FunctionName_Integration_TypeTest --> Integration_Test
    Integration_Test --> Base_Test
    Base_Test --> Utils
```

</p>
</details> 

As you can see, it is much simpler now to follow.

<details><summary>Click to see test tree structure comparison</summary>
<p>

### Before PR

```tree
test/core/integration/shared/
├── lockup
│   ├── Lockup.t.sol
│   ├── cancel.t.sol
│   ├── cancelMultiple.t.sol
│   ├── createWithDurations.t.sol
│   ├── createWithTimestamps.t.sol
│   ├── getWithdrawnAmount.t.sol
│   ├── streamedAmountOf.t.sol
│   ├── withdraw.t.sol
│   ├── withdrawMax.t.sol
│   ├── withdrawMaxAndTransfer.t.sol
│   ├── withdrawMultiple.t.sol
│   └── withdrawableAmountOf.t.sol
├── lockup-dynamic
│   └── LockupDynamic.t.sol
├── lockup-linear
│   └── LockupLinear.t.sol
├── lockup-tranched
│   └── LockupTranched.t.sol
└── nft-descriptor
    └── NFTDescriptor.t.sol
```

### After PR

```tree
test/core/integration/shared/
├── lockup
│   ├── cancelMultiple.t.sol
│   └── withdrawMultiple.t.sol
├── lockup-dynamic
│   └── LockupDynamic.t.sol
├── lockup-linear
│   └── LockupLinear.t.sol
├── lockup-tranched
│   └── LockupTranched.t.sol
└── nft-descriptor
    └── NFTDescriptor.t.sol
```

</p>
</details> 

---

### Conclusion

This PR will bring improvements for both the Sablier team and external contributors, enhancing collaboration and code quality.


andreivladbrg:
I will provide a broad explanation of this PR later.

andreivladbrg:
> Just to confirm this PR does more than what you mentioned in the OP?

yes, it does more

>  these refactors are pre requisite for them?

some of them were a pre requisite of moving the modifiers to its dedicated contract, and after doing so, i realized that we can doo a lot more 

> if you could tl;dr summary of changes and why it was required before proceeding further on this.

you may have missed my [prev comment](https://github.com/sablier-labs/v2-core/pull/1058#issuecomment-2402837503)

i left the PR not as draft, to see if the CI passes, now i can mark as draft until i finish with the explanations

> it would also be a good idea to start a discussion on this to share ideas and then create issues followed by PR instead of directly doing them? Wdyt? 

as mentioned above, it all started with some required refactors due to moving the modifiers. after that, i made additional changes to be more efficient once i got started. i wasn’t even able to compile the code, so it was necessary to first get something functional before having a discussion to see if it works in practice.
but i agree with you, discussions are good to have beforehand

smol-ninja:
As discussed on Slack, lets create a flow chart (sample below) to visualise tests inheritance on `staging` and `test/common-modifiers` so that the difference can be compared and benefits can be measures quantitatively. It will also help us to visualize if there are more room for improvements given that this is a second similar refactor in this year. So lets do it properly once and for all.

Second thing is that now we have to prioritize it since it will cause conflict with all other work as almost all test files have changes. Or other priority 1 task will case conflicts with this one. So if its possible to split it into multiple PRs to avoid much conflicts with other works, that would be great.

Third, from next time, I'd appreciate if we can create discussion for such ideas first and then do them based on the priority. For example, imo, this can also be marked as priority 3 where we alredy have many priority 1 tasks in the v1.3.0 roadmap. 

@andreivladbrg I'd also like to hear your views on the above points if you agree or disagree (and why).

image (1)

andreivladbrg:
@smol-ninja I have updated the PR OP

> conflict with all other work as almost all test files have changes. Or other priority 1 task will case conflicts with this one

I suggest we apply the rule of chronological order—PRs that were opened first should have the priority to be merged first. Subsequent ones will need to resolve any conflicts.

> So if its possible to split it into multiple PRs to avoid much conflicts with other works, that would be great.

Unfortunately, it’s not possible to divide this into multiple PRs, which highlights the complexity of the current structure and the need for change.

> from next time, I'd appreciate if we can create discussion for such ideas first and then do them based on the priority

I’ll do my best to be transparent and proactive about what I’m working on. As I mentioned on Slack, I’ve shared these ideas with you multiple times via text and call (before actually creating this PR), and the major change in this PR (adding the Modifier contract) was already discussed on GitHub, and we all agreed on it.


> this can also be marked as priority 3 where we alredy have many priority 1 tasks in the v1.3.0 roadmap

I would honestly mark this as a higher priority (maybe level 2), as it will help other tasks move forward more efficiently. However, I agree it’s not an urgent issue.

smol-ninja:
> I suggest we apply the rule of chronological order—PRs that were opened first should have the priority to be merged first. Subsequent ones will need to resolve any conflicts.

Yup. But my point was about priority of the tasks. The chronology in which tasks be picked should depend on the priority (thats why the label). Given that its a huge PR with priority 2, it should have been picked after priority 1 tasks are finished. If it were a small PR, then that would have been fine. But now since its open, we will have to follow the chronology of the PRs regardless.

> Unfortunately, it’s not possible to divide this into multiple PRs

No problem.

> I’ll do my best to be transparent and proactive about what I’m working on

Thank you. I had missed you mentioning it. I wasn't sure how complex you were talking about. Creating an issue to track that complex work could have been useful too.

> I would honestly mark this as a higher priority (maybe level 2),

Thats OK. My point was about communicating the same in advance through Github issues labelling system.

Rest good to me. I will review it soon.

andreivladbrg:
> Creating an issue to track that complex work could have been useful too.
> My point was about communicating the same in advance through Github issues labelling system.

not sure what you mean? the issue was already created (1028) has the following labels from 28 August:

image

smol-ninja:
@andreivladbrg can you please update this PR? I would like to review it now.

andreivladbrg:
@smol-ninja it is updated now, you can start the review 👌

test/core/integration/concrete/lockup/burn/burn.t.sol
```diff
@@ -2,177 +2,142 @@
...
-        givenDepletedStream(notTransferableStreamId)
-        whenAuthorizedCaller
-        givenNFTExists
+        givenDepletedStream(lockup, defaultStreamId)
+        whenCallerNotRecipient
     {
-        // It should emit a {MetadataUpdate} event.
-        vm.expectEmit({ emitter: address(lockup) });
-        emit MetadataUpdate({ _tokenId: notTransferableStreamId });
+        resetPrank({ msgSender: users.recipient });
```
smol-ninja, line 83: Shouldn't this be `operator`?

test/core/integration/concrete/lockup/burn/burn.t.sol
```diff
@@ -2,177 +2,142 @@
...
-    function test_WhenCallerApprovedThirdParty()
+    /// @dev For this test, we will skip the `givenDepletedStream` modifier given the `notTransferableStreamId` is not
+    /// yet created.
+    function test_GivenNonTransferableNFT()
         external
         whenNoDelegateCall
         givenNotNull
-        givenDepletedStream(streamId)
-        whenAuthorizedCaller
+        /*  givenDepletedStream(lockup, notTransferableStreamId) */
```
smol-ninja, line 116: - Why its OK to skip `givenDepletedStream` branch for this test? 
- If its OK to skip `givenDepletedStream` for this test, why mention these comments? Are we going to test `givenDepletedStream` status anywhere else?

test/core/integration/concrete/lockup/cancel-multiple/cancelMultiple.t.sol
smol-ninja, line 1: Since `CancelMultiple_Integration_Shared_Test` inherits from `Integration_Test`, why does this contract have to inherit from both `CancelMultiple_Integration_Shared_Test` and `Integration_Test`? 

Same question for `WithdrawMultiple_Integration_Shared_Test`

test/core/integration/concrete/lockup/get-asset/getAsset.t.sol
```diff
@@ -3,13 +3,13 @@ pragma solidity >=0.8.22 <0.9.0;
...
-import { Integration_Test } from "./../../../Integration.t.sol";
-import { Lockup_Integration_Shared_Test } from "./../../../shared/lockup/Lockup.t.sol";
 
-abstract contract GetAsset_Integration_Concrete_Test is Integration_Test, Lockup_Integration_Shared_Test {
-    function setUp() public virtual override(Integration_Test, Lockup_Integration_Shared_Test) { }
+import { Integration_Test } from "../../../Integration.t.sol";
 
+abstract contract GetAsset_Integration_Concrete_Test is Integration_Test {
     function test_RevertGiven_Null() external {
+        Integration_Test.setUp();
```
smol-ninja, line 11: We should call this in `function setUp` and not here.

test/core/integration/concrete/lockup/renounce/renounce.t.sol
```diff
@@ -57,16 +41,6 @@ abstract contract Renounce_Integration_Concrete_Test is Integration_Test, Lockup
         lockup.renounce(defaultStreamId);
     }
 
-    /// @dev This modifier runs the test twice: once with a "PENDING" status, and once with a "STREAMING" status.
-    modifier givenWarmStream() {
-        vm.warp({ newTimestamp: getBlockTimestamp() - 1 seconds });
```
smol-ninja, line 62: I can see that the code of `givenWarmStream` in `Modifiers` contract does not contain the same implementation as this. Was this redundant? If no, then the same logic should also be present in the modifiers contract.

test/utils/Modifiers.t.sol
```diff
@@ -0,0 +1,529 @@
...
+
+    modifier whenAssetERC20() {
+        _;
+    }
+
+    modifier whenAuthorizedCaller() {
+        _;
+    }
+
+    modifier whenCallerAdmin(address admin) {
```
smol-ninja, line 72: To avoid these, I would suggest the following:

1. Create a `Users` contract.
2. Move `Types` into `Users`
3. Initialize `users` variable in `Users` contract and create all of them
4. Inherit `Modifiers` from `Users`
5. Remove users setup from Base.

Simply use `users.admin` etc. in this contract.

test/core/integration/concrete/lockup/withdraw-max-and-transfer/withdrawMaxAndTransfer.t.sol
```diff
@@ -137,6 +137,7 @@ abstract contract WithdrawMaxAndTransfer_Integration_Concrete_Test is
         givenTransferableStream
         givenNotBurnedNFT
         givenNonZeroWithdrawableAmount
+        whenCallerRecipient(users.recipient)
```
smol-ninja, line 140: This modifier should not be needed as the name of the test is `test_WhenCallerCurrentRecipient`.

test/core/integration/concrete/lockup-dynamic/create-with-timestamps/createWithTimestamps.t.sol
```diff
@@ -9,20 +10,16 @@ import { stdError } from "forge-std/src/StdError.sol";
...
-    {
-        LockupDynamic_Integration_Concrete_Test.setUp();
-        CreateWithTimestamps_Integration_Shared_Test.setUp();
+
+import { LockupDynamic_Integration_Shared_Test } from "../LockupDynamic.t.sol";
+
+contract CreateWithTimestamps_LockupDynamic_Integration_Concrete_Test is LockupDynamic_Integration_Shared_Test {
+    uint256 streamId;
+
+    function setUp() public virtual override(LockupDynamic_Integration_Shared_Test) {
```
smol-ninja, line 19: Do we need to mention name of overriding contract here?

test/core/integration/concrete/lockup-dynamic/get-stream/getStream.t.sol
```diff
@@ -4,14 +4,11 @@ pragma solidity >=0.8.22 <0.9.0;
...
 import { LockupDynamic } from "src/core/types/DataTypes.sol";
 
-import { LockupDynamic_Integration_Concrete_Test } from "../LockupDynamic.t.sol";
-
-contract GetStream_LockupDynamic_Integration_Concrete_Test is LockupDynamic_Integration_Concrete_Test {
-    uint256 internal defaultStreamId;
+import { LockupDynamic_Integration_Shared_Test } from "../LockupDynamic.t.sol";
 
+contract GetStream_LockupDynamic_Integration_Concrete_Test is LockupDynamic_Integration_Shared_Test {
     function setUp() public virtual override {
```
smol-ninja, line 10: You can remove this setup entirely from non abstract contracts if it does not contain any necessary logic.

test/core/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -3,35 +3,39 @@ pragma solidity >=0.8.22 <0.9.0;
...
-        override(LockupDynamic_Integration_Concrete_Test, StreamedAmountOf_Integration_Concrete_Test)
-    {
-        LockupDynamic_Integration_Concrete_Test.setUp();
-        StreamedAmountOf_Integration_Concrete_Test.setUp();
+    function setUp() public virtual override(LockupDynamic_Integration_Shared_Test, Integration_Test) {
+        LockupDynamic_Integration_Shared_Test.setUp();
     }
 
-    function test_GivenStartTimeInPresent() external givenSTREAMINGStatus {
+    function test_GivenStartTimeInPresent() external givenSTREAMINGStatus(defaults.WARP_26_PERCENT()) {
```
smol-ninja, line 16: What if we inherit Modifiers from `Defaults`. can we remove the input argument from modifiers completely?

test/core/integration/concrete/lockup-dynamic/token-uri/tokenURI.t.sol
```diff
@@ -35,17 +34,10 @@ contract TokenURI_LockupDynamic_Integration_Concrete_Test is LockupDynamic_Integ
         lockupDynamic.tokenURI({ tokenId: nullStreamId });
     }
 
-    modifier givenNFTExists() {
-        defaultStreamId = createDefaultStream();
```
smol-ninja, line 39: Since this logic has been removed from `givenNFTExists` defined in Modifiers contract, dont we need it?

test/core/integration/concrete/lockup-linear/LockupLinear.t.sol
smol-ninja, line 1: Arguments similar to comments on Lockup Dynamic tests.

test/core/integration/concrete/lockup-tranched/LockupTranched.t.sol
smol-ninja, line 1: Comments similar to Lockup Dynamic.

smol-ninja (changes requested):
Good work @andreivladbrg. I have reviewed the core tests. I will review periphery after you have replied to my comments below. Two things I want to mention:

1. Even after closing all the tabs, the browser was working slow because of large number of changes. This is a good argument in favour of why we should split large PRs like this into multiple PRs so that each PR has changes in not more than 50 files. An idea could be to have one PR for core test, another PR for periphery test etc.

2. Most of the comments below apply throughout the code. So if we decide to resolve one comment by making proposed changes, they will also apply to other parts of the diff where I have not commented.

test/utils/Modifiers.t.sol
```diff
@@ -0,0 +1,529 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity >=0.8.22;
+
+import { ISablierLockup } from "src/core/interfaces/ISablierLockup.sol";
+
+import { Utils } from "./Utils.sol";
+
+abstract contract Modifiers is Utils {
```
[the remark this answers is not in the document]
andreivladbrg: correct it works for me
the only thing is that i remember @PaulRBerg making an argument to have explicit inherits in `Base_Test`. is it still the case?

test/core/unit/concrete/adminable/transfer-admin/transferAdmin.t.sol
```diff
@@ -33,14 +30,14 @@ contract TransferAdmin_Unit_Concrete_Test is Adminable_Unit_Shared_Test {
...
     }
 
-    modifier whenNewAdminNotSameAsCurrentAdmin() {
-        _;
-    }
-
-    function test_WhenNewAdminZeroAddress() external whenCallerAdmin whenNewAdminNotSameAsCurrentAdmin {
+    function test_WhenNewAdminZeroAddress()
+        external
+        whenCallerRecipient(users.admin)
```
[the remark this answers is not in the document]
andreivladbrg: i like the idea, i am gonna make the change

test/core/integration/Integration.t.sol
```diff
@@ -15,10 +19,20 @@ import {
...
+    //////////////////////////////////////////////////////////////////////////*/
+
+    uint256 internal defaultStreamId;
+
     /*//////////////////////////////////////////////////////////////////////////
                                    TEST CONTRACTS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @dev A test contract that is meant to be overridden by the implementing contract, which will be
+    /// either {SablierLockupLinear} or {SablierLockupDynamic}.
```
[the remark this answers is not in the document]
andreivladbrg: i guess this was the previous version from `Lockup_Shared` contract and i missed this mistake

test/core/integration/Integration.t.sol
```diff
@@ -60,4 +67,52 @@ abstract contract Integration_Test is Base_Test {
...
+    function createDefaultStreamWithAsset(IERC20 asset) internal virtual returns (uint256 streamId);
+
+    /// @dev Creates the default stream with the provided broker.
+    function createDefaultStreamWithBroker(Broker memory broker) internal virtual returns (uint256 streamId);
+
+    /// @dev Creates the default stream with the provided end time.
+    function createDefaultStreamWithEndTime(uint40 endTime) internal virtual returns (uint256 streamId);
+
+    /// @dev Creates the default stream with the provided user as the recipient and the sender.
+    function createDefaultStreamWithIdenticalUsers(address user) internal virtual returns (uint256 streamId) {
```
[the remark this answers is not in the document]
andreivladbrg: correct

test/core/integration/concrete/lockup/burn/burn.t.sol
```diff
@@ -2,177 +2,142 @@
...
-
-abstract contract Burn_Integration_Concrete_Test is Integration_Test, Lockup_Integration_Shared_Test {
-    uint256 internal streamId;
-    uint256 internal notTransferableStreamId;
-
-    function setUp() public virtual override(Integration_Test, Lockup_Integration_Shared_Test) {
-        streamId = createDefaultStream();
-        notTransferableStreamId = createDefaultStreamNotTransferable();
-
-        // Make the Recipient (owner of the NFT) the caller in this test suite.
```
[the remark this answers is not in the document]
andreivladbrg: for the first tests (`test_RevertGiven_Null`, `test_RevertGiven_PENDINGStatus` `test_RevertGiven_STREAMINGStatus` `test_RevertGiven_SETTLEDStatus` `test_RevertGiven_CANCELEDStatus`) i guess the caller will be the admin, but it is irelevant since the caller check is not made before these reverts

test/core/integration/concrete/lockup/burn/burn.t.sol
andreivladbrg: have you seen the `dev` natspec?

```solidity
/// @dev For this test, we will skip the `givenDepletedStream` modifier given the `notTransferableStreamId` is not
/// yet created.
```

the problem is that this modifier has extra logic:

```solidity
modifier givenDepletedStream(ISablierLockup lockup, uint256 streamId) {
    vm.warp({ newTimestamp: lockup.getEndTime(streamId) });
    lockup.withdrawMax({ streamId: streamId, to: lockup.getRecipient(streamId) });
    _;
}
```


and passing `notTransferableStreamId` which is not created before this function would cause to revert.

one idea is to declare this id in `Integration_Test` contract (similar to defaultStreamId) and initialize this stream in `Lockup<Model>_Integration_Shared_Test` contracts but i am not sure if we need this stream anywhere else

> Are we going to test givenDepletedStream status anywhere else?

not sure what you mean, in order to burn the NFT we need to have a a depleted status, each non-reverting test must be depleted (this modifier)

test/utils/Modifiers.t.sol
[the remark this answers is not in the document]
smol-ninja: A better structure might be this since some contract may be required in Modifiers.

```mermaid
flowchart TD
Modifiers --> Fuzzers
Fuzzers --> Constants
Fuzzers --> Utils
Base --> Modifiers
```

Because Modifiers are the contac

test/core/integration/concrete/lockup/cancel-multiple/cancelMultiple.t.sol
andreivladbrg: because these two contracts has unique logic for these multiple tests, which i couldn't find a way to nicely abstract at a lower level

are you saying that we can move them in the `Integration_Test`? i have to think about it

test/core/integration/concrete/lockup/get-asset/getAsset.t.sol
andreivladbrg: ups, mistake, it should be completely removed

test/core/unit/concrete/adminable/transfer-admin/transferAdmin.t.sol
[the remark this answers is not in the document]
smol-ninja: If we implement `Users` contract as commented deep below, we can keep modifiers name as `whenCallerRecipient`, `whenCallerSender` without having to require input arguments.

test/utils/Modifiers.t.sol
andreivladbrg: actually this was the initial version: https://github.com/sablier-labs/v2-core/commit/f5c973ce5f4d01771c015a29c5a9e4a49228bd1b

but i decided to change this, as passing the addresses in the modifiers, is more explicit IMO

test/core/integration/concrete/lockup/renounce/renounce.t.sol
```diff
@@ -76,16 +50,12 @@ abstract contract Renounce_Integration_Concrete_Test is Integration_Test, Lockup
...
-        _;
-    }
-
     function test_RevertGiven_NonCancelableStream()
         external
         whenNoDelegateCall
         givenNotNull
         givenWarmStream
-        whenCallerSender
+        whenCallerSender(users.sender)
```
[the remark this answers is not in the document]
andreivladbrg: i personally like this suggestion https://github.com/sablier-labs/v2-core/pull/1058#discussion_r1819034860

test/core/integration/concrete/lockup/set-nft-descriptor/setNFTDescriptor.t.sol
```diff
@@ -44,15 +33,15 @@ abstract contract SetNFTDescriptor_Integration_Concrete_Test is Integration_Test
         lockup.tokenURI({ tokenId: defaultStreamId });
     }
 
-    function test_WhenProvidedAddressNotMatchCurrentNFTDescriptor() external whenCallerAdmin {
+    function test_WhenProvidedAddressNotMatchCurrentNFTDescriptor() external whenCallerRecipient(users.admin) {
```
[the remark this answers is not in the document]
andreivladbrg: it is not missleading, it is just incorrect, it should be `whenCallerAdmin(users.admin)` :))

test/core/integration/concrete/lockup/burn/burn.t.sol
smol-ninja: Since we have `users.operator`, we should do this:

- In `Integration_Test`, `recipient` should call `setApprovalForAll` on `users.operator`. This will make operator an approved third party for all the tokens.
- Then we should prank to `users.operator` wherever we have `test_WhenCallerApprovedThirdParty` in the test. If we prank to alice there, then why do we have to operator in the first place.

Wdyt?

test/core/integration/concrete/lockup/transfer-from/transferFrom.t.sol
```diff
@@ -1,15 +1,14 @@
...
 
+import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
+import { IERC721 } from "@openzeppelin/contracts/token/ERC721/IERC721.sol";
+
 import { Errors } from "src/core/libraries/Errors.sol";
-import { Integration_Test } from "./../../../Integration.t.sol";
-import { Lockup_Integration_Shared_Test } from "./../../../shared/lockup/Lockup.t.sol";
 
-abstract contract TransferFrom_Integration_Concrete_Test is Integration_Test, Lockup_Integration_Shared_Test {
-    function setUp() public virtual override(Integration_Test, Lockup_Integration_Shared_Test) {
```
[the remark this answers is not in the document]
andreivladbrg: it is not reverting because the first logic executed is `_update` function (which check for transferability)

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/f96237308f6d25ef69bd4c153e5651794651011f/contracts/token/ERC721/ERC721.sol#L143-L146

IMO it is irelevant who the caller is if the first check is reverting correctly, does it make sense?

```mermaid
sequenceDiagram
    participant Caller
    participant BurnFunction

    Caller ->> BurnFunction: call burn function
    BurnFunction ->> BurnFunction: 1. Check transferability
    BurnFunction ->> Caller: 2. Check caller
    BurnFunction ->> BurnFunction: 3. Burn
```

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
```diff
@@ -37,7 +37,12 @@ abstract contract Withdraw_Integration_Concrete_Test is Integration_Test, Withdr
         lockup.withdraw({ streamId: defaultStreamId, to: users.recipient, amount: withdrawAmount });
     }
 
-    function test_RevertWhen_WithdrawalAddressZero() external whenNoDelegateCall givenNotNull givenNotDEPLETEDStatus {
+    function test_RevertWhen_WithdrawalAddressZero()
+        external
+        whenNoDelegateCall
+        givenNotNull
+        givenNotDEPLETEDStatus(defaults.START_TIME())
```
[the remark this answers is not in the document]
andreivladbrg: agree

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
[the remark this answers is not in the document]
andreivladbrg: why is this needed?
we are setting the recipient as the caller, where is the tree branch says so or it is a requirement to be

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
```diff
@@ -211,12 +223,12 @@ abstract contract Withdraw_Integration_Concrete_Test is Integration_Test, Withdr
...
         whenNoDelegateCall
         givenNotNull
-        givenNotDEPLETEDStatus
+        givenNotDEPLETEDStatus(defaults.START_TIME())
         whenWithdrawalAddressNotZero
         whenNonZeroWithdrawAmount
         whenWithdrawAmountNotOverdraw
         whenWithdrawalAddressRecipient
-        whenCallerSender
+        whenCallerSender(users.sender)
```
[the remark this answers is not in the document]
andreivladbrg: it's fine to me, but we will change it to `whenCallerIS()` modifier

test/core/integration/concrete/lockup/renounce/renounce.t.sol
andreivladbrg: removed by mistake, as there were places where this modifier didn't have this logic:

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/test/core/integration/shared/lockup/cancel.t.sol#L38-L40

test/core/integration/concrete/lockup-dynamic/create-with-durations/createWithDurations.t.sol
```diff
@@ -1,27 +1,18 @@
...
-        virtual
-        override(LockupDynamic_Integration_Concrete_Test, CreateWithDurations_Integration_Shared_Test)
-    {
-        LockupDynamic_Integration_Concrete_Test.setUp();
-        CreateWithDurations_Integration_Shared_Test.setUp();
-        streamId = lockupDynamic.nextStreamId();
+import { LockupDynamic_Integration_Shared_Test } from "../LockupDynamic.t.sol";
+
+contract CreateWithDurations_LockupDynamic_Integration_Concrete_Test is LockupDynamic_Integration_Shared_Test {
+    function setUp() public virtual override(LockupDynamic_Integration_Shared_Test) {
```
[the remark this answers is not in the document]
andreivladbrg: yes, we can remove it enteirely

test/core/integration/concrete/lockup-dynamic/create-with-timestamps/createWithTimestamps.t.sol
andreivladbrg: no, we don't

test/core/integration/concrete/lockup-dynamic/token-uri/tokenURI.t.sol
andreivladbrg: this logic is implemented:

https://github.com/sablier-labs/v2-core/blob/80de93c10a7e10a0529354c11ef8da8f3f5eeebb/test/core/integration/shared/lockup-dynamic/LockupDynamic.t.sol#L54

https://github.com/sablier-labs/v2-core/blob/80de93c10a7e10a0529354c11ef8da8f3f5eeebb/test/core/integration/shared/lockup-linear/LockupLinear.t.sol#L30

https://github.com/sablier-labs/v2-core/blob/80de93c10a7e10a0529354c11ef8da8f3f5eeebb/test/core/integration/shared/lockup-tranched/LockupTranched.t.sol#L53

test/core/integration/concrete/lockup/burn/burn.t.sol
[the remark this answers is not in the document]
smol-ninja: Fair point and I agree with you.

test/core/integration/concrete/lockup/burn/burn.t.sol
smol-ninja: > have you seen the dev natspec?

Yes I have.

> one idea is to declare this id in Integration_Test contract (similar to defaultStreamId) and initialize this stream

How about creating `notTransferableStreamId` in `setUp` of this contract? Can that fix it?

> not sure what you mean, in order to burn the NFT we need to have a a depleted status, each non-reverting test must be depleted

Yes so we should think of including `givenDepletedStream` because we need that in order to call `burn`.

test/core/integration/concrete/lockup/cancel-multiple/cancelMultiple.t.sol
smol-ninja: I am saying can we not do this?

```solidity
abstract contract CancelMultiple_Integration_Concrete_Test is CancelMultiple_Integration_Shared_Test {}
```

since `CancelMultiple_Integration_Shared_Test` already inherits from `Integration_Test`.

test/utils/Modifiers.t.sol
smol-ninja: I see. Thanks for sharing the commit reference. 

> but i decided to change this, as passing the addresses in the modifiers, is more explicit IMO

It depends on the naming convention. When the modifier name is `whenCallerAdmin` and there is only one admin user declared as `users.admin`, there should not be any need to pass admin inside `whenCallerAdmin` when modifier name is explicit enough.

Why do you not agree with my suggestion? It would let us keep the modifier names as `whenCallerAdmin` and `whenCallerSender` without us requiring to pass input arguments.

test/core/integration/concrete/lockup/renounce/renounce.t.sol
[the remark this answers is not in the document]
smol-ninja: Curious why you prefer modifier names with argument over modifier names without argument? I think modifier names without argument are friendlier because the names are explicit enough.

test/core/integration/concrete/lockup/transfer-from/transferFrom.t.sol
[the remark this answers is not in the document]
smol-ninja: This is an example of black box vs white box testing. You could argue that because you know the internal logic of `transferFrom`. On the other hand, without knowing the internal logic of `transferFrom` but just the natspecs of it, how would you have tested it otherwise?

IMO in that case, the first intuitive check would be to see who the caller is. And therefore it would become relevant who is the caller even before the first check. I would recommend to use black box testing approach where we test based on how the function should behave without getting so much into the internal logic. Having said that, the caller should be recipient for the whole test suite.

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
```diff
@@ -147,9 +154,14 @@ abstract contract Withdraw_Integration_Concrete_Test is Integration_Test, Withdr
 
         // It should emit {WithdrawFromLockupStream} and {MetadataUpdate} events.
         vm.expectEmit({ emitter: address(lockup) });
-        emit WithdrawFromLockupStream({ streamId: defaultStreamId, to: users.alice, asset: dai, amount: withdrawAmount });
+        emit ISablierLockup.WithdrawFromLockupStream({
```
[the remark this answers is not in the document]
smol-ninja: IIUC, as per BTT, if the name of the test is `test_WhenCallerApprovedThirdPartyOrRecipient` there should be no need to have `whenWithdrawalAddressNotRecipient` modifier. cc @PaulRBerg

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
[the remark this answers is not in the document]
smol-ninja: because recipient is the main actor for withdraw. Same argument as https://github.com/sablier-labs/v2-core/pull/1058#discussion_r1819828802

test/core/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.t.sol
smol-ninja: Lets bring it back (https://github.com/sablier-labs/v2-core/pull/1058#discussion_r1819823111). I vote for modifiers without arguments. The modifier have very good names and therefore are very explicit about what they do inside. Arguments are unnecessary imo.

test/core/integration/concrete/lockup-dynamic/token-uri/tokenURI.t.sol
smol-ninja: Got it.

test/core/integration/concrete/lockup/burn/burn.t.sol
andreivladbrg: we can't do this in `Integration_Test` as the LD/LL/LT are not yet deployed there, it should be

it should be in `LockupDynamic_Integration_Shared_Test` `LockupLinear_Integration_Shared_Test` `LockupTranched_Integration_Shared_Test` if we want to do this

test/core/integration/concrete/lockup/burn/burn.t.sol
andreivladbrg: > How about creating notTransferableStreamId in setUp of this contract? Can that fix it?

we can't create streams in `setUp` as the LD/LL/LT are not yet deployed at its level

```mermaid
flowchart TD
    FunctionName_LockupModel_Integration_TypeTest --> LockupModel_Integration_Shared_Test
    FunctionName_LockupModel_Integration_TypeTest --> FunctionName_Integration_TypeTest
    LockupModel_Integration_Shared_Test --> Integration_Test
    FunctionName_Integration_TypeTest --> Integration_Test
    Integration_Test --> Base_Test
    Base_Test --> Utils

    %% Adding a subgraph as a note
    subgraph note [ ]
        direction TB
        note_text(["Here the contracts still don't exist"])
    end

    note -.-> FunctionName_Integration_TypeTest
```

the tests are run in the child contracts, so for them it is not problem

test/core/integration/concrete/lockup/transfer-from/transferFrom.t.sol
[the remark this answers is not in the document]
andreivladbrg: since we are talking about only 1 test, i will prank the caller directly in the test function
and we can furthur discuss about black box vs white box testing later so that we can come up with a architecture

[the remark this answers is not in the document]
andreivladbrg: btw, contradictory comment: https://github.com/sablier-labs/v2-core/pull/1058#discussion_r1819812368

test/utils/Defaults.sol
```diff
@@ -94,6 +94,10 @@ contract Defaults is Constants, Merkle {
                                       HELPERS
     //////////////////////////////////////////////////////////////////////////*/
 
+    function getUsers() public view returns (Users memory) {
```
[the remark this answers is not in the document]
andreivladbrg: removed it completely

test/core/unit/concrete/adminable/transfer-admin/transferAdmin.t.sol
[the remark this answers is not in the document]
andreivladbrg: ended up reintroducing the defaults and users in Modifiers

test/utils/Modifiers.t.sol
andreivladbrg: re introduced the previous version

andreivladbrg:
> Even after closing all the tabs, the browser was working slow because of large number of changes. This is a good argument in favour of why we should split large PRs like this into multiple PRs so that each PR has changes in not more than 50 files. An idea could be to have one PR for core test, another PR for periphery test etc.

I’m so sorry to hear the review process was difficult for you. I agree that breaking down work into multiple PRs can make reviews smoother (remember the package tethering PRs?). In this case, however, the process was challenging for me as well. Unlike the package tethering, where I had a clear understanding of the necessary changes from the outset, this time, every adjustment seemed to cause something else to break. 
In the future, I’ll do my best to make the review process easier. This experience was tough for both of us and highlights how exhaustive the test structure really was (and how needed this change was).

---

@smol-ninja I reintroduced the `defaults` and `users` variables in the `Modifiers` contract, which should address most of your feedback. Could you please take another look at my latest commit 91f4d5039b08840194fd6459b36310f6e72004e6 ?

PaulRBerg:
Also: Loom videos for explaining stuff in large PRs would be helpful.

test/core/integration/concrete/lockup/burn/burn.t.sol
smol-ninja: Why cant we do it in `Integration_Test.setUp()` or `Base_Test.setUp`? The lockup contracts are deployed in Base setup through `deployProtocolConditionally` so I wonder why we cant do this.

test/core/integration/concrete/lockup/burn/burn.t.sol
smol-ninja: I didn't understand. Are you saying we cannot initialize `notTransferableStreamId` in `Burn_Integration_Concrete_Test.setUp()`? Ain't LD/LL/LT deployed in Base test?

test/core/integration/concrete/lockup/transfer-from/transferFrom.t.sol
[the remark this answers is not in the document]
smol-ninja: How is it contradictory?

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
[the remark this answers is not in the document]
smol-ninja: Agree to move fast. Will be closing this thread.

test/utils/Modifiers.t.sol
smol-ninja: Nice thanks.

test/core/integration/concrete/lockup/withdraw/withdraw.t.sol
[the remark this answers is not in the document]
smol-ninja: This was an error on my part. `whenWithdrawalAddressNotRecipient` is about the withdrawal address whereas `test_WhenCallerApprovedThirdPartyOrRecipient` is about the caller. So they both can be present in the same test.

test/core/integration/concrete/lockup/burn/burn.t.sol
smol-ninja: This is what I was talking about: https://github.com/sablier-labs/v2-core/pull/1058/commits/7cc0f9dbbeb5f2608c7440b572769339759e0433

test/core/integration/concrete/lockup/burn/burn.t.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1058/commits/7cc0f9dbbeb5f2608c7440b572769339759e0433

test/core/integration/concrete/lockup/burn/burn.t.sol
andreivladbrg: yes it looks good, i was trying to use `lockup.setApprovalForAll` which doesn't work

andreivladbrg: but was there problem in keeping them in Lockup<Model>_Integration_Shared_Test

smol-ninja: Yes there is a small problem which is since we initialize all users including `operator` in the Base contract, `setApprovalForAll` permission is a part of setting up `operator` user. The operator is not an operator until `setApprovalForAll` is called by the recipient. Therefore, its better to have it next to users declaration than in the `Lockup<Model>_Integration_Shared_Test`.

smol-ninja (changes requested):
I re-reviewed the core tests again and discovered a few more areas where changes may be required. I also found redundant codes which I removed in https://github.com/sablier-labs/v2-core/pull/1058/commits/f28239bd268d0116cd84816cf454fcdd56e63777. The commit includes some changes related to coding pattern. Let me know if you have any question about it.

I will be reviewing periphery now.

test/periphery/integration/merkle-campaign/shared/clawback/clawback.t.sol
```diff
@@ -21,6 +18,15 @@ abstract contract Clawback_Integration_Test is MerkleCampaign_Integration_Shared
         test_Clawback(users.campaignOwner);
     }
 
+    modifier whenFirstClaimMade() {
```
smol-ninja, line 21: What do you think about the following:
1. Move this modifier into `Modifiers`
2. Define a virtual `claim` function in `Modifiers`
3. Override `claim` through `MerkleCampaign_Integration_Test`

This will allow us to move all modifiers into `Modifiers`. Having a virtual `claim` function in Modifiers shouldn't be an issue. We already use similar technique in Core tests with create stream functions.

test/periphery/integration/merkle-campaign/lt/MerkleLT.t.sol
```diff
@@ -26,40 +26,35 @@ abstract contract MerkleLT_Integration_Shared_Test is MerkleCampaign_Integration
                                 SHARED TESTS
 //////////////////////////////////////////////////////////////////////////*/
 
-contract Clawback_MerkleLT_Integration_Test is Clawback_Integration_Test, MerkleLT_Integration_Shared_Test {
-    function setUp() public override(Clawback_Integration_Test, MerkleLT_Integration_Shared_Test) {
-        Clawback_Integration_Test.setUp();
+contract Clawback_MerkleLT_Integration_Test is MerkleLT_Integration_Shared_Test, Clawback_Integration_Test {
+    function setUp() public override(MerkleLT_Integration_Shared_Test, MerkleCampaign_Integration_Test) {
```
smol-ninja, line 30: I remember asking `setUp` functions from `Clawback_Integration_Test` once and you told me (I may be wrong about this memory) that it provides a better readability to have explicit overrides. And I agreed so I'd like to make the same suggestion. 

1. bring base the `setUp` in `Clawback_Integration_Test`

then

```solidity
contract Clawback_MerkleLT_Integration_Test is MerkleLT_Integration_Shared_Test, Clawback_Integration_Test {
    function setUp() public override(MerkleLT_Integration_Shared_Test, Clawback_Integration_Test) {
        MerkleLT_Integration_Shared_Test.setUp();
    }
}
```

Otherwise it creates some ambiguity as to why it overrides `MerkleCampaign_Integration_Test` whereas inheriting from `Clawback_Integration_Test`

test/periphery/integration/merkle-campaign/lt/claim/claim.t.sol
```diff
@@ -4,19 +4,16 @@ pragma solidity >=0.8.22 <0.9.0;
...
+import { MerkleLT_Integration_Shared_Test, MerkleCampaign_Integration_Test } from "../MerkleLT.t.sol";
 
 contract Claim_MerkleLT_Integration_Test is Claim_Integration_Test, MerkleLT_Integration_Shared_Test {
-    function setUp() public override(Claim_Integration_Test, MerkleLT_Integration_Shared_Test) {
-        super.setUp();
-    }
-
-    modifier whenTotalPercentageNot100() {
-        _;
+    function setUp() public virtual override(MerkleLT_Integration_Shared_Test, MerkleCampaign_Integration_Test) {
```
smol-ninja, line 15: Same argument about overriding contracts and all other similar places.

smol-ninja (changes requested):
Here is periphery review.

test/core/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -42,7 +37,8 @@ contract StreamedAmountOf_LockupLinear_Integration_Concrete_Test is
         assertEq(actualStreamedAmount, expectedStreamedAmount, "streamedAmount");
     }
 
-    function test_GivenCliffTimeInFuture() external view givenSTREAMINGStatus {
+    function test_GivenCliffTimeInFuture() external givenSTREAMINGStatus {
+        vm.warp({ newTimestamp: defaults.CLIFF_TIME() - 1 });
```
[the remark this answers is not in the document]
andreivladbrg: well, the test name is: `test_GivenCliffTimeInFuture` 
which means we need to warp before the cliff time, so that the streamed amount is 0

we can indeed remove the warp from the `test_GivenEndTimeInFuture` test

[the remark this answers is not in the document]
smol-ninja: Thanks for explaining. I get it now. Previously, `givenSTREAMINGStatus` didn't warping anything inside and so by default the current time was set to the `START_TIME` and therefore it was correct. Now `givenSTREAMINGStatus` warps to 26% which is well beyond cliff time and therefore, you have to warp it to `defaults.CLIFF_TIME() - 1`.

Feel free to mark it resolve then.

test/core/integration/concrete/lockup-linear/token-uri/tokenURI.t.sol
```diff
@@ -35,16 +34,12 @@ contract TokenURI_LockupLinear_Integration_Concrete_Test is LockupLinear_Integra
...
-        defaultStreamId = createDefaultStream();
-        vm.warp({ newTimestamp: defaults.START_TIME() + defaults.TOTAL_DURATION() / 4 });
-        _;
-    }
-
     /// @dev If you need to update the hard-coded token URI:
     /// 1. Use "vm.writeFile" to log the strings to a file.
     /// 2. Remember to escape the EOL character \n with \\n.
     function test_WhenTokenURIDecoded() external skipOnMismatch givenNFTExists {
+        vm.warp({ newTimestamp: defaults.START_TIME() + defaults.TOTAL_DURATION() / 4 });
```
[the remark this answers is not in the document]
andreivladbrg: actually it should be to both, mistake

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/test/core/integration/concrete/lockup-linear/token-uri/tokenURI.t.sol#L38-L42

test/core/integration/concrete/lockup-linear/withdrawable-amount-of/withdrawableAmountOf.t.sol
```diff
@@ -3,32 +3,23 @@ pragma solidity >=0.8.22 <0.9.0;
...
-        override(LockupLinear_Integration_Concrete_Test, WithdrawableAmountOf_Integration_Concrete_Test)
-    {
-        LockupLinear_Integration_Concrete_Test.setUp();
-        WithdrawableAmountOf_Integration_Concrete_Test.setUp();
+    function setUp() public virtual override(LockupLinear_Integration_Shared_Test, Integration_Test) {
+        LockupLinear_Integration_Shared_Test.setUp();
     }
 
-    function test_GivenCliffTimeInFuture() external view givenSTREAMINGStatus {
+    function test_GivenCliffTimeInFuture() external givenSTREAMINGStatus {
```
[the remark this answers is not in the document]
andreivladbrg: the "streaming" status is when the start time is greater than `block.timestamp`

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/src/core/abstracts/SablierLockup.sol#L495-L503

> What is the status of a LL stream before the cliff time?

depends on the `block.timestamp`

---

since this a different discussion (not related to this PR) i'd suggest to  open a different discussion for this

test/utils/Modifiers.sol
```diff
@@ -0,0 +1,543 @@
...
+        _;
+    }
+
+    modifier givenNotDepletedStream() {
+        _;
+    }
+
+    modifier givenDepletedStream(ISablierLockup lockup, uint256 streamId) {
+        vm.warp({ newTimestamp: lockup.getEndTime(streamId) });
+        lockup.withdrawMax({ streamId: streamId, to: lockup.getRecipient(streamId) });
```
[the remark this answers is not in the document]
andreivladbrg: forgot to make this update after introducing the defaults here 👍🏻

test/core/integration/concrete/lockup/burn/burn.t.sol
```diff
@@ -2,177 +2,136 @@
...
+        // Burn the NFT so that it no longer exists.
+        lockup.burn(defaultStreamId);
+
+        // Run the test.
+        vm.expectRevert(abi.encodeWithSelector(IERC721Errors.ERC721NonexistentToken.selector, defaultStreamId));
+        lockup.burn(defaultStreamId);
     }
 
-    function test_WhenCallerApprovedThirdParty()
+    /// @dev For this test, we will skip the `givenDepletedStream` modifier given the `notTransferableStreamId` is not
```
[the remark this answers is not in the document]
andreivladbrg: the modifier, yes

test/core/integration/concrete/lockup-linear/token-uri/tokenURI.t.sol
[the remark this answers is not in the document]
smol-ninja: Got it. And why do we need those warps?

test/core/integration/concrete/lockup/cancel/cancel.tree
```diff
@@ -13,12 +13,12 @@ Cancel_Integration_Concrete_Test
       │  └── given SETTLED status
       │     └── it should revert
       └── given warm stream
-         ├── when unauthorized caller
+         ├── when caller not sender
```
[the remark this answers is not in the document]
andreivladbrg: good catch

test/core/integration/concrete/lockup-linear/withdrawable-amount-of/withdrawableAmountOf.t.sol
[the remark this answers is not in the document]
smol-ninja: I see. The status is then `STREAMING` as long as `block.timestamp > startTime`. Then we should change the natspec over `modifier givenSTREAMINGStatus`. The line "In LockupLinear, the streaming starts after the cliff time" is confusing.

> since this a different discussion

How is this a different discussion?

test/core/integration/concrete/lockup/cancel/cancel.tree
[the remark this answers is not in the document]
smol-ninja: Thank you. I reviewed this PR with all my heart and patience.

test/core/integration/fuzz/lockup/withdrawMaxAndTransfer.t.sol
```diff
@@ -52,7 +48,9 @@ abstract contract WithdrawMaxAndTransfer_Integration_Fuzz_Test is
 
         // Expect the relevant event to be emitted.
         vm.expectEmit({ emitter: address(lockup) });
-        emit Transfer({ from: users.recipient, to: newRecipient, tokenId: defaultStreamId });
+        emit IERC721.Transfer({ from: users.recipient, to: newRecipient, tokenId: defaultStreamId });
+
+        resetPrank({ msgSender: users.recipient });
```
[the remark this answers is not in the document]
andreivladbrg: sounds good

test/core/integration/concrete/lockup-linear/withdrawable-amount-of/withdrawableAmountOf.t.sol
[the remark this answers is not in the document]
andreivladbrg: > The status is then STREAMING as long as block.timestamp > startTime. Then we should change the natspec over modifier givenSTREAMINGStatus. The line "In LockupLinear, the streaming starts after the cliff time" is confusing.

i see, i just copied the modifier (including the natspec) here, i didn't read the natspec tbh when moving it

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/test/core/integration/shared/lockup/cancel.t.sol#L30-L36

fyi, it was written by you :))

image


will remove it

> How is this a different discussion?

i tought you are saying to change the `statusOf` function, that's why i said it is a diff discussion

test/core/integration/concrete/lockup-linear/token-uri/tokenURI.t.sol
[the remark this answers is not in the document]
andreivladbrg: the initial idea was to have the progress bar with percentage > 0
but these test are not ran for so long due to `skipOnMismatch` modifier :))

test/utils/Modifiers.sol
```diff
@@ -0,0 +1,543 @@
...
+    modifier givenPreviousWithdrawal() {
+        _;
+    }
+
+    modifier givenRecipientAllowedToHook() {
+        _;
+    }
+
+    modifier givenStartTimeInPast() {
+        _;
```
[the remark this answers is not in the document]
andreivladbrg: if we add this warp in the modifier, it wouldn't be relevant, as we have in each function that it's used a call to `warp` (most of them warp to a different timestamp - so we can't generalize it here)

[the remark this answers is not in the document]
smol-ninja: Agree

test/core/integration/concrete/lockup/cancel/cancel.tree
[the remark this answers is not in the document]
andreivladbrg: appreciate it
https://tenor.com/view/jay-z-tip-hat-stare-gif-1327869528750982278

test/periphery/integration/merkle-campaign/shared/clawback/clawback.t.sol
andreivladbrg: it’s an interesting idea; however, since this is the only place where this modifier is being used, i would keep it here.

> Modifiers shouldn't be an issue. We already use similar technique in Core tests with create stream functions

this isn’t quite the same thing, as in core we have create virtual functions that don’t exist in the base contract (`SablierLockup`). here, `claim` is actually declared in the merkle base contract

test/core/integration/concrete/lockup/renounce/renounce.t.sol
```diff
@@ -57,16 +41,6 @@ abstract contract Renounce_Integration_Concrete_Test is Integration_Test, Lockup
         lockup.renounce(defaultStreamId);
     }
 
-    /// @dev This modifier runs the test twice: once with a "PENDING" status, and once with a "STREAMING" status.
-    modifier givenWarmStream() {
```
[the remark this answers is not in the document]
andreivladbrg: the problem is that i couldn't find an elegant way to fix this. 

overriding it doesn't work.

the only idea i have is to redeclare it in these test file

test/periphery/integration/merkle-campaign/lt/MerkleLT.t.sol
andreivladbrg: i don’t remember exactly what you’re referring to, sorry. (we didn't have a shared pattern in periphery before your claim PR)

in this PR, we’ve changed how things work, so there’s no point in adding `setUp` functions

test/core/integration/concrete/lockup-linear/token-uri/tokenURI.t.sol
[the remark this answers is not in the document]
andreivladbrg: https://github.com/sablier-labs/v2-core/issues/1071

test/periphery/integration/merkle-campaign/shared/clawback/clawback.t.sol
smol-ninja: - Those virtual functions in core that don't exist in base contract. And then they are declared in LL/LD/LT test contracts.
- `claim` does not exist in the base contract. And `claim` is declared in MerkleCampaign contract. 

How are these two not same things? Also even if its just one Modifier, why make it an exception when we can keep all of them in the same place called `Modifiers`?

test/periphery/integration/merkle-campaign/lt/MerkleLT.t.sol
smol-ninja: Wouldn't it be confusing to do it this way?

```solidity
contract Clawback_MerkleLT_Integration_Test is MerkleLT_Integration_Shared_Test, Clawback_Integration_Test {
    function setUp() public override(MerkleLT_Integration_Shared_Test, MerkleCampaign_Integration_Test) {
        MerkleLT_Integration_Shared_Test.setUp();
    }
}
```

vs 

```solidity
contract Clawback_MerkleLT_Integration_Test is MerkleLT_Integration_Shared_Test, Clawback_Integration_Test {
    function setUp() public override(MerkleLT_Integration_Shared_Test, Clawback_Integration_Test) {
        MerkleLT_Integration_Shared_Test.setUp();
    }
}
```

Second one explicitly overrides the inherited contract whereas first one overrides the parent contract. To be able to understand if overrides are accurate, I had to go inside the `Clawback_Integration_Test` to see if all is OK.

test/core/integration/concrete/lockup/renounce/renounce.t.sol
```diff
@@ -41,7 +41,15 @@ abstract contract Renounce_Integration_Concrete_Test is Integration_Test {
         lockup.renounce(defaultStreamId);
     }
 
-    function test_RevertWhen_CallerNotSender() external whenNoDelegateCall givenNotNull givenWarmStream {
+    modifier givenWarmStreamRenounce() {
```
smol-ninja, line 44: Why put it here? Similar to how we used to have all events in `Events`, errors in `Errors`, and similar pattern, why not just move all the modifiers into one Modifiers file?

test/core/integration/concrete/lockup/renounce/renounce.t.sol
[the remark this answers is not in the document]
smol-ninja: I have seen your recent commit. Looks good to me. But why keep it in this file instead of placing the new modifier in the `Modifiers` contract.

test/periphery/integration/merkle-campaign/lt/MerkleLT.t.sol
smol-ninja: After thinking about it, I agree with your changes. So consider it resolved from my side.

andreivladbrg: > After thinking about it, I agree with your changes. So consider it resolved from my side.

great, thank you

test/periphery/integration/merkle-campaign/shared/clawback/clawback.t.sol
andreivladbrg: > * Those virtual functions in core that don't exist in base contract. And then they are declared in LL/LD/LT test contracts

they are declared as _virtual_ in the `Integration_Test` which +/- the same thing with `Base_Test`

> How are these two not same things? 

`LL is BaseLockup`

`create` functions do not exist in `BaseLockup` but is fully implemented `LL`

`MerkleLL is MerkleBase`
 
`claim` function exists in `MerkleBase` and it just partially implemented in `MerkleLL` (as an internal function

so that we can create a default stream in `lockup` (shared dir) we "forced" to declare the function as virtual (as we can't call `lockup.create`

> There are a few more modifiers that are also only used once yet we have them in the Modifiers contract

most likely they don't have unique logic and are empty modifiers, if you have strong argument, feel free to push this change

test/core/integration/concrete/lockup/renounce/renounce.t.sol
andreivladbrg: https://github.com/sablier-labs/v2-core/pull/1058#discussion_r1824241668

i don't understand if it looks good or not

test/periphery/integration/merkle-campaign/shared/clawback/clawback.t.sol
smol-ninja: Lets move forward with your approach then. My main motivation was to move "all" modifiers into one contract regardless of what logic they contain inside.

test/core/integration/concrete/lockup/renounce/renounce.t.sol
smol-ninja: It looks fine. I couldn't think of a way to keep `givenWarmStream` here because the logic is different. I suppose the branches need some improvisation here. We can look into it in the future. Lets move forward now.

smol-ninja (approved):
LGTM

andreivladbrg:
It was journey, thank you again Shub🚀

PR #1070, opened 2024-10-23
feat: add batch function 

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/1053


smol-ninja:
Thank you for the PR Andrei. 

I'd like to make a small suggestion. When a PR depends on another PR, can we please have the base branch set as the branch from another PR? This makes it easier to review a child PR asynchronously. I know that by changing the URL I can see the diff but I use the "viewed checkbox" feature in Github UI to track which files have been reviewed and which are left to be reviewed. 

To be more explicit, In the description, it can be mentioned that this PR requires this another PR to be merged first.

Does that make sense?

test/core/integration/concrete/lockup/batch/batch.t.sol
```diff
@@ -0,0 +1,41 @@
...
+import { Errors } from "src/core/libraries/Errors.sol";
+
+import { Integration_Test } from "./../../../Integration.t.sol";
+
+abstract contract Batch_Integration_Concrete_Test is Integration_Test {
+    function test_RevertWhen_CallFunctionNotExist() external {
+        bytes[] memory calls = new bytes[](1);
+        calls[0] = abi.encodeCall(dai.balanceOf, (users.recipient));
+
+        Batch(address(lockup)).batch(calls);
```
smol-ninja, line 14: Missting `expectRevert` statement.

src/core/abstracts/Batch.sol
```diff
@@ -0,0 +1,26 @@
...
+/// @notice This contract implements logic to batch call any function.
+/// @dev Forked from: https://github.com/boringcrypto/BoringSolidity/blob/master/contracts/BoringBatchable.sol
+abstract contract Batch {
+    /*//////////////////////////////////////////////////////////////////////////
+                         USER-FACING NON-CONSTANT FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Allows batched call to self, `this` contract.
+    /// @param calls An array of inputs for each call.
+    function batch(bytes[] calldata calls) external {
```
smol-ninja, line 16: This function should also be a part of `ISablierLockup` interface. We forgot to do the same with Flow.

test/core/integration/concrete/lockup/batch/batch.t.sol
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    function test_WhenDataValid() external whenCallFunctionExists {
+        resetPrank({ msgSender: users.sender });
+
+        bytes[] memory calls = new bytes[](1);
+        calls[0] = abi.encodeCall(lockup.cancel, (defaultStreamId));
+
+        Batch(address(lockup)).batch(calls);
+        assertTrue(lockup.wasCanceled(defaultStreamId));
```
smol-ninja, line 39: Should we put a `assertFalse(lockup.wasCanceled(defaultStreamId));` before calling `batch()` to verify that it was indeed false before the call.

test/core/integration/concrete/lockup/batch/batch.t.sol
```diff
@@ -0,0 +1,41 @@
...
+        Batch(address(lockup)).batch(calls);
+    }
+
+    function test_WhenDataValid() external whenCallFunctionExists {
+        resetPrank({ msgSender: users.sender });
+
+        bytes[] memory calls = new bytes[](1);
+        calls[0] = abi.encodeCall(lockup.cancel, (defaultStreamId));
+
+        Batch(address(lockup)).batch(calls);
```
smol-ninja, line 38: This can be changed to interface if we put `batch` in the interface file of Lockup.

smol-ninja (changes requested):
Good work on this PR. Left a few comments below:

src/core/abstracts/Batch.sol
andreivladbrg: agree

smol-ninja: That works as well.

test/core/integration/concrete/lockup/batch/batch.t.sol
andreivladbrg: hmm, curious why the test isn't failing then, i guess the branch should say `it should do nothing`

andreivladbrg:
@smol-ninja just pushed a commit to addres your feedback, lmk if it looks good

test/core/integration/concrete/lockup/batch/batch.t.sol
smol-ninja: Did you figure out why its not failing?

smol-ninja: It did not revert because `dai.balanceOf` encodes to `balanceOf(address)` selector which is a valid function in `ERC721` contract.  I fixed it in https://github.com/sablier-labs/v2-core/pull/1070/commits/4865b3c15a340f314afdf53396c3d5f9cc51e55a.

-----------

I also want to mention that because the test was not failing, changing branch to `it should do nothing should not be a solution. This is an example of white box testing where you tried to curve fit the test to made them pass. Referring to my discussion on [white box testing vs black box testing](https://github.com/sablier-labs/flow/discussions/311), in a black box testing, we wont change the test but rather investigate the internal logic in such cases.

I look forward to see your comments on my discussion.

smol-ninja (approved):
Feel free to merge it after merging https://github.com/sablier-labs/v2-core/pull/1058.

PR #1075, opened 2024-11-04
feat: add unlock amounts functionality

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/1043 and https://github.com/sablier-labs/v2-core/issues/1067

Since we have the invariant: $`\text{cliff time} = 0 \implies \text{cliff unlock amount} = 0`$, we would have these 6 possible scenarios for the streamed amount function: 

| Scenario | Cliff Time | Start Unlock Amount | Cliff Unlock Amount | Shape                                      |
|----------|------------|---------------------|---------------------|--------------------------------------------|
| 1        | No         | No                  | No                  | [Simple Linear](https://docs.sablier.com/concepts/protocol/stream-types#lockup-linear)                              |
| 2        | Yes        | No                  | No                  | Constant, then linear (see below)                 |
| 3        | No         | Yes                 | No                  | [Unlock Linear](https://docs.sablier.com/concepts/protocol/stream-types#unlock-linear)                              |
| 4        | Yes        | Yes                 | No                  | Initial amount, constant, then linear (see below)   |
| 5        | Yes        | No                  | Yes                 | Constant, then unlock with [linear function](https://docs.sablier.com/concepts/protocol/stream-types#cliffs) |
| 6        | Yes        | Yes                 | Yes                 | [Unlock Cliff](https://docs.sablier.com/concepts/protocol/stream-types#unlock-cliff)                               |

<details><summary>Plot graphs with docs code</summary>
<p>
    
### Scenario 2

image

#### Code to update the docs shapes with:

```mdx
<FunctionPlot
  options={{
    data: [
      { fn: "0", range: [0, 25], color: "#f77423" },
      { fn: "(4/3) * (x - 25)", range: [25, 100], color: "#f77423" },
      {
        points: [
          [0, 0],
          [25, 0],
        ],
        fnType: "points",
        graphType: "polyline",
        color: "#f77423",
      },
    ],
  }}
/>

```
    
### Scenario 4

image

#### Code to update the docs shapes with:

```mdx
<FunctionPlot
  options={{
    data: [
      { fn: "25", range: [0, 50], color: "#f77423" }, // Constant at y=25 from x=0 to x=50
      { fn: "1.5 * (x - 50) + 25", range: [50, 100], color: "#f77423" }, // Linear rise with slope 1.5 starting from y=25 at x=50
      {
        points: [
          [0, 25],
          [50, 25],
        ],
        fnType: "points",
        graphType: "polyline",
        color: "#f77423",
      },
    ],
  }}
/>

```
</p>
</details> 

---

**Note:** Since the mathematical function has changed in `SablierLockupLinear._calculateStreamedAmount`, in order to achieve the previous [cliff linear shape](https://docs.sablier.com/concepts/protocol/stream-types#cliffs), it is required to pass an `unlockAmount.cliff > 0`; otherwise, the second scenario shape will be achieved. This is also the reason I needed to change the tests from warping to `WARP_26_PERCENT` instead of `CLIFF_TIME,` as many tests were failing. The cliff amount has a remainder of 2534 due to a different "streamable" range.

As you will see in the code, I am referring to the range that calculates the elapsed time as "streamable," which might be confusing since the entire function is called "streamed amount." However, we will eventually change this terminology to "vesting," so don’t mind the current terminology if you find it unclear.

---

Below you can see an useful diagram to vizualize while reviewing the new function.

<details><summary>New function flow</summary>
<p>

```mermaid
flowchart TD
    start_time_check([Is startTime >= current block timestamp?]):::check
    return_zero([Return 0]):::return
    end_time_check([Is current block timestamp >= endTime?]):::check
    return_deposited([Return depositedAmount]):::return
    cliff_time_check([Is cliffTime set and cliffTime > current block timestamp?]):::check
    return_start_unlock([Return start unlock amount]):::return
    unlock_sum_check([Is unlockAmountsSum >= depositedAmount?]):::check
    elapsed_calculation_check([Calculate elapsedTime and streamableDuration]):::calculation
    etp_calculation([Calculate elapsedTimePercentage]):::calculation
    streamable_amount_calc([Calculate streamableAmount]):::calculation
    streamed_amount_calc([Calculate streamedAmount]):::calculation
    final_check([Is streamedAmount > depositedAmount?]):::check
    return_withdrawn([Return withdrawn amount]):::return
    return_streamed([Return streamedAmount]):::return

    start_time_check -- "Yes" --> return_zero
    start_time_check -- "No" --> end_time_check
    end_time_check -- "Yes" --> return_deposited
    end_time_check -- "No" --> cliff_time_check
    cliff_time_check -- "Yes" --> return_start_unlock
    cliff_time_check -- "No" --> unlock_sum_check
    unlock_sum_check -- "Yes" --> return_deposited
    unlock_sum_check -- "No" --> elapsed_calculation_check

    elapsed_calculation_check --> etp_calculation
    etp_calculation --> streamable_amount_calc
    streamable_amount_calc --> streamed_amount_calc
    streamed_amount_calc --> final_check
    final_check -- "Yes" --> return_withdrawn
    final_check -- "No" --> return_streamed

    classDef check fill:#FFC107,stroke:#333,stroke-width:2px;
    classDef return fill:#4CAF50,stroke:#333,stroke-width:2px;
    classDef calculation fill:#03A9F4,stroke:#333,stroke-width:2px;
```

</p>
</details>


andreivladbrg:
@smol-ninja the PR is ready for your review, I have updated the OP with some useful comments, please get first through them before reviewing,

andreivladbrg:
@smol-ninja just rebased from `refactor/singleton-contract`. in case i have missed something, i have created a [recovery branch](https://github.com/sablier-labs/v2-core/tree/temp/unlock-amounts). 

**Note:** the size of the contract is at its limit, adding new things to the contract would require us to decrease the optimizer runs.

| Contract                      | Size (B) | Margin (B) |
| -----| ----| ----| 
 | SablierLockup                 |   24,450 |        126 |

smol-ninja:
That was a very fast rebase @andreivladbrg. I will only review the `src` since the tests will require another rebase from [#1077](https://github.com/sablier-labs/v2-core/pull/1077) and then a re-review. Does that sound good?

andreivladbrg:
> That was a very fast rebase @andreivladbrg. I will only review the `src` since the tests will require another rebase from [#1077](https://github.com/sablier-labs/v2-core/pull/1077) and then a re-review. Does that sound good?

but that rebase is on you ;) as discussed on slack :)

smol-ninja:
> but that rebase is on you (in the singleton test PR) as discussed on slack

Thats fine. The concern is that in this PR, src are compatible with singleton but tests are compatible with previous structure. So, src and tests no longer connect to each other. I think a possible ways to go forward is:

1. Create the previous unlock PR against staging
2. Review and merge it into staging - all tests passed
3. Rebase #1069 from staging i.e. and fix conflict
4. Rebase https://github.com/sablier-labs/v2-core/pull/1077 from #1069 and fix conflict.

Or we can choose what we discussed on slack. I am fine with anything but reviewing tests in this PR against the refactored `src` would be not be the best use of time.

andreivladbrg:
as discussed on slack, we will merge #1077 in #1069 and after we will rebase from 1069 here, and then merge everything in `staging`

smol-ninja:
As decided on Slack, we will take the following approach:
1. Review `src` diff of this PR (by me)
2. Review and merge #1077 into #1069 
3. Review and merge #1069 into staging
4. Rebase this PR from staging (by me)
5. Review and merge this PR into staging

src/core/SablierLockup.sol
```diff
@@ -108,6 +111,17 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
...
 
+    /// @inheritdoc ISablierLockup
+    function getUnlockAmounts(uint256 streamId)
+        external
+        view
+        override
+        notNull(streamId)
+        returns (LockupLinear.UnlockAmounts memory unlockAmounts)
+    {
+        unlockAmounts = _unlockAmounts[streamId];
```
smol-ninja, line 122: It should revert if the model is not Lockup Linear.

src/core/SablierLockup.sol
```diff
@@ -236,15 +252,16 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
     /// @inheritdoc ISablierLockup
     function createWithTimestampsLL(
         Lockup.CreateWithTimestamps calldata params,
-        uint40 cliffTime
+        uint40 cliffTime,
+        LockupLinear.UnlockAmounts calldata unlockAmounts
```
smol-ninja, line 256: Lets change the order of input parameters similar to other functions:
1. Lockup.CreateWithTimestamps calldata params
2. LockupLinear.UnlockAmounts calldata unlockAmounts
3. uint40 cliffTime

src/core/SablierLockup.sol
```diff
@@ -411,7 +429,8 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
     /// @dev See the documentation for the user-facing functions that call this internal function.
     function _createLL(
         Lockup.CreateWithTimestamps memory params,
-        uint40 cliffTime
+        uint40 cliffTime,
+        LockupLinear.UnlockAmounts calldata unlockAmounts
```
smol-ninja, line 433: Same here: change the order of input parameters

src/core/libraries/Helpers.sol
```diff
@@ -201,11 +210,20 @@ library Helpers {
                 revert Errors.SablierLockup_CliffTimeNotLessThanEndTime(cliffTime, timestamps.end);
             }
         }
+        // Check: the cliff unlock amount is zero when the cliff time is zero.
+        else if (unlockAmounts.cliff > 0) {
+            revert Errors.SablierLockup_CliffUnlockAmountNotZero(unlockAmounts.cliff);
```
smol-ninja, line 215: The error name is unclear. How about `SablierLockup_CliffTimeZeroUnlockAmountNotZero`?

src/core/interfaces/ISablierLockup.sol
```diff
@@ -110,13 +113,24 @@ interface ISablierLockup is ISablierLockupBase {
...
     /// @dev Reverts if `streamId` references a null stream or a non Lockup Tranched stream.
     /// @param streamId The stream ID for the query.
+    /// @return tranches See the documentation in {DataTypes}.
     function getTranches(uint256 streamId) external view returns (LockupTranched.Tranche[] memory tranches);
 
+    /// @notice Retrieves the unlock amounts used to compose the linear distribution function.
+    /// @dev Reverts if `streamId` references a null stream.
+    /// @param streamId The stream ID for the query.
+    /// @return unlockAmounts See the documentation in {DataTypes}.
+    function getUnlockAmounts(uint256 streamId)
```
smol-ninja, line 129: Should revert for non-linear streams.

src/core/interfaces/ISablierLockup.sol
```diff
@@ -52,6 +52,8 @@ interface ISablierLockup is ISablierLockupBase {
     /// @param transferable Boolean indicating whether the stream NFT is transferable or not.
     /// @param timestamps Struct encapsulating (i) the stream's start time and (ii) end time, all as Unix timestamps.
     /// @param cliffTime The Unix timestamp for the cliff period's end. A value of zero means there is no cliff.
+    /// @param unlockAmounts Struct encapsulating (i) the amount unlocked at the start time and (ii) the amount unlocked
```
smol-ninja, line 55: Should not be past tense here because `unlockAmounts` refer to the future.

```suggestion
    /// @param unlockAmounts Struct encapsulating (i) the amount to unlock at the start time and (ii) the amount to unlock
```

src/core/interfaces/ISablierLockup.sol
```diff
@@ -110,13 +113,24 @@ interface ISablierLockup is ISablierLockupBase {
     /// @notice Retrieves the segments used to compose the dynamic distribution function.
     /// @dev Reverts if `streamId` references a null stream or a non Lockup Dynamic stream.
     /// @param streamId The stream ID for the query.
+    /// @return segments See the documentation in {DataTypes}.
```
smol-ninja, line 116: Any reason why you decided to add `@return` tag here but not in above functions such as `getCliffTime`?

src/core/interfaces/ISablierLockup.sol
```diff
@@ -152,10 +166,13 @@ interface ISablierLockup is ISablierLockupBase {
     ///
     /// @param params Struct encapsulating the function parameters, which are documented in {DataTypes}.
     /// @param durations Struct encapsulating (i) cliff period duration and (ii) total stream duration, both in seconds.
+    /// @param unlockAmounts Struct encapsulating (i) the amount unlocked at the start time and (ii) the amount unlocked
```
smol-ninja, line 169:
```suggestion
    /// @param unlockAmounts Struct encapsulating (i) the amount to unlock at the start time and (ii) the amount to unlock
```

src/core/interfaces/ISablierLockup.sol
```diff
@@ -230,14 +247,20 @@ interface ISablierLockup is ISablierLockupBase {
...
     /// - `params.recipient` must not be the zero address.
     /// - `params.sender` must not be the zero address.
+    /// - The sum of `params.unlockAmounts.start` and `params.unlockAmounts.cliff` must be less than or equal to
+    /// deposit amount.
+    /// - If `params.timestamps.cliff` not set, the `params.unlockAmounts.cliff` must be zero.
     /// - `msg.sender` must have allowed this contract to spend at least `params.totalAmount` assets.
     ///
     /// @param params Struct encapsulating the function parameters, which are documented in {DataTypes}.
     /// @param cliffTime The Unix timestamp for the cliff period's end. A value of zero means there is no cliff.
+    /// @param unlockAmounts Struct encapsulating (i) the amount unlocked at the start time and (ii) the amount unlocked
```
smol-ninja, line 257: Same here.

src/core/interfaces/ISablierLockup.sol
```diff
@@ -230,14 +247,20 @@ interface ISablierLockup is ISablierLockupBase {
     /// `params.timestamps.end`.
     /// - `params.recipient` must not be the zero address.
     /// - `params.sender` must not be the zero address.
+    /// - The sum of `params.unlockAmounts.start` and `params.unlockAmounts.cliff` must be less than or equal to
+    /// deposit amount.
```
smol-ninja, line 251: Since deposit amount does not exist in the context of create function.

```suggestion
    /// `params.totalAmount`.
```

src/core/libraries/VestingMath.sol
```diff
@@ -99,19 +99,21 @@ library VestingMath {
...
-    /// f(x) = x * d + c
+    /// f(x) = x * sa + s + c
     /// $$
     ///
     /// Where:
     ///
-    /// - $x$ is the elapsed time divided by the stream's total duration.
-    /// - $d$ is the deposited amount.
-    /// - $c$ is the cliff amount.
+    /// - $x$ is the elapsed time in the streamable range divided by the streamable total duration.
```
smol-ninja, line 107: Since you have already defined the total duration as the streamable range.

```suggestion
    /// - $x$ is the elapsed time in the streamable range divided by the streamable range.
```

src/core/libraries/VestingMath.sol
```diff
@@ -99,19 +99,21 @@ library VestingMath {
     /// @dev Lockup linear model uses the following distribution function:
     ///
     /// $$
-    /// f(x) = x * d + c
+    /// f(x) = x * sa + s + c
```
smol-ninja, line 102: I'd prefer `s + c + x * sa`, the order in which they are unlocked.

src/core/libraries/VestingMath.sol
```diff
@@ -99,19 +99,21 @@ library VestingMath {
...
     /// $$
     ///
     /// Where:
     ///
-    /// - $x$ is the elapsed time divided by the stream's total duration.
-    /// - $d$ is the deposited amount.
-    /// - $c$ is the cliff amount.
+    /// - $x$ is the elapsed time in the streamable range divided by the streamable total duration.
+    /// - $sa$ is the streamable amount, i.e. deposited amount minus unlock amounts' sum.
```
smol-ninja, line 108: Since its a library, we can be explicit to avoid any confusion
```suggestion
    /// - $sa$ is the streamable amount, i.e. deposited amount - start unlocked amount - cliff unlocked amount.
```

src/core/libraries/VestingMath.sol
```diff
@@ -120,36 +122,48 @@ library VestingMath {
...
 
-            // Divide the elapsed time by the stream's total duration.
-            UD60x18 elapsedTimePercentage = elapsedTime.div(totalDuration);
+            UD60x18 elapsedTime;
+            UD60x18 streamableDuration;
+
+            // Calculate the streamable range.
+            if (cliffTime == 0) {
+                elapsedTime = ud(blockTimestamp - startTime);
+                streamableDuration = ud(endTime - startTime);
```
smol-ninja, line 146: How about calling it `streamableRange` since thats what you have been using throughout the natspecs?

src/core/libraries/VestingMath.sol
```diff
@@ -120,36 +122,48 @@ library VestingMath {
...
         unchecked {
-            // Calculate how much time has passed since the stream started, and the stream's total duration.
-            UD60x18 elapsedTime = ud(blockTimestamp - startTime);
-            UD60x18 totalDuration = ud(endTime - startTime);
+            uint128 unlockAmountsSum = unlockAmounts.start + unlockAmounts.cliff;
+
+            //  If the sum of the unlock amounts is greater than or equal to the deposited amount, return the deposited
+            // amount. The ">=" operator is used as a safety measure in case of a bug, as the sum of the unlock amounts
+            // should never exceed the deposited amount.
+            if (unlockAmountsSum >= depositedAmount) {
```
smol-ninja, line 136: Why is this necessary since we already have a check down here?

```solidity
if (streamedAmount > depositedAmount) {
    return withdrawnAmount;
}
```
If `unlockAmountsSum > depositedAmount` implies `streamedAmount > depositedAmount` so in case of a bug (i.e. ), user wont be able to withdraw any amount than already withdrawn any longer.

Therefore, this is not necessary.

src/periphery/SablierMerkleLL.sol
```diff
@@ -36,6 +36,9 @@ contract SablierMerkleLL is
     /// @inheritdoc ISablierMerkleLL
     MerkleLL.Schedule public override schedule;
 
+    /// @inheritdoc ISablierMerkleLL
+    LockupLinear.UnlockAmounts public override unlockAmounts;
```
smol-ninja, line 40: Instead of creating is as a separate variable, how about merging it into `schedule`. The name `schedule` is very generic and can include amounts as well.

smol-ninja (changes requested):
Nice work on the PR. Reviewed the `src` and left my comments below.

src/core/SablierLockup.sol
andreivladbrg: yes, waiting for the last rebase

smol-ninja: I have already rebased from #1069 this morning.

src/core/interfaces/ISablierLockup.sol
andreivladbrg: correct

src/core/interfaces/ISablierLockup.sol
andreivladbrg: to point out the reader to `DataTypes` file

src/core/libraries/VestingMath.sol
andreivladbrg: since it is a mathematical notation (function of time - x), i believe keeping like this  is the correct choice

in mathematics, expressions are typically written with variable-dependent and higher-degree terms first, followed by constants, to highlight the function’s main behavior and make its structure immediately clear

src/core/interfaces/ISablierLockup.sol
smol-ninja: You are right. I missed that.

src/core/libraries/VestingMath.sol
smol-ninja: Oh you are right. How could I forget: `y = mx + c`.

src/core/libraries/VestingMath.sol
smol-ninja: Fair enough.

src/core/libraries/VestingMath.sol
andreivladbrg: `unlockAmountsSum` and `streamedAmount`  are not the same.

my idea with checking the `unlockAmountsSum` to deposit amount, is to return as soon as possible in case they equal to not perform latter calculations (when `block.timestamp` is between the streamable range)

src/periphery/SablierMerkleLL.sol
andreivladbrg: hmm, not sure what to say, will think more

src/core/libraries/VestingMath.sol
smol-ninja: Good thought. Agree with your explanation.

src/core/SablierLockup.sol
andreivladbrg: my rational was that the cliff time would be more common than the unlock amounts, but i don't have a strong opinion

what is your reasoning for this?

smol-ninja: > my rational was that the cliff time would be more common than the unlock amounts

Mb but guessing at this stage is hard, unlock at start is also a very popular distribution strategy. It can also be popular for instant airdrops (no more 1 sec vesting hack). I tend to follow the below rule:

1. Have structs first followed by simple value types.
2. Order structs alphabetically using variable names.
3. Order value types alphabetically using variable names.

If you have other ideas, I'd love to hear that. My reasoning is weak as well but inspired by the fact that its sometimes arbitrary how we order input parameters in the function names. So having a rule would help. We can also place it in [Sablier Coding Practices](https://sablier.notion.site/Sablier-Coding-Practices-Contracts-2f5920cf497646b4a2ea982c1bacb5a8).

andreivladbrg: ended up making this change

src/periphery/SablierMerkleLL.sol
andreivladbrg: i would like to keep the current version to have two different structs 1. for time related variables 2. amount related amounts

andreivladbrg:
@smol-ninja I have addressed all the `src` feedback, and I have finally rebased from test PRs—it was a mess… but we’re here. I hope I didn’t miss anything while rebasing.

src/core/SablierLockup.sol
```diff
@@ -408,36 +433,47 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
         });
     }
 
+    struct CreateLLParams {
```
andreivladbrg, line 436: temporarily added this due to stack too deep error

will try to think of a better option, but couldn't find one yet (tried scoping `{}` blocks of code, tried a different function for the event, tried splitting the logic, and didn't work unfortunetely)

src/periphery/SablierMerkleLL.sol
smol-ninja: 1. https://chatgpt.com/share/67361585-525c-8000-a6fc-dfe1c0e46207
2. https://ltse.com/insights/what-is-a-standard-vesting-schedule

As per above two links, schedule can have both vesting amounts and durations.

src/core/SablierLockup.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1075/commits/5764dd34aa56f544ff1e5caaa0be00bf9f9a610e

test/core/integration/concrete/lockup-linear/create-with-timestamps-ll/createWithTimestampsLL.tree
```diff
@@ -1,6 +1,8 @@
 CreateWithTimestampsLL_Integration_Concrete_Test
 └── when asset contract
    ├── when cliff time zero
+   │  ├── when cliff unlock amount not zero
+   │  │  └── it should revert
```
smol-ninja, line 5: The next branch should be `when cliff unlock amount zero`

test/core/integration/concrete/lockup-linear/get-unlock-amounts/getUnlockAmounts.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        uint256 streamId = createDefaultStream();
+        LockupLinear.UnlockAmounts memory unlockAmounts = lockup.getUnlockAmounts(streamId);
+        assertEq(unlockAmounts.start, 0, "unlockAmounts.start");
+        assertEq(unlockAmounts.cliff, 0, "unlockAmounts.cliff");
+    }
+
+    function test_GivenStartUnlockAmountZero() external view givenNotNull givenLinearModel givenOnlyOneAmountZero {
+        LockupLinear.UnlockAmounts memory unlockAmounts = lockup.getUnlockAmounts(defaultStreamId);
+        assertEq(unlockAmounts.start, 0, "unlockAmounts.start");
+        assertEq(unlockAmounts.cliff, _defaultParams.unlockAmounts.cliff, "unlockAmounts.cliff");
```
smol-ninja, line 36: defaultStreamId is created using `_defaultParams`. So comparing these two would be wrong. The other one should be `defaults.CLIFF_AMOUNT`

test/core/integration/concrete/lockup-linear/get-unlock-amounts/getUnlockAmounts.t.sol
```diff
@@ -0,0 +1,55 @@
...
+        assertEq(unlockAmounts.start, 1, "unlockAmounts.start");
+        assertEq(unlockAmounts.cliff, 0, "unlockAmounts.cliff");
+    }
+
+    function test_GivenBothAmountsNotZero() external givenNotNull givenLinearModel {
+        _defaultParams.unlockAmounts.start = 1;
+        uint256 streamId = createDefaultStream();
+        LockupLinear.UnlockAmounts memory unlockAmounts = lockup.getUnlockAmounts(streamId);
+        assertEq(unlockAmounts.start, 1, "unlockAmounts.start");
+        assertEq(unlockAmounts.cliff, _defaultParams.unlockAmounts.cliff, "unlockAmounts.cliff");
```
smol-ninja, line 53: Same here.

test/core/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -1,16 +1,17 @@
 StreamedAmountOf_Lockup_Linear_Integration_Concrete_Test
-├── given PENDING status
```
smol-ninja, line 2: Rationale behind removing these tests?

test/core/integration/concrete/lockup-tranched/create-with-durations-lt/createWithDurationsLT.t.sol
```diff
@@ -51,14 +51,14 @@ contract CreateWithDurationsLT_Integration_Concrete_Test is Lockup_Tranched_Inte
...
-        uint256 index = 2;
+        uint256 index = 1;
+        tranches[index].duration = 0;
         vm.expectRevert(
             abi.encodeWithSelector(
                 Errors.SablierHelpers_TrancheTimestampsNotOrdered.selector,
                 index,
-                startTime + tranches[0].duration + tranches[1].duration,
-                startTime + tranches[0].duration + tranches[1].duration
+                startTime + tranches[0].duration,
```
smol-ninja, line 60: Rationale behind changing tranche size from 3 to 2?

test/core/integration/concrete/lockup-tranched/create-with-timestamps-lt/createWithTimestampsLT.t.sol
```diff
@@ -140,6 +140,7 @@ contract CreateWithTimestampsLT_Integration_Concrete_Test is CreateWithTimestamp
         whenStartTimeLessThanFirstTimestamp
     {
         // Swap the tranche timestamps.
+        // LockupTranched.Tranche[] memory tranches = defaults.tranches();
```
smol-ninja, line 143: This should be removed.

test/core/integration/concrete/lockup-tranched/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -3,10 +3,7 @@ StreamedAmountOf_Lockup_Tranched_Integration_Concrete_Test
    ├── given start time in present
    │  └── it should return zero
    └── given start time in past
-      ├── given end time not in future
```
smol-ninja, line 6: Rationale behind removing them? Also, modifiers still exist in the corresponding test contracts.

src/periphery/SablierMerkleLL.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1088

smol-ninja (changes requested):
Nice work on tests. Left my review below. 

Also created a PR: https://github.com/sablier-labs/v2-core/pull/1088 about `Schedule` struct that can be merged into this.

test/utils/Defaults.sol
```diff
@@ -161,34 +167,40 @@ contract Defaults is Constants, Merkle {
...
-        tranchesWithDurations_[0] = LockupTranched.TrancheWithDuration({ amount: 2500e18, duration: 2500 seconds });
-        tranchesWithDurations_[1] = LockupTranched.TrancheWithDuration({ amount: 100e18, duration: 100 seconds });
-        tranchesWithDurations_[2] = LockupTranched.TrancheWithDuration({ amount: 7400e18, duration: 7400 seconds });
+        tranchesWithDurations_ = new LockupTranched.TrancheWithDuration[](2);
+        tranchesWithDurations_[0] = LockupTranched.TrancheWithDuration({ amount: 2600e18, duration: 2600 seconds });
+        tranchesWithDurations_[1] = LockupTranched.TrancheWithDuration({ amount: 7400e18, duration: 7400 seconds });
+    }
+
+    function unlockAmounts() public pure returns (LockupLinear.UnlockAmounts memory) {
+        return LockupLinear.UnlockAmounts({ start: 0, cliff: CLIFF_AMOUNT });
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1088

package.json
```diff
@@ -12,7 +12,7 @@
   },
   "dependencies": {
     "@openzeppelin/contracts": "5.0.2",
-    "@prb/math": "4.0.3"
+    "@prb/math": "4.1.0"
```
smol-ninja, line 15: Should be changed in peerDependencies as well.

test/core/fork/LockupLinear.t.sol
```diff
@@ -205,10 +231,12 @@ abstract contract Lockup_Linear_Fork_Test is Fork_Test {
         assertEq(lockup.getSender(vars.streamId), params.sender, "sender");
         assertEq(lockup.getStartTime(vars.streamId), params.timestamps.start, "startTime");
         assertFalse(lockup.wasCanceled(vars.streamId), "wasCanceled");
+        assertEq(lockup.getUnlockAmounts(vars.streamId).start, params.unlockAmounts.start, "unlockAmounts.start");
+        assertEq(lockup.getUnlockAmounts(vars.streamId).cliff, params.unlockAmounts.cliff, "unlockAmounts.cliff");
 
         // Assert that the stream's status is correct.
         vars.actualStatus = lockup.statusOf(vars.streamId);
-        if (params.timestamps.end <= vars.blockTimestamp) {
+        if (vars.streamedAmount == vars.createAmounts.deposit) {
```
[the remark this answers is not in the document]
andreivladbrg: because the status is calculated based on an amount and not on time (expect if it is `PENDING`):

https://github.com/sablier-labs/v2-core/blob/6a8ec1318f598c351606eb196b9ccebed7dedbab/src/core/abstracts/SablierLockupBase.sol#L513-L516

similar to dynamic, where you can have 3 segments stream with: `s[0].amount + s[1].amount == depositedAmount && s[2].amount == 0`, here is the same, you can have the `unlockAmounts.start + unlockAmounts.cliff == deposited`, so it does not matter the `block.timestamp`

test/utils/Defaults.sol
```diff
@@ -27,22 +27,24 @@ contract Defaults is Constants, Merkle {
...
-    uint128 public constant CLIFF_AMOUNT = 2500e18;
+    uint128 public constant CLIFF_AMOUNT = 2500e18 + 2534;
     uint40 public immutable CLIFF_TIME;
     uint40 public constant CLIFF_DURATION = 2500 seconds;
     uint128 public constant DEPOSIT_AMOUNT = 10_000e18;
     uint40 public immutable END_TIME;
     uint256 public constant MAX_COUNT = 10_000;
     uint40 public immutable MAX_SEGMENT_DURATION;
-    uint128 public constant REFUND_AMOUNT = DEPOSIT_AMOUNT - CLIFF_AMOUNT;
+    uint256 public constant MAX_TRANCHE_COUNT = 10_000;
```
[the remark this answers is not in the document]
andreivladbrg: no need for this, accidentally remained when rebasing (solving conflicts is super prone to errors 😁)

test/utils/Defaults.sol
```diff
@@ -27,22 +27,24 @@ contract Defaults is Constants, Merkle {
     uint64 public constant BATCH_SIZE = 10;
     UD60x18 public constant BROKER_FEE = UD60x18.wrap(0.003e18); // 0.3%
     uint128 public constant BROKER_FEE_AMOUNT = 30.090270812437311935e18; // 0.3% of total amount
-    uint128 public constant CLIFF_AMOUNT = 2500e18;
+    uint128 public constant CLIFF_AMOUNT = 2500e18 + 2534;
```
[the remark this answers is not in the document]
andreivladbrg: any suggestions, i can't think of something short

the idea is to have the streamed amount at `WARP_26_PERCENT` equal to `WITHDRAW_AMOUNT`

test/core/fork/LockupLinear.t.sol
```diff
@@ -184,14 +195,29 @@ abstract contract Lockup_Linear_Fork_Test is Fork_Test {
...
+            params.timestamps.start,
+            params.cliffTime,
+            params.timestamps.end,
+            vars.createAmounts.deposit,
+            params.unlockAmounts
+        );
+
         // Check if the stream is settled. It is possible for a Lockup Linear stream to settle at the time of creation
-        // in case end time is in the past.
-        vars.isSettled = params.timestamps.end <= vars.blockTimestamp;
```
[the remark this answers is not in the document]
andreivladbrg: explained here https://github.com/sablier-labs/v2-core/pull/1075#discussion_r1843451814

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol
```diff
@@ -184,9 +184,10 @@ contract WithdrawMultiple_Integration_Concrete_Test is Integration_Test {
         givenNoNullStreams
         givenNoDEPLETEDStreams
         whenNoZeroAmounts
+        whenCallerAuthorizedForAllStreams
```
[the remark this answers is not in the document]
andreivladbrg: it does not matter, in your singletone test PR, you removed it while this has dedicated logic to run the tests with the `msg.sender` as these:

https://github.com/sablier-labs/v2-core/blob/5764dd34aa56f544ff1e5caaa0be00bf9f9a610e/test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol#L162-L165

otherwise we can remove it as it is not used

BTT is incomplete, the main purpose of it, is to simply divide in binary branches the logic.

package.json
andreivladbrg: correct, missed it
it should be: "4.x.x"

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol
```diff
@@ -40,7 +40,7 @@ contract WithdrawMultiple_Integration_Concrete_Test is Integration_Test {
         // 2. A stream with an early end time
         // 3. A stream meant to be canceled before the withdrawal is made
         streamIds[0] = createDefaultStream();
-        streamIds[1] = createDefaultStreamWithEndTime(defaults.WARP_26_PERCENT());
+        streamIds[1] = createDefaultStreamWithEndTime(defaults.WARP_26_PERCENT() + 1);
```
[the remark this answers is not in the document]
andreivladbrg: have changed the first `segment.timestamp` to :

https://github.com/sablier-labs/v2-core/blob/5764dd34aa56f544ff1e5caaa0be00bf9f9a610e/test/utils/Defaults.sol#L151

so we need to have a value for the end time greater than 26%, otherwise we would get the timestamps not ordered error

test/core/integration/concrete/lockup-linear/create-with-timestamps-ll/createWithTimestampsLL.tree
andreivladbrg: correct, missed it

test/core/integration/concrete/lockup-linear/get-unlock-amounts/getUnlockAmounts.t.sol
andreivladbrg: sorry, i don't get it

`_defaultParams.unlockAmounts.cliff` and  `defaults.CLIFF_AMOUNT` are the same thing

test/core/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree
andreivladbrg: duplicated tests (in the common `lockup-base` file):

https://github.com/sablier-labs/v2-core/blob/5764dd34aa56f544ff1e5caaa0be00bf9f9a610e/test/core/integration/concrete/lockup-base/streamed-amount-of/streamedAmountOf.tree#L10-L17

test/core/integration/concrete/lockup-linear/get-unlock-amounts/getUnlockAmounts.t.sol
andreivladbrg: answered above

test/core/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -1,12 +1,6 @@
 StreamedAmountOf_Lockup_Dynamic_Integration_Concrete_Test
 └── given STREAMING status
-   ├── given start time in present
```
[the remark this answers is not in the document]
andreivladbrg: forgot about modifiers

also, since we are declaring the tree branches in terms of the status, wdyt about doing this:

```solidity
modifier runBothCases() {
    vm.warp({ newTimestamp: defaults.START_TIME() });
    _;

    vm.warp({ newTimestamp: defaults.WARP_26_PERCENT() });
    _;
}

function test_GivenSTREAMINGStatus() external givenNotNull givenNotCanceledStream runBothCases {
    // It should return the correct streamed amount.
    uint128 actualStreamedAmount = lockup.streamedAmountOf(defaultStreamId);
    uint128 expectedStreamedAmount = getBlockTimestamp() == defaults.START_TIME() ? 0 : defaults.WITHDRAW_AMOUNT();
    assertEq(actualStreamedAmount, expectedStreamedAmount, "streamedAmount");
}
```

`STREAMING` status means: 
1. `block.timestamp == startTime`
2. `block.timestamp > startTime`

test/core/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -14,19 +14,6 @@ contract StreamedAmountOf_Lockup_Dynamic_Integration_Concrete_Test is
         Lockup_Dynamic_Integration_Concrete_Test.setUp();
     }
 
-    function test_GivenStartTimeInPresent() external givenSTREAMINGStatus {
```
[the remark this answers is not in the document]
andreivladbrg: the reason is that this test should be covered by the `STREAMING` status. as explained above. lmk what you think about my proposal

test/core/integration/concrete/lockup-tranched/create-with-durations-lt/createWithDurationsLT.t.sol
andreivladbrg: Previously, we declared the tranches with a size of 3 due to not having the "streaming" component in this distribution model, so that when warping to `WARP_26_PERCENT`, the withdrawable amount would be `WITHDRAW_AMOUNT`. Now, I have also updated the segments to have the first timestamp as `START_TIME + WARP_26_PERCENT` and can do the same for the tranches

previously segments:

https://github.com/sablier-labs/v2-core/blob/6a8ec1318f598c351606eb196b9ccebed7dedbab/test/utils/Defaults.sol#L143-L151

previously tranches:

https://github.com/sablier-labs/v2-core/blob/6a8ec1318f598c351606eb196b9ccebed7dedbab/test/utils/Defaults.sol#L176-L181

now they are in sync: 

https://github.com/sablier-labs/v2-core/blob/5764dd34aa56f544ff1e5caaa0be00bf9f9a610e/test/utils/Defaults.sol#L145-L157

https://github.com/sablier-labs/v2-core/blob/5764dd34aa56f544ff1e5caaa0be00bf9f9a610e/test/utils/Defaults.sol#L182-L186

test/core/integration/concrete/lockup-tranched/streamed-amount-of/streamedAmountOf.tree
andreivladbrg: same reason as here

https://github.com/sablier-labs/v2-core/pull/1075#discussion_r1843568546

already tested in `SETTLED` status branch

also re this https://github.com/sablier-labs/v2-core/pull/1075#discussion_r1843598254

we should consider for `SETTLED` this modifier for both `now == end && now > end`:

```solidity
modifier runBothCases() {
    vm.warp({ newTimestamp: defaults.END_TIME() });
    _;

    vm.warp({ newTimestamp: defaults.END_TIME() + 1 });
    _;
}
```

test/utils/Defaults.sol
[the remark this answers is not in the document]
smol-ninja: > solving conflicts is super prone to errors

How do you resolve conflicts? I don't use VSCode Conflict resolver. If you use that, then yes its a bad experience and can lead to such errors. But if you just open a file and use those hyperlinks to resolve conflicts, its pretty good experience and I have never felt bad resolving conflicts with them.

test/core/fork/LockupLinear.t.sol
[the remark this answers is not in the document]
smol-ninja: Fair point.

test/core/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.tree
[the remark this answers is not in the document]
smol-ninja: Technically works but may I know your concern with the previous approach? Do you want to DRY'ify it using a single modifier? 

I, personally, preferred previous approach where we had clear tree structure and corresponding tests. Our tests are large. To me, it was a no-brainer to review whether the tests are correct or not. Now, it requires some efforts to verify the same since they dont correctly comply with each other. Having said that, I'd personally prefer repeat code in such tests over DRY'ify. And this is why I strongly insist to stick to BTT unless we have find a better framework to follow.

PS: I may be wrong in understanding your reply here. Feel free to correct me if thats the case.

test/core/integration/concrete/lockup-linear/get-unlock-amounts/getUnlockAmounts.t.sol
smol-ninja: Yes they are. But `_defaultParams` can be overridden where defaults are fixed. So far, we have also been comparing values from function calls with either defaults or constants.

test/core/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.tree
smol-ninja: [This is](https://github.com/sablier-labs/v2-core/pull/1075#discussion_r1843999835) my reply to this as well.

package.json
smol-ninja: Thanks. Should also be updated in README where we have installation guidelines.

andreivladbrg: yes, we should

test/utils/Defaults.sol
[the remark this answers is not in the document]
andreivladbrg: > VSCode Conflict resolver

yes, i use it

> then yes its a bad experience and can lead to such errors. But if you just open a file and use those hyperlinks to resolve conflicts

hmm, ok, will give it a try next time, thx

test/utils/Defaults.sol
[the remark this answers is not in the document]
andreivladbrg: > `STREAMED_AMOUNT_26` 

sounds good👍, similar to `ONE_MONTH_STREAMED` from flow

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol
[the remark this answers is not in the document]
andreivladbrg: > I would recommend you to start a discussion on ideas to improve it.

that's a good idea, but it takes time, and we need to take a decision to not stop our dev proces

---

also, why does it make not Bulloak compatible if we include other modifiers (that have unique logic) to the function tests? 
```solidity
// BTT modifier
modifier whenX {
   _; 
}
// non BTT modifier 
modifier whenAllCallersPossible {
    // run test with caller 1
    _;
    // run test in caller 2
    _;
    // run test in caller 3
    _;
}

function testFoo_SomeY() public whenX  whenAllCallersPossible {
    assertFalse(Some(condition));
    assertTrue(Some(otherCondition));
}
```

test/core/integration/concrete/lockup-dynamic/streamed-amount-of/streamedAmountOf.tree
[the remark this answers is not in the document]
andreivladbrg: > but may I know your concern with the previous approach? 

there were duplicated tests which were testing the same thing, but the test branches were formulated differently

e.g.

in streamed amount test from lockup-base dir:

https://github.com/sablier-labs/v2-core/blob/6a8ec1318f598c351606eb196b9ccebed7dedbab/test/core/integration/concrete/lockup-base/streamed-amount-of/streamedAmountOf.t.sol#L36-L43

in  streamed amount test from lockup-linear dir:

https://github.com/sablier-labs/v2-core/blob/6a8ec1318f598c351606eb196b9ccebed7dedbab/test/core/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol#L27-L33

---

> it requires some efforts to verify the same since they dont correctly comply with each other

i am sorry, i didn't get this

> 'd personally prefer repeat code in such tests over DRY'ify. And this is why I strongly insist to stick to BTT unless we have find a better framework to follow.

once again, same question from above, why adding other modifiers that have unique logic, non-BTT compatible?

test/core/integration/concrete/lockup-base/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -43,6 +43,15 @@ abstract contract StreamedAmountOf_Integration_Concrete_Test is Integration_Test
     }
 
     function test_GivenSTREAMINGStatus() external givenNotNull givenNotCanceledStream {
+        vm.warp({ newTimestamp: defaults.WARP_26_PERCENT() });
+
+        // It should return the correct streamed amount.
+        uint128 actualStreamedAmount = lockup.streamedAmountOf(defaultStreamId);
+        uint128 expectedStreamedAmount = defaults.WITHDRAW_AMOUNT();
```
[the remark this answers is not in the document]
andreivladbrg: does this commit addresses this? 

7e1bad8c9da39e922b312332a486b735ba38eb49

test/utils/Defaults.sol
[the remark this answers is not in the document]
andreivladbrg: 7e1bad8c9da39e922b312332a486b735ba38eb49

test/core/integration/concrete/lockup-linear/get-unlock-amounts/getUnlockAmounts.t.sol
andreivladbrg: dfe17c3da8d38100210c70aa5bf35d696af18e40

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol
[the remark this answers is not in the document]
smol-ninja: Alongside `whenAllCallersPossible ` there would also be a case when all callers are not possible. A realistic scenario would be

```tree
Doge_Test
├── when not x
│   └── it should revert
└── when x
    ├── when one caller is not authorized
    │   └── it should revert
    └── when all callers are authorized
        └── when some Y
            └── it do somethin
```

for which bulloak would generate 

```solidity
contract Doge_Test {
    function test_RevertWhen_NotX() external {
        // it should revert
    }

    modifier whenX() {
        _;
    }

    function test_RevertWhen_OneCallerIsNotAuthorized() external whenX {
        // it should revert
    }

    modifier whenAllCallersAreAuthorized() {
        _;
    }

    function test_WhenSomeY() external whenX whenAllCallersAreAuthorized {
        // it do somethin
    }
}
```

in which you can override `whenAllCallersPossible`

Does that make sense?

test/core/integration/concrete/lockup-base/streamed-amount-of/streamedAmountOf.t.sol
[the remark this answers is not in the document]
smol-ninja: Yes. Thank you.

test/utils/Defaults.sol
```diff
@@ -27,22 +27,26 @@ contract Defaults is Constants, Merkle {
...
+    uint128 public constant START_AMOUNT = 0;
+    uint128 public constant STREAMED_AMOUNT_26_PERCENT = 2600e18;
     uint128 public constant TOTAL_AMOUNT = 10_030.090270812437311935e18; // deposit + broker fee
     uint40 public constant TOTAL_DURATION = 10_000 seconds;
-    uint256 public constant TRANCHE_COUNT = 3;
+    uint256 public constant TRANCHE_COUNT = 2;
     uint128 public constant TOTAL_TRANSFER_AMOUNT = DEPOSIT_AMOUNT * uint128(BATCH_SIZE);
-    uint128 public constant WITHDRAW_AMOUNT = 2600e18;
-    uint40 public immutable WARP_26_PERCENT; // 26% of the way through the stream
+    uint128 public constant WITHDRAW_AMOUNT = STREAMED_AMOUNT_26_PERCENT;
```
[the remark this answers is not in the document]
andreivladbrg: nice to have for `withdraw` tests

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol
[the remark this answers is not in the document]
andreivladbrg: > in which you can override whenAllCallersPossible
Does that make sense?

Yes, it makes sense, but I believe you are missing the fact that the caller tree branches are covered in the `withdraw.tree` tests (which we don’t need to specifically re-test here). Here, we are talking about `withdrawMultiple`. By addind the `whenAllCallersPossible` we just adding "nice have" elegant (without being redundant), that is not a requirement

---

This question was not a question; could you answer it, please? Why does it make it not compatible with Bulloak if we include other modifiers (that have unique logic) in the function tests?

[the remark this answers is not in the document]
smol-ninja: > also, why does it make not Bulloak compatible if we include other modifiers (that have unique logic) to the function tests?

Even though bulloak check would pass since it ignored additional modifiers, the goal with the bulloak and BTT is to write tree first and then generate the test contract automatically. The automation also adds required modifiers. If we add a modifier to the test that is not generated automatically by the bulloak, I would consider the test to be incompatible with the tree file.

Imagine a modifier is added but not a part of the tree and that modifier has some logic inside, this can introduce errors to the tests.

smol-ninja (approved):
Approving the PR. There are still some unresolved comments regarding tests but since they are related to tests we can look into them later. The contract code looks good to me so given the audit date is near, feel free to merge it now.

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.t.sol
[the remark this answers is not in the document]
andreivladbrg: > the goal with the bulloak and BTT is to write tree first and then generate the test contract automatically.

yes, but we are _not_ achieving this since we have agreed to skip the modifiers since we had a lot of duplicated modifiers in each file

> The automation also adds required modifiers

yes, which we ignore since we have the `Modifiers` file

> I would consider the test to be incompatible with the tree file

by what you said above, even the current version is incompatible


> Imagine a modifier is added but not a part of the tree and that modifier has some logic inside, this can introduce errors to the tests.

for sure if it adds an error to the test, we would catch this when we run `forge test` 
as said above: "By adding the whenAllCallersPossible we just add a "nice have" in an elegant way (without being redundant), that is not a requirement"

andreivladbrg:
Thanks, let's goo 🚀

PR #1084, opened 2024-11-12
refactor: include batch lockup contract in core

andreivladbrg:
Depends on #1075 Closes https://github.com/sablier-labs/v2-core/issues/1073

PR #1087, opened 2024-11-14
Enable ETH fees

### LINKED ISSUE #1076: ETH fee in withdraw function
## Context

As discussed [here](https://github.com/sablier-labs/company-discussions/discussions/72#discussioncomment-11115807), we want to be able to monetize the Lockup streams by charging an ETH fee, since this is much easier than charging in USDC or the stream's token.

## Spec

- [x] Make all functions payable so that the UI can set a non-zero `msg.value`.
- [x] Allow anyone to call `withdrawFees` function, and always transfer the ETH fees to the contract admin.

Note: compared to the ETH fee in MerkleLockup, the fee in this proposal is *optional*. The goal is to keep the protocol free but monetize at the UI level — the Sablier UI would introduce a small ETH fee in the transactions signed by users.

andreivladbrg:
Depends on #1084 

Closes #1076  and https://github.com/sablier-labs/v2-core/issues/1090

### DIFF
--- a/src/core/interfaces/ISablierLockupBase.sol
+++ b/src/core/interfaces/ISablierLockupBase.sol
@@ -67,11 +67,11 @@ interface ISablierLockupBase is
         uint256 indexed streamId, address indexed to, IERC20 indexed asset, uint128 withdrawnAmount
     );
 
-    /// @notice Emitted when Sablier fees are withdrawn.
+    /// @notice Emitted when fees are withdrawn.
     /// @param admin The address of the current contract admin.
     /// @param feeAmount The amount of ETH transferred to the provided address.
-    /// @param to The address that has received the Sablier fees.
-    event WithdrawSablierFees(address indexed admin, uint256 indexed feeAmount, address indexed to);
+    /// @param to The address receiving the fees.
+    event WithdrawFees(address indexed admin, uint256 indexed feeAmount, address indexed to);
 
     /*//////////////////////////////////////////////////////////////////////////
                                  CONSTANT FUNCTIONS
@@ -322,25 +322,24 @@ interface ISablierLockupBase is
     /// - `to` must not be the zero address.
     /// - `amount` must be greater than zero and must not exceed the withdrawable amount.
     /// - `to` must be the recipient if `msg.sender` is not the stream's recipient or an approved third party.
-    /// - The `msg.value` must not be less than the Sablier fee.
     ///
     /// @param streamId The ID of the stream to withdraw from.
     /// @param to The address receiving the withdrawn assets.
     /// @param amount The amount to withdraw, denoted in units of the asset's decimals.
     function withdraw(uint256 streamId, address to, uint128 amount) external payable;
 
-    /// @notice Withdraws the Sablier fees accrued to the provided address.
+    /// @notice Withdraws the accrued fees to the provided address.
     ///
-    /// @dev Emits a {WithdrawSablierFees} event.
+    /// @dev Emits a {WithdrawFees} event.
     ///
     /// Notes:
-    /// - This function transfers ETH to the provided address. If the receiver is a contract, it must be
-    /// able to receive ETH.
+    /// - This function transfers ETH to the provided address. If the receiver is a contract, it must be able to receive
+    /// ETH.
     ///
     /// Requirements:
     /// - `msg.sender` must be the contract admin.
     ///
-    /// @param to The address to receive the Sablier fees.
+    /// @param to The address to transfer the fees to.
     function withdrawFees(address payable to) external;
 
     /// @notice Withdraws the maximum withdrawable amount from the stream to the provided address `to`.
--- a/src/core/interfaces/ISablierLockupBase.sol
+++ b/src/core/interfaces/ISablierLockupBase.sol
@@ -62,8 +62,16 @@ interface ISablierLockupBase is
     /// @param streamId The ID of the stream.
     /// @param to The address that has received the withdrawn assets.
     /// @param asset The contract address of the ERC-20 asset that has been withdrawn.
-    /// @param amount The amount of assets withdrawn, denoted in units of the asset's decimals.
-    event WithdrawFromLockupStream(uint256 indexed streamId, address indexed to, IERC20 indexed asset, uint128 amount);
+    /// @param withdrawnAmount The amount of assets withdrawn, denoted in units of the asset's decimals.
+    event WithdrawFromLockupStream(
+        uint256 indexed streamId, address indexed to, IERC20 indexed asset, uint128 withdrawnAmount
+    );
+
+    /// @notice Emitted when Sablier fees are withdrawn.
+    /// @param admin The address of the current contract admin.
+    /// @param feeAmount The amount of ETH transferred to the provided address.
+    /// @param to The address that has received the Sablier fees.
+    event WithdrawSablierFees(address indexed admin, uint256 indexed feeAmount, address indexed to);
 
     /*//////////////////////////////////////////////////////////////////////////
                                  CONSTANT FUNCTIONS
@@ -227,7 +235,7 @@ interface ISablierLockupBase is
     /// - `msg.sender` must be either the NFT owner or an approved third party.
     ///
     /// @param streamId The ID of the stream NFT to burn.
-    function burn(uint256 streamId) external;
+    function burn(uint256 streamId) external payable;
 
     /// @notice Cancels the stream and refunds any remaining assets to the sender.
     ///
@@ -244,7 +252,7 @@ interface ISablierLockupBase is
     /// - `msg.sender` must be the stream's sender.
     ///
     /// @param streamId The ID of the stream to cancel.
-    function cancel(uint256 streamId) external;
+    function cancel(uint256 streamId) external payable;
 
     /// @notice Cancels multiple streams and refunds any remaining assets to the sender.
     ///
@@ -257,7 +265,7 @@ interface ISablierLockupBase is
     /// - All requirements from {cancel} must be met for each stream.
     ///
     /// @param streamIds The IDs of the streams to cancel.
-    function cancelMultiple(uint256[] calldata streamIds) external;
+    function cancelMultiple(uint256[] calldata streamIds) external payable;
 
     /// @notice Removes the right of the stream's sender to cancel the stream.
     ///
@@ -273,7 +281,7 @@ interface ISablierLockupBase is
     /// - The stream must be cancelable.
     ///
     /// @param streamId The ID of the stream to renounce.
-    function renounce(uint256 streamId) external;
+    function renounce(uint256 streamId) external payable;
 
     /// @notice Renounces multiple streams.
     ///
@@ -286,7 +294,7 @@ interface ISablierLockupBase is
     /// - All requirements from {renounce} must be met for each stream.
     ///
     /// @param streamIds An array of stream IDs to renounce.
-    function renounceMultiple(uint256[] calldata streamIds) external;
+    function renounceMultiple(uint256[] calldata streamIds) external payable;
 
     /// @notice Sets a new NFT descriptor contract, which produces the URI describing the Sablier stream NFTs.
     ///
@@ -314,11 +322,26 @@ interface ISablierLockupBase is
     /// - `to` must not be the zero address.
     /// - `amount` must be greater than zero and must not exceed the withdrawable amount.
     /// - `to` must be the recipient if `msg.sender` is not the stream's recipient or an approved third party.
+    /// - The `msg.value` must not be less than the Sablier fee.
     ///
     /// @param streamId The ID of the stream to withdraw from.
     /// @param to The address receiving the withdrawn assets.
     /// @param amount The amount to withdraw, denoted in units of the asset's decimals.
-    function withdraw(uint256 streamId, address to, uint128 amount) external;
+    function withdraw(uint256 streamId, address to, uint128 amount) external payable;
+
+    /// @notice Withdraws the Sablier fees accrued to the provided address.
+    ///
+    /// @dev Emits a {WithdrawSablierFees} event.
+    ///
+    /// Notes:
+    /// - This function transfers ETH to the provided address. If the receiver is a contract, it must be
+    /// able to receive ETH.
+    ///
+    /// Requirements:
+    /// - `msg.sender` must be the contract admin.
+    ///
+    /// @param to The address to receive the Sablier fees.
+    function withdrawFees(address payable to) external;
 
     /// @notice Withdraws the maximum withdrawable amount from the stream to the provided address `to`.
     ///
@@ -333,7 +356,7 @@ interface ISablierLockupBase is
     /// @param streamId The ID of the stream to withdraw from.
     /// @param to The address receiving the withdrawn assets.
     /// @return withdrawnAmount The amount withdrawn, denoted in units of the asset's decimals.
-    function withdrawMax(uint256 streamId, address to) external returns (uint128 withdrawnAmount);
+    function withdrawMax(uint256 streamId, address to) external payable returns (uint128 withdrawnAmount);
 
     /// @notice Withdraws the maximum withdrawable amount from the stream to the current recipient, and transfers the
     /// NFT to `newRecipient`.
@@ -357,6 +380,7 @@ interface ISablierLockupBase is
         address newRecipient
     )
         external
+        payable
         returns (uint128 withdrawnAmount);
 
     /// @notice Withdraws assets from streams to the recipient of each stream.
@@ -374,5 +398,5 @@ interface ISablierLockupBase is
     ///
     /// @param streamIds The IDs of the streams to withdraw from.
     /// @param amounts The amounts to withdraw, denoted in units of the asset's decimals.
-    function withdrawMultiple(uint256[] calldata streamIds, uint128[] calldata amounts) external;
+    function withdrawMultiple(uint256[] calldata streamIds, uint128[] calldata amounts) external payable;
 }
--- a/src/core/libraries/Errors.sol
+++ b/src/core/libraries/Errors.sol
@@ -23,20 +23,7 @@ library Errors {
     error DelegateCall();
 
     /*//////////////////////////////////////////////////////////////////////////
-                                SABLIER-BATCH-LOCKUP
-    //////////////////////////////////////////////////////////////////////////*/
-
-    error SablierBatchLockup_BatchSizeZero();
-
-    /*//////////////////////////////////////////////////////////////////////////
-                               LOCKUP-NFT-DESCRIPTOR
-    //////////////////////////////////////////////////////////////////////////*/
-
-    /// @notice Thrown when trying to generate the token URI for an unknown ERC-721 NFT contract.
-    error LockupNFTDescriptor_UnknownNFT(IERC721Metadata nft, string symbol);
-
-    /*//////////////////////////////////////////////////////////////////////////
-                                    HELPERS
+                                      HELPERS
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @notice Thrown when the broker fee exceeds the maximum allowed fee.
@@ -110,6 +97,19 @@ library Errors {
     );
 
     /*//////////////////////////////////////////////////////////////////////////
+                               LOCKUP-NFT-DESCRIPTOR
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Thrown when trying to generate the token URI for an unknown ERC-721 NFT contract.
+    error LockupNFTDescriptor_UnknownNFT(IERC721Metadata nft, string symbol);
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                SABLIER-BATCH-LOCKUP
+    //////////////////////////////////////////////////////////////////////////*/
+
+    error SablierBatchLockup_BatchSizeZero();
+
+    /*//////////////////////////////////////////////////////////////////////////
                                 SABLIER-LOCKUP-BASE
     //////////////////////////////////////////////////////////////////////////*/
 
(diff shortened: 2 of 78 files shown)


andreivladbrg:
> Shouldn't this PR be independent of https://github.com/sablier-labs/v2-core/pull/1084? Both are separate issues. One is related to batch functions which is primarily create related whereas this one is about withdraw function.

not necessarily. it’s not about one issue being directly related to the other, but rather about anticipating potential conflicts to avoid time wasted resolving them. you can see the independent changes related to eth fees, as the base branch is refactor/include-batch-core. after merging that branch, the base branch for this pr will automatically switch to staging.

we should merge PRs in chronological order

> It's a good practice to create PRs independent of each other. 

why? since all will eventually be merged into `staging`, it’s more about focusing on the changes specific to this issue, which is exactly what you see with the current base branch.

> If one takes time or requires more changes after review, the other PR can be reviewed and merged independently.

since it's just the two of us working on these changes, and this isn’t an open-source project with multiple contributors (where your approach would be preferred), i believe this approach is more efficient.
also, based on my experience with the unlock amount feature and the conflicts that arose, i can confidently say that creating PRs dependent on others is often more efficient.

#### scenario 1 (sequentially dependent PRs)

1. A <-- B <-- C <-- D
2.           B <-- B review
3.           C <-- rebase B review
4.           C <-- C review
5.           D <-- D review
          
#### scenario 2 (independent PRs with higher conflicts)

1. A <-- B 
2. A <-- C # changes similar files that B changes (conflicts warning)
3. A <-- D # changes similar files that B and C changes (conflicts warning)
4. B <-- B review
5. C <-- rebase B
6. D <-- rebase B
7. C <-- C review
8. D <-- rebase C
9. D <-- review D


as you can see, scenario 1 is better

smol-ninja:
You made some good points. I gave it more thought, and I think what you said is accurate when the time complexity of reviewing, resolving comments, and merging all PRs is the same. In many cases, the time complexities are different. Making all PRs dependent on each other would mean it disallows us from reviewing, resolving, and merging them asynchronously. Even though some PRs may be reviewed and merged faster, they will continue to be blocked by the base PR, which may take more time than anticipated. Also, in the case of a chain of PRs, when changes are made to the base PR, that means making changes to all subsequent PRs.

Your point about rebasing is valid, but in my experience, we can identify such PRs/issues that are truly dependent on each other. In those cases, I agree with you. But generalizing it can slow us down in many cases. 

As an example, I rebased https://github.com/sablier-labs/v2-core/pull/1091 from staging in less than 30 minutes, and it can now also be reviewed and merged independently, which otherwise could stay open for as long as the other PRs are open. Another example is that https://github.com/sablier-labs/v2-core/pull/1087 can be worked upon, reviewed and merged independently of other PRs opened before it.

From my experience, I prefer creating independent PRs as much as possible, which you can review and merge independent of other PRs. This, imo, speeds up closing the issues faster without getting blocked by the base PR.

andreivladbrg:
@smol-ninja i have marked the PR as ready to review, tho the fork tests are failing, we can ignore them for now, since we will remove the "periphery" contract in the upcoming PR

smol-ninja:
Have you also included #1090 in this PR as discussed [here](https://github.com/sablier-labs/v2-core/issues/1090#issuecomment-2483369990)? If yes, can you please mention it in the description?

andreivladbrg:
> Have you also included #1090 in this PR as discussed [here](https://github.com/sablier-labs/v2-core/issues/1090#issuecomment-2483369990)? If yes, can you please mention it in the description?

not yet, facing some issue with it, we might be able to implement it, due to requiring all functions within `calls` array to be payable as well

andreivladbrg:
@smol-ninja Thank you for your detailed reply. 

> You made some good points. I gave it more thought, and I think what you said is accurate when the time complexity of reviewing, resolving comments, and merging all PRs is the same.

Indeed, complexity plays a significant role in this discussion.

> In many cases, the time complexities are different

You are correct that PR complexities can vary. IMO, the sequential approach works best for handling highly complex PRs, whereas the parallel approach is more suited for small changes.

> Making all PRs dependent on each other would mean it disallows us from reviewing

That's incorrect and I disagree with your assertion regarding the base branch of the PR. As I mentioned in my initial comment, what truly matters is the base branch the PR is raised against. For example, if I start from `refactor/include-batch-core` and create a PR with the base branch set as `refactor/include-batch-core`, you will only review the changes specific to `feat/eth-fee`. This is because the git diff will display only those changes.


> Even though some PRs may be reviewed and merged faster, they will continue to be blocked by the base PR

I don’t quite understand why this point is relevant. Whether a PR is merged into staging after 1 day or after 5 days doesn’t seem to matter as long as all PRs ultimately need to be merged into `staging` for the code to be audited.

Let me give an example: Suppose everything must be merged into staging. I create branches A and B, both involving complex changes and targeting the same files and lines (guaranteeing conflicts). Branch B is created after Branch A, but the PR for Branch B is raised first. The review processes for both start simultaneously. If PR A takes 1 day for review and changes, and PR B takes 2 days, merging PR A immediately upon completion would require PR B to be rebased and conflicts resolved, adding another day (3 days total).

In contrast, if Branch B were created sequentially from Branch A, rebasing the review changes of PR A into Branch B would only take half a day. In this case, having both branches on staging would take 2.5 days instead of 3 days. For more complex PRs, I still believe the sequential method is more efficient.

Another argument is that, for us (contract development), we don’t need specific features merged into the `main` branch as soon as possible, unlike other repositories (like `v2-interfaces`). In that case, I completely agree with you that parallel PRs are the way to go. However, for us, since we release our code after it has been audited, we don’t need to merge into the `main` or `staging` branch immediately. I hope I have made myself clear

> which may take more time than anticipated. Also, in the case of a chain of PRs, when changes are made to the base PR, that means making changes to all subsequent PRs.

As mentioned above, in the end, it might take more time to have all changes (what really matters) merged into staging 

> As an example, I rebased https://github.com/sablier-labs/v2-core/pull/1091 from staging in less than 30 minutes, and it can now also be reviewed and merged independently, 


This is a small PR, and it’s understandable that it didn’t take much time (which I agree is not that relevant to handle in sequence). However, I come from the singleton and unlock amounts PR, where rebasing took a lot of time (1 day).

> which otherwise could stay open for as long as the other PRs are open. Another example is that https://github.com/sablier-labs/v2-core/pull/1087 can be worked upon, reviewed and merged independently of other PRs opened before it.

IMO, as said above, that's not that relevant when we merge it if we don't have all changes ready for auditing. But I understand the argument since it is a small PR

> From my experience, its easier to deal with independent PRs which I also prefer as much as possible, which you can review and merge independent of other PRs. 

And from my experience, when it comes to more complex PRs, it’s generally easier to handle them sequentially.”

> speeds up closing the issues faster without getting blocked by the base PR

not necessarily

---

PS: You don’t need to reply to this now. We already discussed on Slack using the parallel option

andreivladbrg:
hmm, a weird error we are facing when compiling

image

not sure yet what causes it

smol-ninja:
> not sure yet what causes it

I will investigate. Feel free to go to bed :))

src/core/interfaces/ISablierLockupBase.sol
```diff
@@ -314,11 +322,26 @@ interface ISablierLockupBase is
     /// - `to` must not be the zero address.
     /// - `amount` must be greater than zero and must not exceed the withdrawable amount.
     /// - `to` must be the recipient if `msg.sender` is not the stream's recipient or an approved third party.
+    /// - The `msg.value` must not be less than the Sablier fee.
```
smol-ninja, line 325: This can be removed since there is no concept of Sablier fee yet.

src/core/libraries/Errors.sol
```diff
@@ -162,6 +168,9 @@ library Errors {
     /// @notice Thrown when trying to withdraw to the zero address.
     error SablierLockupBase_WithdrawToZeroAddress(uint256 streamId);
 
+    /// @notice Thrown when trying to withdraw with an insufficient fee payment.
+    error SablierLockupBase_WithdrawInsufficientFeePayment(uint256 feePaid, uint256 sablierFee);
```
smol-ninja, line 172: Unsued.

test/Base.t.sol
```diff
@@ -84,7 +93,7 @@ abstract contract Base_Test is Assertions, Calculations, DeployOptimized, Modifi
         nftDescriptorMock = new NFTDescriptorMock();
 
         // Set the Sablier fee on the Merkle factory.
-        merkleFactory.setDefaultSablierFee(defaults.DEFAULT_SABLIER_FEE());
+        merkleFactory.setDefaultSablierFee(SABLIER_FEE);
```
smol-ninja, line 96: Lets revert it since this is going to be removed in https://github.com/sablier-labs/v2-core/pull/1089

test/Base.t.sol
```diff
@@ -58,19 +61,25 @@ abstract contract Base_Test is Assertions, Calculations, DeployOptimized, Modifi
 
     function setUp() public virtual {
         // Deploy the base test contracts.
+        contractWithoutReceiveEth = new ContractWithoutReceiveEth();
+        contractWithReceiveEth = new ContractWithReceiveEth();
         dai = new ERC20Mock("Dai Stablecoin", "DAI");
         noop = new Noop();
         recipientGood = new RecipientGood();
+        vm.deal({ account: address(recipientGood), newBalance: 100 ether });
```
smol-ninja, line 69: Why does `recipientGood` need ETH?

test/core/integration/Integration.t.sol
```diff
@@ -257,4 +268,133 @@ abstract contract Integration_Test is Base_Test {
...
+        returns (uint256 streamId)
+    {
+        streamId = lockup.createWithTimestampsLT{ value: 0 }(params, tranches);
+    }
+
+    function renounce(uint256 streamId) internal {
+        lockup.renounce{ value: 0 }(streamId);
+    }
+
+    modifier balanceTest() {
```
smol-ninja, line 354: Why not put this modifier in the `Modifiers.sol` contract and override it like other examples?

test/core/integration/Integration.t.sol
```diff
@@ -257,4 +268,133 @@ abstract contract Integration_Test is Base_Test {
             "settled status call return data"
         );
     }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                  MIRROR-FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function burn(uint256 streamId) internal {
```
smol-ninja, line 276: IMO these functions are overkill. If the `value` is 0, then it is equivalent to to not passing any value at all.

```solidity
lockup.burn{ value: 0 }(streamId);
```

is same as 

```solidity
lockup.burn(streamId);
```

test/core/integration/Integration.t.sol
```diff
@@ -257,4 +268,133 @@ abstract contract Integration_Test is Base_Test {
...
+        return lockup.withdrawMax{ value: SABLIER_FEE }(streamId, to);
+    }
+
+    function withdrawMaxWithBalTest(uint256 streamId, address to) internal balanceTest returns (uint128) {
+        uint128 withdrawn = withdrawMax(streamId, to);
+        return withdrawn;
+    }
+
+    function withdrawMaxAndTransfer(uint256 streamId, address newRecipient) internal returns (uint128) {
+        return lockup.withdrawMaxAndTransfer{ value: SABLIER_FEE }(streamId, newRecipient);
```
smol-ninja, line 378: I'd be in favour of removing these helpers and use the function itself where it is needed with a non-zero `value`

test/core/integration/concrete/lockup-base/batch/batch.t.sol
```diff
@@ -35,7 +35,7 @@ contract Batch_Integration_Concrete_Test is Integration_Test {
         bytes[] memory calls = new bytes[](1);
         calls[0] = abi.encodeCall(lockup.cancel, (defaultStreamId));
 
-        lockup.batch(calls);
+        lockup.batch{ value: 0 }(calls);
```
smol-ninja, line 38: We can replace it with 

```solidity
lockup.batch(calls);
```

test/core/integration/concrete/lockup-base/burn/burn.t.sol
```diff
@@ -19,28 +19,28 @@ contract Burn_Integration_Concrete_Test is Integration_Test {
     function test_RevertGiven_PENDINGStatus() external whenNoDelegateCall givenNotNull givenNotDepletedStream {
         vm.warp({ newTimestamp: getBlockTimestamp() - 1 seconds });
         vm.expectRevert(abi.encodeWithSelector(Errors.SablierLockupBase_StreamNotDepleted.selector, defaultStreamId));
-        lockup.burn(defaultStreamId);
+        burn(defaultStreamId);
```
smol-ninja, line 22: We should revert it back.

test/core/integration/concrete/lockup-base/withdraw-fees/withdrawFees.tree
```diff
@@ -0,0 +1,19 @@
...
+   │  └── it should revert
+   └── when withdrawal address not zero
+      ├── when provided address not contract
+      │  ├── it should transfer fee collected in ETH to the provided address
+      │  └── it should set the ETH balance to 0
+      └── when provided address contract
+         ├── when provided address not implement receive eth
+         │  └── it should revert
+         └── when provided address implement receive eth
+            ├── given no fees collected
```
smol-ninja, line 15: I feel this is redundant branch. We should be able to remove it.

test/core/integration/concrete/lockup-base/withdraw-hooks/withdrawHooks.t.sol
```diff
@@ -62,11 +63,7 @@ contract WithdrawHooks_Integration_Concrete_Test is Integration_Test {
         });
 
         // Make the withdrawal.
-        lockup.withdraw({
-            streamId: differentSenderRecipientStreamId,
-            to: address(recipientGood),
-            amount: withdrawAmount
-        });
+        withdraw({ streamId: differentSenderRecipientStreamId, to: address(recipientGood), amount: withdrawAmount });
```
smol-ninja, line 66: The advantage of using `lockup.withdraw` with explicit `value` is that we can differentiate which tests is with 0 value and which are not. In the current approach, its hard to differentiate between the two by looking at the code.

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
```diff
@@ -6,7 +6,7 @@ WithdrawMultiple_Integration_Concrete_Test
    │  └── it should revert
    └── when equal arrays length
       ├── when zero array length
-      │  └── it should do nothing
+      │  └── it should charge the fee
```
smol-ninja, line 9: this is not being tested in the test contract so no need to mention that it should charge the fee.

test/core/integration/concrete/lockup-base/withdraw/withdraw.t.sol
```diff
@@ -166,18 +170,22 @@ abstract contract Withdraw_Integration_Concrete_Test is Integration_Test {
         givenNotNull
         givenNotDEPLETEDStatus
         whenWithdrawalAddressNotZero
+        givenMsgValueNotLessThanSablierFee
```
smol-ninja, line 173: Not included in the tree.

test/core/integration/concrete/lockup-base/withdraw/withdraw.tree
```diff
@@ -62,4 +62,5 @@ Withdraw_Integration_Concrete_Test
                                           ├── it should make the withdrawal
                                           ├── it should update the withdrawn amount
                                           ├── it should make Sablier run the recipient hook
+                                          ├── it should transfer the ETH to the lockup contract
```
smol-ninja, line 65: I could not find where its tested in the test code.

test/core/integration/fuzz/lockup-base/withdraw.t.sol
```diff
@@ -12,7 +12,7 @@ abstract contract Withdraw_Integration_Fuzz_Test is Integration_Test {
     /// @dev Given enough fuzz runs, all of the following scenarios will be fuzzed:
     ///
     /// - Multiple caller addresses.
-    function testFuzz_Withdraw_UnknownCaller(address caller)
+    function testFuzz_Withdraw_UnknownCaller(address payable caller)
```
smol-ninja, line 15: Why does `caller` have to be `payable`?

test/core/invariant/handlers/LockupHandler.sol
```diff
@@ -167,7 +167,7 @@ contract LockupHandler is BaseHandler {
         }
 
         // Withdraw from the stream.
-        lockup.withdraw({ streamId: currentStreamId, to: to, amount: withdrawAmount });
+        lockup.withdraw{ value: SABLIER_FEE }({ streamId: currentStreamId, to: to, amount: withdrawAmount });
```
smol-ninja, line 170: Why do we need a non-zero `SABLIER_FEE` in invariant if fee is not a part of any invariant statement?

smol-ninja (changes requested):
Great work @andreivladbrg. Left some comments below. I have also included a commit similar to https://github.com/sablier-labs/airdrops/pull/5/commits/0611cf9bf877c7665b12f0c861fd87ce5d139afd which renamed sablier fee to simply fee.

src/core/interfaces/ISablierLockupBase.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1087/commits/1f8d30b76893c80e77a6c6a1ea24737fa32b81ba

src/core/libraries/Errors.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1087/commits/1f8d30b76893c80e77a6c6a1ea24737fa32b81ba

test/Base.t.sol
andreivladbrg: Because it is used to call withdraw in some tests, and if it has a zero balance it would revert

test/core/integration/Integration.t.sol
andreivladbrg: My rationale is that the modifiers placed in that file, are the BTT modifiers

test/core/integration/Integration.t.sol
andreivladbrg: Fair enough, agree

test/core/integration/concrete/lockup-base/withdraw-hooks/withdrawHooks.t.sol
andreivladbrg: All calls to withdraw have value greater than zero. So not sure what you are referring to

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
andreivladbrg: That is incorrect, the function that runs the balanceTest modifier is being is used

test/core/invariant/handlers/LockupHandler.sol
andreivladbrg: Good point, we don't need it

test/core/integration/concrete/lockup-base/withdraw/withdraw.t.sol
andreivladbrg: We shall remove it

test/core/integration/concrete/lockup-base/withdraw/withdraw.tree
andreivladbrg: It is by using withdrawWithBalTest function

test/Base.t.sol
smol-ninja: If it calls to `withdraw` then thats a reentrancy test. Don't we use `reentrantRecipient` for that? I thought `recipientGood` is the good recipient that does not try to reenter.

test/core/integration/Integration.t.sol
smol-ninja: The original decision was to dump all modifiers to a single file. https://github.com/sablier-labs/v2-core/issues/1028#issuecomment-2315376028. And override whenever necessary. It was not related to BTT.

test/core/integration/Integration.t.sol
smol-ninja: > Do you find it useful to test the ETH balance increase in all functions

No I don't. One concrete test for `withdrawFee` is enough imo as its a fairly simple logic.

test/core/integration/concrete/lockup-base/withdraw-hooks/withdrawHooks.t.sol
smol-ninja: I see. I thought it is only in `withdraw` tests and not all calls. In that case, can we please have explicitness instead of using these helper functions? May be have non zero fee in withdraw test but 0 feel at all other places in Concrete.

test/core/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
smol-ninja: Oh. Usually we test assertions at the end of the logic. So thats where I looked for it.

test/core/integration/concrete/lockup-base/withdraw/withdraw.tree
smol-ninja: Same as https://github.com/sablier-labs/v2-core/pull/1087#discussion_r1851757246.

PaulRBerg:
Closing in favor of https://github.com/sablier-labs/v2-core/pull/1093.

PR #1088, opened 2024-11-15
refactor: `Schedule` struct

smol-ninja:
As per https://github.com/sablier-labs/v2-core/pull/1075#discussion_r1835522854


andreivladbrg (approved):
now that i see the code, i can say it looks cleaner, thanks for insisting 🫡

Issue #1090, opened 2024-11-18
Make "batch" function payable


PaulRBerg:
https://github.com/sablier-labs/company-discussions/discussions/72#discussioncomment-11292236

- [x] Mark all functions as `payable`.
- [x] Do **NOT** emit the fee in any event because the withdraw function can be called in multiple ways, e.g, directly via `withdraw`, via `withdrawMultiple`, or via `batch`, which would make the accounting difficult to handle.

We will keep track of ETH transfers through other means, e.g., Etherscan, subgraphs, etc.

smol-ninja:
Lets include this change alongside making other withdraw functions payable (`withdrawMultiple`) as a part of https://github.com/sablier-labs/v2-core/pull/1087 (cc @andreivladbrg).

andreivladbrg:
I was trying to implement this and am facing some issues. I’m afraid we won’t be able to achieve this via batch.

If we declare `batch` as `payable` and actually send a msg.value greater than 0, all functions within the `calls` array would also need to be declared as `payable`; otherwise, it would revert. Also, if we declare the `batch` function as payable (and all the other functions within calls are also `payable`), the fee paid by the caller would be `msg.value * length of the array`, which we don’t want.

This leads to the conclusion that we can’t declare it as `payable`, and we can’t use the `withdraw` function within the `batch` function.

smol-ninja:
In the for loop inside `batch`, we can compare the selectors and if it matches with `withdraw` and `withdraw` related functions, then only we would forward `msg.value` to the delegate call.

What do you think of this?

smol-ninja:
`delegatecall` forwards `msg.value`. [See here](https://medium.com/@olatunjimayowa0396/problem-with-payable-functions-using-delegatecall-inside-a-loop-c4f928314444) without specifying `value` parameters as long as the `batch` is `payable`.

andreivladbrg:
we can't do this: `address(this).delegatecall{ value: msg.value }(calls[i]);`

razgraf:
Curious, why does Uniswap's [multicall](https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/Multicall.sol) work? What's different in their implementation?

andreivladbrg:
> Curious, why does Uniswap's [multicall](https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/Multicall.sol) work? What's different in their implementation?

[...]

PR #1093, opened 2024-11-21
feat: payable and eth fees

### LINKED ISSUE #1076: ETH fee in withdraw function

PaulRBerg:
Closes #1076 and #1090

Replaces PR #1087

Also, note that I didn't write any concrete tests for checking that `payable` and `msg.value` work because this is none of our business. It's EVM and Solidity logic that we should assume works correctly.


src/abstracts/SablierLockupBase.sol
```diff
@@ -462,12 +473,29 @@ abstract contract SablierLockupBase is
         _transfer({ from: currentRecipient, to: newRecipient, tokenId: streamId });
     }
 
+    /// @inheritdoc ISablierLockupBase
+    function withdrawFees() external override {
```
smol-ninja, line 477: I see that you have changed the design of this function:
1. The function is no long admin only. So anyone can call this function and send fees to `admin`
2. In case we rotate admin address, we also will have to move ETH fee from the old address to the new address.

If we make it admin only, we can specify a particular address and send eth fee to that regardless of admin change situation.

Just to confirm if this is intentional.

test/Base.t.sol
```diff
@@ -76,7 +88,7 @@ abstract contract Base_Test is Assertions, Calculations, DeployOptimized, Modifi
         // Deploy the NFT descriptor mock.
         nftDescriptorMock = new NFTDescriptorMock();
 
-        // Create users for testing.
+        // Create users for testing. Note that this has to go after the protocol is deployed due to ERC-20 approvals.
```
smol-ninja, line 91: Not sure what you meant here. Why does it have to go after deployment?

PS: for to do tasks, we add `TODO` and use [TODO Tree](https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree) to identify them.

test/Base.t.sol
```diff
@@ -51,19 +54,28 @@ abstract contract Base_Test is Assertions, Calculations, DeployOptimized, Modifi
...
         // Deploy the base test contracts.
+        contractWithoutReceive = new ContractWithoutReceive();
+        contractWithReceive = new ContractWithReceive();
         dai = new ERC20Mock("Dai Stablecoin", "DAI");
         noop = new Noop();
         recipientGood = new RecipientGood();
         usdt = new ERC20MissingReturn("Tether USD", "USDT", 6);
 
+        // Fund the contracts that need it.
+        vm.deal({ account: address(contractWithoutReceive), newBalance: 100 ether });
```
smol-ninja, line 65: Since they receive eth, they do not need to be funded with eth

smol-ninja (approved):
Any reason to not include `payable` in `burn`?

test/Base.t.sol
smol-ninja: https://github.com/sablier-labs/v2-core/pull/1093/commits/13d8af9906bb19e767fc608543703d004856559f

src/abstracts/SablierLockupBase.sol
smol-ninja: Also in the [original specs](https://github.com/sablier-labs/v2-core/issues/1076), it is mentioned to allow admin to withdraw fee.

test/Base.t.sol
PaulRBerg: I've tried to move that code before the `deployProtocolConditionally` function, and I got an error.

The reason was that the user creation function assumes the Lockup contracts to be deployed.

test/Base.t.sol
PaulRBerg: I used to make those contracts the `msg.sender` in the `wtihdrawFees` tests, but I then switched to Alice. In any case, it works now without funding them with ETH.

PaulRBerg:
> Any reason to not include payable in burn?

No, I will add it.

PR #1094, opened 2024-11-22
feat: shape name in create lockup

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/1086.

### Changelog

- I did not fuzz shape names in Fork and Invariant tests since its only emitted in event. So, concrete tests and fuzz tests are sufficient imo.
- Re HTML injection attacks, a 32 bytes restriction on shape name has been applied. 
- The variable in the input parameter is called `shape` and not `shapeName` as I suppose the "name" keyword is implicit given that its a `string` type


test/utils/Defaults.sol
```diff
@@ -31,6 +31,7 @@ contract Defaults is Constants {
     uint256 public constant MAX_TRANCHE_COUNT = 10_000;
     uint128 public constant REFUND_AMOUNT = DEPOSIT_AMOUNT - WITHDRAW_AMOUNT;
     uint256 public constant SEGMENT_COUNT = 2;
+    string public constant SHAPE_NAME = "";
```
andreivladbrg, line 34: should we declare a non zero shape?
we could:

```solidity
string public constant SHAPE_NAME = "emitted in the event";
```

script/Init.s.sol
```diff
@@ -51,7 +51,8 @@ contract Init is BaseScript {
                     asset: asset,
                     cancelable: true,
                     transferable: true,
-                    broker: Broker(address(0), ud60x18(0))
+                    broker: Broker(address(0), ud60x18(0)),
+                    shape: "cliff"
```
andreivladbrg, line 55: should we be more precise? "Cliff Linear"

script/Init.s.sol
```diff
@@ -82,7 +83,8 @@ contract Init is BaseScript {
                 asset: asset,
                 cancelable: true,
                 transferable: true,
-                broker: Broker(address(0), ud60x18(0))
+                broker: Broker(address(0), ud60x18(0)),
+                shape: "dynamic"
```
andreivladbrg, line 87: similarly, "Exponential Dynamic"

test/integration/concrete/lockup-base/create-with-timestamps/createWithTimestamps.t.sol
```diff
@@ -68,7 +68,13 @@ abstract contract CreateWithTimestamps_Integration_Concrete_Test is Integration_
         expectRevert_DelegateCall(callData);
     }
 
-    function test_RevertWhen_BrokerFeeExceedsMaxValue() external whenNoDelegateCall {
+    function test_RevertWhen_ShapeNameExceeds32Bytes() external whenNoDelegateCall {
+        _defaultParams.createWithTimestamps.shape = "this name is longer than 32 bytes";
+        vm.expectRevert(abi.encodeWithSelector(Errors.SablierLockup_ShapeNameTooLong.selector, 33, 32));
```
andreivladbrg, line 73: similarly to fuzz, should we:
```solidity
string memory longShapeName = "this shape name is longer than 32 bytes";
_defaultParams.createWithTimestamps.shape = longShapeName;
vm.expectRevert(
    abi.encodeWithSelector(Errors.SablierLockup_ShapeNameTooLong.selector, bytes(longShapeName).length, 32)
);
```

src/libraries/Errors.sol
```diff
@@ -171,4 +171,7 @@ library Errors {
 
     /// @notice Thrown when a function is called on a stream that does not use the expected Lockup model.
     error SablierLockup_NotExpectedModel(Lockup.Model actualLockupModel, Lockup.Model expectedLockupModel);
+
+    /// @notice Thrown when the shape name is longer than max length.
+    error SablierLockup_ShapeNameTooLong(uint256 nameLength, uint256 maxLength);
```
andreivladbrg, line 176: should we name it more precisely, `ShapeNameExceeds32Bytes`?

andreivladbrg (changes requested):
Great work, left some comments below:

test/integration/concrete/lockup-base/create-with-timestamps/createWithTimestamps.t.sol
smol-ninja: The shape name is fixed: "this name is longer than 32 bytes" in concrete and its length if fixed too: 33. That's why having a `33` here instead of `bytes(longShapeName).length` also tests that name length is calculated correctly. This was the reason I used `33` here instead of calculating length on runtime.

smol-ninja:
Thanks for the review @andreivladbrg. I have incorporated all your suggestions except one. Let me know what you think about that.

test/integration/concrete/lockup-base/create-with-timestamps/createWithTimestamps.t.sol
andreivladbrg: fair enough, let's keep the current version 👍

PaulRBerg (commented):
Unfortunately, 32 bytes is not enough to prevent HTML injection. From [ChatGPT convo](https://chatgpt.com/share/67461b51-4c34-800e-94f6-9caa79c10fbe):

```html
<img src=x onerror=alert(1)>
```

I think we just have to accept the potential risk and sanitize the shape string at the UI level, cc @sablier-labs/frontend.

script/Init.s.sol
```diff
@@ -82,7 +83,8 @@ contract Init is BaseScript {
                 asset: asset,
                 cancelable: true,
                 transferable: true,
-                broker: Broker(address(0), ud60x18(0))
+                broker: Broker(address(0), ud60x18(0)),
+                shape: "Exponential Dynamic"
```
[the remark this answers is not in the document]
smol-ninja: Yes. It would.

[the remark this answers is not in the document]
smol-ninja: This is an `Init` script which is used to launch dummy streams when the lockup protocol is launched on a new chain. We can definitely refactor it to include a more popular shape, however, the time it would take to refactor (even small) would not justify the utility of this script.

PR #1095, opened 2024-11-25
test: refactor test directory to tests

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/1036

### Changelog
- Renames `test` directory to `tests
- Adds a note about Bulloak in CONTRIBUTING file
- Replaces wiki link with Bulloak homepage in README file
- Fix import path in benchmark tests

### PR Dependency
- https://github.com/sablier-labs/v2-core/pull/1094

PR #1096, opened 2024-11-26
Small fixes and polishes

andreivladbrg:
In this PR https://github.com/sablier-labs/v2-core/pull/1089/ the `script/protocol` dir was removed without updating the `fs_permissions` in `foundry.toml` and the `.gitignore` file.

Also, found out, we don't need all the assert functions declared in `Assertions`.
The `assertEq` functions for segments are already emitting the fact that we compare an array of segments/tranches.

Depends on: #1094 and #1095

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,5 +19,5 @@ out-svg
 lcov.info
 package-lock.json
 pnpm-lock.yaml
-script/protocol/*.md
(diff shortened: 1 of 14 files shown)


smol-ninja (approved):
Good observation.

PR #1097, opened 2024-11-26
Refactor asset to token

andreivladbrg:
Closes https://github.com/sablier-labs/v2-core/issues/991

Notes: in order to keep the backwards compatibility with previous versions of the lockup contracts in `LockupNFTDescriptor` I had to include the logic for the `getAsset` in the `LockupNFTDescriptor.tokenURI` function.

PR #1098, opened 2024-11-26
style: add ascii art in batch lockup

andreivladbrg:
Realized that `SablierBatchLockup` is missing the ascii art.

Depends on #1097


smol-ninja (approved):
LGTM

PR #1099, opened 2024-11-26
refactor: get underlying token

PaulRBerg:
In https://github.com/sablier-labs/v2-core/pull/1097, @andreivladbrg renamed `getAsset` to `getToken`.

`getToken` can be easily confused with the ERC-721 NFT functions.

It'd be better to rename it to `getUnderlyingToken`.


PaulRBerg:
> Which ERC-721 function can getToken be confused with?

All functions that take a `tokenId` as input.

> wouldn't it be more explicit to call it `getUnderlyingVestingToken` or `getVestingToken`?

Technically, yes, but since we have decided to keep the streaming terminology, I would prefer not to introduce the vesting terminology in this particular case.

We could do `getStreamedToken` or `getStreamingToken`, but I like `getUnderlyingToken` more.

smol-ninja (approved):
Appreciate the explanation. LGTM.

PR #1101, opened 2024-11-28
feat: use delegate-call in withdraw-multiple

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/917

### DIFF
--- a/tests/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
+++ b/tests/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
@@ -8,23 +8,12 @@ WithdrawMultiple_Integration_Concrete_Test
       ├── when zero array length
       │  └── it should do nothing
       └── when non zero array length
-         ├── when one null stream
-         │  ├── it should emit {InvalidWithdrawalInWithdrawMultiple} event for the null stream
-         │  └── it should emit 2 {WithdrawFromLockupStream} events
-         └── when no null streams
-            ├── when one DEPLETED stream
-            │  ├── it should emit {InvalidWithdrawalInWithdrawMultiple} event for the depleted stream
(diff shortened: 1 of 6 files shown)


tests/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
```diff
@@ -8,17 +8,21 @@ WithdrawMultiple_Integration_Concrete_Test
...
+         ├── when one null stream
+         │  ├── it should emit {InvalidWithdrawalInWithdrawMultiple} event for the null stream
+         │  └── it should emit 2 {WithdrawFromLockupStream} events
+         └── when no null streams
+            ├── when one DEPLETED stream
+            │  ├── it should emit {InvalidWithdrawalInWithdrawMultiple} event for the depleted stream
+            │  └── it should emit 2 {WithdrawFromLockupStream} events
+            └── when no DEPLETED streams
+               ├── when withdrawing zero amounts
+               │  ├── it should emit 1 {WithdrawFromLockupStream} event for first stream
```
andreivladbrg, line 20: i don't think we need to test each particular case when it reverts

i would rather have this:

```tree
├── when at least one ID reverts
│  ├── it should emit {InvalidWithdrawalInWithdrawMultiple} event for that ID
│  └── it should wihtdraw from the rest IDs
└── when no IDs revert
```

andreivladbrg (approved):
LGTM 👍
just one comment

tests/integration/concrete/lockup-base/withdraw-multiple/withdrawMultiple.tree
smol-ninja: Agree. I kept all those tests in the beginning to check all possible scenarios. But will do as per your suggestion.

PR #1106, opened 2024-12-03
build: simultaneously compilation

### LINKED ISSUE #1103: Get rid of "optimized" foundry profiles
Foundry has released a new version which allows you to set up multiple profiles that are compiled simultaneously. Ref [here](https://x.com/gretzke/status/1861175241814282496).

Thus, we should remove these profiles:
- https://github.com/sablier-labs/v2-core/blob/b5d35c40d1d411da81b8ba13ba2a1376126792b6/foundry.toml#L40
- https://github.com/sablier-labs/v2-core/blob/b5d35c40d1d411da81b8ba13ba2a1376126792b6/foundry.toml#L83

also, we would be able to update the `deployConditionally` logic.

---

questions: 
1. should we updated the CI reusables to not pass the foundry profile?
2. @PaulRBerg should we keep the `smt` profiles anymore? we haven’t used them at all.

andreivladbrg:
Closes #1103 

#### Changes 

build: use simultaneously compilation
build: remove no longer needed optimized profiles
build: update shell scripts
test: remove no longer needed Utils functions
test: use the interfaces in Base_Test instead of the contracts
chore: update precompiles bytecode
chore: update scripts in package.json
ci: update ci workflows to pass the default profile


andreivladbrg:
closing as explained here: https://github.com/sablier-labs/flow/pull/342#issuecomment-2517686317

PR #1108, opened 2024-12-09
Update `tokenURI` tests

### LINKED ISSUE #1071: Change contract address and expected token URI strings in tests
The idea is to no longer skip these tests after we finish all changes related to 2.0.0:

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/test/core/integration/concrete/lockup-dynamic/token-uri/tokenURI.t.sol#L20-L24

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/test/core/integration/concrete/lockup-tranched/token-uri/tokenURI.t.sol#L20-L25

https://github.com/sablier-labs/v2-core/blob/997444866a2f9279ffe4bdde59462e4598d6d87e/test/core/integration/concrete/lockup-linear/token-uri/tokenURI.t.sol#L20-L24

andreivladbrg:
Closes #1071


smol-ninja:
Can you please give a tl'dr of what you changed?

- In the issue description, it is mentioned that "The idea is to no longer skip these tests after we finish all changes related to 2.0.0" but I do not understand how you fixed that in this PR?

- Also what is that address that you used `0x923b5Ab3714FD343316aF5A5434582Fd16722523`?

- Given that we only have one contract now, can we have just one contract as `TokenURI_Lockup_Integration_Concrete_Test`  instead of three?

smol-ninja:
> we can indeed keep only one contract. hmm, which makes me think, why not mot modify the NFTDescriptor to include the lockupModel

IMO we can but it will just be an additional information. What Traders / Users need to know is already a part of the SVG. I think the end users do not care about the specific stream model but only the shape which can only be visualised on the UI. 

Regarding your other points, I have sent you a msg on Slack for some clarification.

andreivladbrg:
fair enough, let's keep the current version

smol-ninja:
> let's keep the current version

What about DRY'ify it with a single test contract `TokenURI_Lockup_Integration_Concrete_Test`?

andreivladbrg:
> What about DRY'ify it with a single test contract TokenURI_Lockup_Integration_Concrete_Test?

yeah, i wasn’t referring to the `tests` but to the `NFTDescriptor` contract (this was the subject of the matter).

as for the tests, like i mentioned in my initial response, we can have one contract

tests/integration/concrete/lockup-base/token-uri/tokenURI.t.sol
```diff
@@ -0,0 +1,58 @@
...
...ZGl0aXZlPSJzdW0iIGF0dHJpYnV0ZU5hbWU9InN0YXJ0T2Zmc2V0IiBiZWdpbj0iMHMiIGR1cj0iNTBzIiBmcm9tPSIwJSIgcmVwZWF0Q291bnQ9ImluZGVmaW5pdGUiIHRvPSIxMDAlIi8+MHhmNjI4NDlmOWEwYjViZjI5MTNiMzk2MDk4ZjdjNzAxOWI1MWE4MjBhIOKAoiBFUkMyMDwvdGV4dFBhdGg+PHRleHRQYXRoIHN0YXJ0T2Zmc2V0PSI1MCUiIGhyZWY9IiNGbG9hdGluZ1RleHQiIGZpbGw9IiNmZmYiIGZvbnQtZmFtaWx5PSInQ291cmllciBOZXcnLEFyaWFsLG1vbm9zcGFjZSIgZmlsbC1vcGFjaXR5PSIuOCIgZm9udC1zaXplPSIyNnB4Ij48YW5pbWF0ZSBhZGRpdGl2ZT0ic3VtIiBhdHRyaWJ1dGVOYW1lPSJzdGFydE9mZnNldCIgYmVnaW49IjBzIiBkdXI9IjUwcyIgZnJvbT0iMCUiIHJlcGVhdENvdW50PSJpbmRlZmluaXRlIiB0bz0iMTAwJSIvPjB4ZjYyODQ5ZjlhMGI1YmYyOTEzYjM5NjA5OGY3YzcwMTliNTFhODIwYSDigKIgRVJDMjA8L3RleHRQYXRoPjwvdGV4dD48dXNlIGhyZWY9IiNHbG93IiBmaWxsLW9wYWNpdHk9Ii45Ii8+PHVzZSBocmVmPSIjR2xvdyIgeD0iMTAwMCIgeT0iMTAwMCIgZmlsbC1vcGFjaXR5PSIuOSIvPjx1c2UgaHJlZj0iI0xvZ28iIHg9IjE3MCIgeT0iMTcwIiB0cmFuc2Zvcm09InNjYWxlKC42KSIvPjx1c2UgaHJlZj0iI0hvdXJnbGFzcyIgeD0iMTUwIiB5PSI5MCIgdHJhbnNmb3JtPSJyb3RhdGUoMTApIiB0cmFuc2Zvcm0tb3JpZ2luPSI1MDAgNTAwIi8+PHVzZSBocmVmPSIjUHJvZ3Jlc3MiIHg9IjE0NCIgeT0iNzkwIi8+PHVzZSBocmVmPSIjU3RhdHVzIiB4PSIzNjgiIHk9Ijc5MCIvPjx1c2UgaHJlZj0iI0Ftb3VudCIgeD0iNTY4IiB5PSI3OTAiLz48dXNlIGhyZWY9IiNEdXJhdGlvbiIgeD0iNzA0IiB5PSI3OTAiLz48L3N2Zz4="}';
```
smol-ninja, line 46: Why is `"value": "ERC20"` now? Previously it was "value": "DAI"? In the comment, it is said "The stream token must have the DAI symbol".

smol-ninja, line 46: `"name":"Sablier Sablier Lockup"`: `Sablier` is repeated. It should be `"name":"Sablier Lockup"`

smol-ninja (changes requested):
I used https://www.diffchecker.com/ to compare the file. Two comments below.

tests/integration/concrete/lockup-base/token-uri/tokenURI.t.sol
andreivladbrg: good catch, that's an issue introduced in this https://github.com/sablier-labs/v2-core/pull/994/ PR and the singletone PR. 

we shall remove this:

https://github.com/sablier-labs/v2-core/blob/9ef51a35fe0abe97fd5e4e6e93074f5212e37cf8/src/LockupNFTDescriptor.sol#L326-L330

change this 
https://github.com/sablier-labs/v2-core/blob/9ef51a35fe0abe97fd5e4e6e93074f5212e37cf8/src/LockupNFTDescriptor.sol#L306-L323


to:


```solidity
return string.concat(
    "This NFT represents a payment stream in a Sablier Lockup",
    " contract. The owner of this NFT can withdraw the streamed tokens, which are denominated in ",
    tokenSymbol,
    ".\\n\\n- Stream ID: ",
    streamId,
    "\\n- ",
    "Sablier Lockup Address: ",
    lockupStringified,
    "\\n- ",
    tokenSymbol,
    " Address: ",
    tokenAddress,
    "\\n\\n",
    info
);
```

andreivladbrg: just realized, this would not be backwards compatible with previous versions

PaulRBerg:
While we are at this, we should remove the `payment` qualifier in `payment stream` to avoid any confusion with the UI (the UI now refers to Flow streams as 'payment streams'). So we should say just 'stream'.

smol-ninja:
As decided on Slack, @andreivladbrg can you please include the change that removes lines of code related to backward compatibility? We can also delete the fork test now.

smol-ninja:
Closing it in favour of https://github.com/sablier-labs/v2-core/pull/1113.

PR #1116, opened 2024-12-12
refactor: minor rewording in NFT descriptor

PaulRBerg:


smol-ninja (changes requested):
This needs to be updated in tests as well.

smol-ninja (approved):

PR #1117, opened 2024-12-12
build: lower optimizer runs

andreivladbrg:
The size of the `SablierLockup` contract passes the limit:

```
| SablierLockup                 |           25,069 |            26,085 |               -493 |              23,067 |
```

So we need to lower the optimizer runs to 800.
Also, update the package json version to `2.0.0`

### DIFF
--- a/foundry.toml
+++ b/foundry.toml
@@ -10,7 +10,7 @@
   ]
   gas_limit = 9223372036854775807
   optimizer = true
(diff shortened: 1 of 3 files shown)


smol-ninja (changes requested):
Version change in `test_ConstructCreate2Salt` as well. Not sure why the PR did not trigger the CI

PR #1126, opened 2024-12-23
feat: refactor batch to revert and return values

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/v2-core/issues/1123
- Closes https://github.com/sablier-labs/v2-core/issues/1124

Regarding tests, because the `batch` function now returns values, I have taken a different approach to test it with Lockup functions. Please let me know if I have missed anything.


smol-ninja:
Pushed a new commit. Now, I think, we have the most comprehensive tests for `batch` function.

tests/integration/Integration.t.sol
```diff
@@ -108,6 +100,38 @@ abstract contract Integration_Test is Base_Test {
...
+        vm.label({ account: address(recipientInterfaceIDIncorrect), newLabel: "Recipient Interface ID Incorrect" });
+        vm.label({ account: address(recipientInterfaceIDMissing), newLabel: "Recipient Interface ID Missing" });
+        vm.label({ account: address(recipientInvalidSelector), newLabel: "Recipient Invalid Selector" });
+        vm.label({ account: address(recipientReentrant), newLabel: "Recipient Reentrant" });
+        vm.label({ account: address(recipientReverting), newLabel: "Recipient Reverting" });
+
+        // Allow the recipients to Hook.
+        resetPrank({ msgSender: users.admin });
+        lockup.allowToHook(address(recipientReverting));
+        resetPrank({ msgSender: users.sender });
```
PaulRBerg, line 132: Why allow only `recipientReverting`, and not the rest?

tests/mocks/BatchMock.sol
```diff
@@ -0,0 +1,48 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22;
+
+import { Batch } from "src/abstracts/Batch.sol";
+
+contract BatchMock is Batch {
+    error InvalidNumber(uint256);
+
+    uint256 internal _number = 10;
```
PaulRBerg, line 9: Let's use 42 instead.

tests/unit/concrete/batch/batch.tree
```diff
@@ -0,0 +1,24 @@
+Batch_Unit_Concrete_Test
+├── when function does not exist
+│  └── it should revert
+└── when function exists
+   ├── given non state changing function
```
PaulRBerg, line 5: this should be a `when`

All the other branches refer to properties of the same function.

tests/integration/concrete/batch/batch.t.sol
```diff
@@ -0,0 +1,166 @@
...
+        bytes[] memory calls = new bytes[](5);
+        // It should not return anything.
+        calls[0] = abi.encodeCall(lockup.withdraw, (defaultStreamId, users.recipient, 1));
+        // It should return the withdrawn amount.
+        calls[1] = abi.encodeCall(lockup.withdrawMax, (defaultStreamId, users.recipient));
+
+        uint256[] memory streamIds = new uint256[](2);
+        streamIds[0] = streamIds[1] = notCancelableStreamId;
+        uint128[] memory amounts = new uint128[](2);
+        amounts[0] = amounts[1] = 1;
```
PaulRBerg, line 146: We can use `Solarray` here.

PaulRBerg (approved):
great work, made some suggestions

tests/integration/Integration.t.sol
smol-ninja: I have to test that. I tried moving all of them here then some tests failed because some of the revert branches require them not to be allowed to hook at first.

However, I will isolate those recipient contract and move the rest here. Will do it through a separate PR

tests/unit/concrete/batch/batch.tree
smol-ninja: Good point.

tests/integration/concrete/batch/batch.t.sol
```diff
@@ -0,0 +1,165 @@
...
+        assertEq(abi.decode(results[1], (uint256)), expectedNextStreamId + 1, "batch results[1]");
+        assertEq(abi.decode(results[2], (uint256)), expectedNextStreamId + 2, "batch results[2]");
+        assertEq(abi.decode(results[3], (uint256)), expectedNextStreamId + 3, "batch results[3]");
+        assertEq(abi.decode(results[4], (uint256)), expectedNextStreamId + 4, "batch results[4]");
+        assertEq(abi.decode(results[5], (uint256)), expectedNextStreamId + 5, "batch results[5]");
+        assertEq(address(lockup).balance, initialEthBalance + 1 wei, "batch contract balance");
+    }
+
+    /// @dev The batch call includes:
+    /// - Payable functions
```
smol-ninja, line 100: It should be `ETH value` because the point is to mention that `batch` contains `{ value: 1 wei }` when making the call. Similar, in other test, the batch may not contain `{ value: 1 wei }`. 

To signal that functions are all payable in the `calls` array, I have used the test name as `test_BatchPayable_`

tests/integration/concrete/lockup-base/renounce-multiple/renounceMultiple.t.sol
```diff
@@ -31,13 +31,13 @@ contract RenounceMultiple_Integration_Concrete_Test is Integration_Test {
...
 
-    function test_RevertGiven_AtleastOneNullStream() external whenNoDelegateCall whenNonZeroArrayLength {
+    function test_RevertGiven_AtLeastOneNullStream() external whenNoDelegateCall whenNonZeroArrayLength {
         expectRevert_Null({
             callData: abi.encodeCall(lockup.renounceMultiple, Solarray.uint256s(streamIds[0], nullStreamId))
         });
     }
 
-    function test_RevertGiven_AtleastOneColdStream()
+    function test_RevertGiven_AtLeastOneColdStream()
```
smol-ninja, line 40: Shouldn't it be `Atleast`?

tests/integration/concrete/batch/batch.t.sol
PaulRBerg: Alright! Feel free to push a commit to fix this.

tests/integration/concrete/lockup-base/renounce-multiple/renounceMultiple.t.sol
PaulRBerg: No, "at least" is written separately, I've also updated the BTT tree

tests/integration/concrete/batch/batch.t.sol
```diff
@@ -0,0 +1,165 @@
...
+        streamIds[0] = recipientRevertStreamId;
+        calls[3] = abi.encodeCall(lockup.renounceMultiple, (streamIds));
+
+        bytes[] memory results = lockup.batch{ value: 1 wei }(calls);
+
+        assertEq(results.length, 4);
+        assertEq(results[0], hex"");
+        assertEq(results[1], hex"");
+        assertEq(results[2], hex"");
+        assertEq(results[3], hex"");
```
andreivladbrg, line 126: we don't need to specifically add `hex`

same above & below

tests/unit/concrete/batch/batch.t.sol
andreivladbrg, line 1: i guess the reason for not adding the modifiers to the `Modifiers` file is because we will eventually move these tests into the utils repo, right?

tests/unit/concrete/batch/batch.tree
```diff
@@ -0,0 +1,24 @@
+Batch_Unit_Concrete_Test
+├── when function does not exist
+│  └── it should revert
+└── when function exists
+   ├── when non state-changing function
```
andreivladbrg, line 5: it would be the only place where we have "-" in the test branch (not talking about what _it_ should do)

should we change it to not use "-"?

andreivladbrg (commented):
great work, left some comments below:

tests/unit/concrete/batch/batch.t.sol
```diff
@@ -0,0 +1,139 @@
...
+
+        batchMock = new BatchMock();
+    }
+
+    function test_RevertWhen_FunctionDoesNotExist() external {
+        calls = new bytes[](1);
+        calls[0] = abi.encodeWithSignature("nonExistentFunction()");
+
+        // It should revert.
+        vm.expectRevert(bytes(""));
```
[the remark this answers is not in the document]
smol-ninja: To capture the specific error. vm.expectRevert() only looks for error regardless of revert data.

This ensures that revert contains no data.

tests/integration/concrete/batch/batch.t.sol
smol-ninja: How would you compare that the result is empty then?

tests/unit/concrete/batch/batch.t.sol
smol-ninja: Yes.

src/abstracts/Batch.sol
```diff
@@ -13,15 +13,28 @@ abstract contract Batch is IBatch {
...
+    /// @dev Since `msg.value` can be reused across calls, be VERY CAREFUL when using it. Refer to
+    /// https://paradigm.xyz/2021/08/two-rights-might-make-a-wrong for more information.
+    function batch(bytes[] calldata calls) external payable override returns (bytes[] memory results) {
         uint256 count = calls.length;
+        results = new bytes[](count);
 
         for (uint256 i = 0; i < count; ++i) {
             (bool success, bytes memory result) = address(this).delegatecall(calls[i]);
+
+            // Check: If the delegate call failed, load and bubble up the revert data.
```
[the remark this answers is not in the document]
smol-ninja: @PaulRBerg made this change so mb he can answer.

tests/unit/concrete/batch/batch.tree
smol-ninja: @PaulRBerg made this change so mb he can answer.

src/abstracts/Batch.sol
```diff
@@ -13,15 +13,28 @@ abstract contract Batch is IBatch {
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc IBatch
-    /// @dev The `msg.value` should not be used on any method called in the batch.
-    function batch(bytes[] calldata calls) external payable override {
+    /// @dev Since `msg.value` can be reused across calls, be VERY CAREFUL when using it. Refer to
+    /// https://paradigm.xyz/2021/08/two-rights-might-make-a-wrong for more information.
```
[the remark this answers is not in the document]
PaulRBerg: Explicitness is good in this case.

Yes, let's remove the "Forked from" comment.

src/abstracts/Batch.sol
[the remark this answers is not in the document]
PaulRBerg: We can use `delegatecall` as a single word, I don't mind.

tests/unit/concrete/batch/batch.tree
PaulRBerg: We can remove the hyphen.

PaulRBerg:
I won't make any edits here, feel free to push a new commit to address the feedback above.

tests/integration/concrete/batch/batch.t.sol
andreivladbrg: `assertEq(results[0], "");`

src/abstracts/Batch.sol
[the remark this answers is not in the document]
andreivladbrg: fair enough, let's keep both

i also agree we can remove the "fork" comment

tests/integration/concrete/batch/batch.t.sol
smol-ninja: Thanks for the suggestion, I tested and it works. I remember trying it but it was not working which led me to add `hex`. May be I did not try this but `0x`.

smol-ninja:
I have included all the suggestions. Let me know if it looks good now.

PaulRBerg (approved):
yes ser, looks good now

PaulRBerg:
You have the honor to squash & merge.

PR #1131, opened 2025-01-05
Address audit findings 3, 5, and 11

andreivladbrg:
### Changes 
- refactor: add blockTimestamp param in VestingMath functions
- refactor: remove unchecked in calculateStreamedPercentage
- docs: add assumptions in VestingMath function
- docs: improve explanatory comments

Addresses these audit findings:
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/3
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/5
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/11

### DIFF
--- a/src/SablierLockup.sol
+++ b/src/SablierLockup.sol
@@ -297,7 +297,7 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
 
     /// @inheritdoc SablierLockupBase
     function _calculateStreamedAmount(uint256 streamId) internal view override returns (uint128) {
-        // Load the stream's parameters in memory.
+        // Load in memory the parameters used in {VestingMath}.
         uint40 blockTimestamp = uint40(block.timestamp);
         uint128 depositedAmount = _streams[streamId].amounts.deposited;
         Lockup.Model lockupModel = _streams[streamId].lockupModel;
@@ -305,7 +305,7 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
         Lockup.Timestamps memory timestamps =
             Lockup.Timestamps({ start: _streams[streamId].startTime, end: _streams[streamId].endTime });
 
-        // Calculate streamed amount for the Lockup Dynamic model.
+        // Calculate the streamed amount for the Lockup Dynamic model.
         if (lockupModel == Lockup.Model.LOCKUP_DYNAMIC) {
             streamedAmount = VestingMath.calculateLockupDynamicStreamedAmount({
                 depositedAmount: depositedAmount,
@@ -315,7 +315,7 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
                 withdrawnAmount: _streams[streamId].amounts.withdrawn
             });
         }
-        // Calculate streamed amount for the Lockup Linear model.
+        // Calculate the streamed amount for the Lockup Linear model.
         else if (lockupModel == Lockup.Model.LOCKUP_LINEAR) {
             streamedAmount = VestingMath.calculateLockupLinearStreamedAmount({
                 depositedAmount: depositedAmount,
@@ -326,7 +326,7 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
                 withdrawnAmount: _streams[streamId].amounts.withdrawn
             });
         }
-        // Calculate streamed amount for the Lockup Tranched model.
+        // Calculate the streamed amount for the Lockup Tranched model.
         else if (lockupModel == Lockup.Model.LOCKUP_TRANCHED) {
             streamedAmount = VestingMath.calculateLockupTranchedStreamedAmount({
                 depositedAmount: depositedAmount,
--- a/src/SablierLockup.sol
+++ b/src/SablierLockup.sol
@@ -297,30 +297,21 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
 
     /// @inheritdoc SablierLockupBase
     function _calculateStreamedAmount(uint256 streamId) internal view override returns (uint128) {
+        // Load the stream's parameters in memory.
+        uint40 blockTimestamp = uint40(block.timestamp);
+        uint128 depositedAmount = _streams[streamId].amounts.deposited;
+        Lockup.Model lockupModel = _streams[streamId].lockupModel;
+        uint128 streamedAmount;
         Lockup.Timestamps memory timestamps =
             Lockup.Timestamps({ start: _streams[streamId].startTime, end: _streams[streamId].endTime });
-
-        // If the start time is in the future, return zero.
-        uint40 blockTimestamp = uint40(block.timestamp);
-        if (timestamps.start > blockTimestamp) {
-            return 0;
-        }
-
-        // If the end time is not in the future, return the deposited amount.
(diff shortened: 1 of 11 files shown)


src/libraries/VestingMath.sol
```diff
@@ -33,18 +33,23 @@ library VestingMath {
     /// 2. The stream's start time must be in the past so that the calculations below do not overflow.
     /// 3. The stream's end time must be in the future so that the loop below does not panic with an "index out of
     /// bounds" error.
+    ///
+    /// Assumptions:
+    /// 1. The sum of all segment amounts does not overflow uint128.
+    /// 2. The segment timestamps are ordered chronologically.
+    /// 3. There are no duplicate segment timestamps.
+    /// 4. The number of segments does not exceed MAX_COUNT.
```
smol-ninja, line 41: I think its not relevant to the library since it can take any number of segments. `MAX_COUNT` is specific to the contract calling the library function and is not even defined in the context of the library.

src/libraries/VestingMath.sol
```diff
@@ -177,13 +188,20 @@ library VestingMath {
     /// Where:
     ///
     /// - $\Sigma(eta)$ is the sum of all vested tranches' amounts.
-    function calculateLockupTranchedStreamedAmount(LockupTranched.Tranche[] memory tranches)
+    ///
+    /// Assumptions:
+    /// 1. The sum of all tranche amounts does not overflow uint128.
+    /// 2. The tranche timestamps are ordered chronologically.
+    /// 3. There are no duplicate tranche timestamps.
+    /// 4. The number of segments does not exceed MAX_COUNT.
```
smol-ninja, line 196: Same as above.

smol-ninja (changes requested):
Looks good @andreivladbrg. Only one comment below:

src/libraries/VestingMath.sol
andreivladbrg: I also thought about this, but ultimately, all the assumptions can be viewed as contract-related. For example, the function would execute easily if there were only duplicate segment timestamps

smol-ninja: The first 3 points are fine, they are assumptions to make the best use of this library but `MAX_COUNT` depends on the EVM block limit so its only specific to the Lockup contract where it matters. For this library to work, `MAX_COUNT` is not required at all.

andreivladbrg: > the best use of this library 

I am ok not including it

@PaulRBerg what's your take on this?

smol-ninja: Update: we talked about this with auditors and they agreed to remove `MAX_COUNT` assumption from the library.

src/SablierLockup.sol
```diff
@@ -297,45 +297,43 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
 
     /// @inheritdoc SablierLockupBase
     function _calculateStreamedAmount(uint256 streamId) internal view override returns (uint128) {
-        Lockup.Timestamps memory timestamps =
-            Lockup.Timestamps({ start: _streams[streamId].startTime, end: _streams[streamId].endTime });
-
-        // If the start time is in the future, return zero.
+        // Load the stream's parameters in memory.
```
smol-ninja, line 300: The next line loads `block.timestamp` into memory.

src/libraries/VestingMath.sol
```diff
@@ -177,12 +209,30 @@ library VestingMath {
...
+        LockupTranched.Tranche[] memory tranches
+    )
         public
-        view
+        pure
         returns (uint128)
     {
-        uint256 blockTimestamp = block.timestamp;
+        // If the start time is in the future, return zero.
+        if (timestamps.start > blockTimestamp) {
```
smol-ninja, line 228: This if block is redundant. We already have the following logic in this function:

```solidity
        if (tranches[0].timestamp > blockTimestamp) {
            return 0;
        }
```

smol-ninja (changes requested):
Nice work Andrei. Left few comments below. I also realized that we do not have anything like the following in `calculateLockupTranchedStreamedAmount` while we have it in LL and LD calculations.

```solidity
            // Although the streamed amount should never exceed the deposited amount, this condition is checked
            // without asserting to avoid locking tokens in case of a bug. If this situation occurs, the withdrawn
            // amount is considered to be the streamed amount, and the stream is effectively frozen.
            if (streamedAmount > depositedAmount) {
                return withdrawnAmount;
            }
```

src/SablierLockup.sol
andreivladbrg: correct. what do you say about: "// Load in memory the parameters used in VestingMath."

src/libraries/VestingMath.sol
andreivladbrg: That’s correct—in our case, this check can be removed. However, I left it intentionally because if someone calls this function with an incorrect set of tranches but the correct start time, it would return 0.

andreivladbrg:
>  I also realized that we do not have anything like the following in calculateLockupTranchedStreamedAmount while we have it in LL and LD calculations.


Correct. The reason we have these checks in linear/dynamic is because they involve math-sensitive operations in Solidity, the use of `elapsedTimePercentage` (division followed by multiplication). In the case of tranche, we only perform a sum (i.e. addition), which should be safe.

src/libraries/VestingMath.sol
```diff
@@ -109,19 +116,23 @@ library VestingMath {
     /// - $sa$ is the streamable amount, i.e. deposited amount minus unlock amounts' sum.
```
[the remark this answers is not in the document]
andreivladbrg: that is not the same

streamable amount is: `deposited - unlockAmounts.start - unlockAmounts.cliff`

i.e. the amount that is multiplied by the `elapsedTimePercentage`

and the streamed amount is: `streamableAmount + unlockAmountsSum`

src/libraries/VestingMath.sol
smol-ninja: "start time" does not exist in the context of tranched streams. So, one solution could be to only pass `timestamps.end` as the input parameter but then I suppose, for the API consistencies, you decided to pass `timestamps` struct, which is good choice. 

How about that, instead of reverting on invalid `timestamps.start` value, we add it as an assumption i.e. this function does not check for `timestamps.start` value?

src/libraries/VestingMath.sol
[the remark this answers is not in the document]
PaulRBerg: Alright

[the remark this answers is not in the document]
PaulRBerg: We should have used a different acronym then, not $sa$, but something else to differentiate it from `streamed amount`.

src/libraries/VestingMath.sol
smol-ninja: shw from cantina also recommended something similar: https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/5#comment-1583e631-54c2-45b8-aff3-4422bba95142

PR #1132, opened 2025-01-06
fix: address audit findings 4, 8, 10

smol-ninja:
This PR addresses the following audit findings:
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/4
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/8
- https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/10

### Note
Regarding 10th finding about `MetadataUpdate` event, I have removed it from `renounce` but kept it in `_update` since transfering stream changes recipient and recipient is a part of stream's metadata.


PaulRBerg:
> transfering stream changes recipient and recipient is a part of stream's metadata

The recipient per se is not included in the NFT metadata, but it would be helpful to update the metadata when the NFT changes hands so that the new recipient can see the most up-to-date information on marketplaces like OpenSea.

.gitignore
```diff
@@ -1,14 +1,18 @@
 # directories
 artifacts
+artifacts-zk
 broadcast
 cache
+cache_hardhat-zk
```
PaulRBerg, line 6: Can we use a hyphen here instead of an underscore?

src/SablierLockup.sol
```diff
@@ -141,12 +141,14 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
         noDelegateCall
         returns (uint256 streamId)
     {
+        uint40 startTime = uint40(block.timestamp);
```
PaulRBerg, line 144: Let's add a comment above this:

```
// Use the block timestamp as the start time.
```

So that all lines are nicely commented.

src/SablierLockup.sol
```diff
@@ -221,12 +219,14 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
...
```
PaulRBerg, line 222: Add the same comment here.

src/interfaces/ISablierLockupBase.sol
```diff
@@ -228,7 +228,7 @@ interface ISablierLockupBase is
 
     /// @notice Burns the NFT associated with the stream.
     ///
-    /// @dev Emits a {Transfer} event.
+    /// @dev Emits a {Transfer}, and {MetadataUpdate} event.
```
PaulRBerg, line 231: No need for a comma when there are only two items.

tests/integration/concrete/lockup-base/renounce-multiple/renounceMultiple.t.sol
```diff
@@ -90,15 +89,11 @@ contract RenounceMultiple_Integration_Concrete_Test is Integration_Test {
         givenNoColdStreams
         whenCallerAuthorizedForAllStreams
     {
-        // It should emit {MetadataUpdate} and {RenounceLockupStream} events for both the streams.
+        // It should emit {RenounceLockupStream} events for both the streams.
```
PaulRBerg, line 92: Just `both streams` would be enough → no need for `the`

tests/integration/concrete/lockup-linear/create-with-durations-ll/createWithDurationsLL.tree
```diff
@@ -3,20 +3,14 @@ CreateWithDurationsLL_Integration_Concrete_Test
 │  └── it should revert
 └── when no delegate call
    ├── when cliff duration not zero
-   │  ├── when cliff time calculation overflows
-   │  │  └── it should revert
```
PaulRBerg, line 7: hmm interesting, so when there was an overflow, both of the following scenarios were possible: (i) revert and (ii) non-revert. We had only tested the former scenario.

PaulRBerg (changes requested):
feedback below

.gitignore
smol-ninja: The folder, `cache_hardhat-zk`, is created by the hardhat so its not in our control.

tests/integration/concrete/lockup-linear/create-with-durations-ll/createWithDurationsLL.tree
smol-ninja: Yes.

- revert if overflow leads to cliff time < start time and cliff time > 0
- non-revert if overflow leads to cliff time = 0 because a cliff time of 0 implies no cliff

andreivladbrg: Yes, that’s correct. The case where it was not reverting occurred when the cliff time was set to 0, and unfortunately it wasn't tested

smol-ninja:
Incorporated all your suggestions @PaulRBerg.

.gitignore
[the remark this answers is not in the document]
andreivladbrg: they should be a part of the zk related branch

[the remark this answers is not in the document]
smol-ninja: Why cant they be a part of this as well? Switching between the two branches creates these hardhat folders which then need to be removed manually so that I do not push them accidentally on these branches. I am sure you must have experienced the same problem :))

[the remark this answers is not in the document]
andreivladbrg: fair enough

andreivladbrg (approved):
LGTM👍

tests/integration/concrete/lockup-linear/create-with-durations-ll/createWithDurationsLL.tree
PaulRBerg: Got it thanks for explaining.

Why hasn't it been caught by the fuzz tests ??

PaulRBerg (approved):
LFG

tests/integration/concrete/lockup-linear/create-with-durations-ll/createWithDurationsLL.tree
andreivladbrg: well, it was not reverting, there nothing to be caught

PaulRBerg: With a fuzz test that looked for reverts, the test would have failed if there had been no revert.

PR #1136, opened 2025-01-09
Polish assumptions

andreivladbrg:
According to this comment: [https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/5#comment-1583e631-54c2-45b8-aff3-4422bba95142](https://cantina.xyz/code/52d6aa1e-6aa4-4e73-96ed-6077652cf0f1/findings/5#comment-1583e631-54c2-45b8-aff3-4422bba95142)

I have added one more assumption in the dynamic and tranched functions. 

Also, I have updated the first assumption in the linear function to specify that the sum of unlock amounts must be less than or equal to the deposit amount.

### DIFF
--- a/src/libraries/VestingMath.sol
+++ b/src/libraries/VestingMath.sol
@@ -36,8 +36,10 @@ library VestingMath {
     ///
     /// Assumptions:
     /// 1. The sum of all segment amounts does not overflow uint128.
-    /// 2. The segment timestamps are ordered chronologically.
-    /// 3. There are no duplicate segment timestamps.
+    /// 2. The first segment's timestamp is greater than the start time and the last segment's timestamp
+    /// equals the end time.
+    /// 3. The segment timestamps are ordered chronologically.
+    /// 4. There are no duplicate segment timestamps.
     function calculateLockupDynamicStreamedAmount(
         uint128 depositedAmount,
         LockupDynamic.Segment[] memory segments,
@@ -128,7 +130,8 @@ library VestingMath {
     /// - $c$ is the cliff unlock amount.
     ///
     /// Assumptions:
-    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128.
+    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128 and is less than or equal to
+    /// deposit amount.
     /// 2. The start time is before the end time, and the block timestamp is between the start and end times.
     /// 3. If the cliff time is not zero, it is after the start time and before the end time.
     function calculateLockupLinearStreamedAmount(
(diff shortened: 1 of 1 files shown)


src/libraries/VestingMath.sol
```diff
@@ -128,7 +130,8 @@ library VestingMath {
     /// - $c$ is the cliff unlock amount.
     ///
     /// Assumptions:
-    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128.
+    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128 and is less than or equal to
+    /// deposit amount.
     /// 2. The start time is before the end time, and the block timestamp is between the start and end times.
```
smol-ninja, line 135: "the block timestamp is between the start and end times" is not accurate. block timestamp can be anywhere. If its less than start time, then function return 0. If its more than end time, then function returns deposited amount. So lets just keep "The start time is before the end time".

src/libraries/VestingMath.sol
```diff
@@ -36,8 +36,10 @@ library VestingMath {
     ///
     /// Assumptions:
     /// 1. The sum of all segment amounts does not overflow uint128.
-    /// 2. The segment timestamps are ordered chronologically.
-    /// 3. There are no duplicate segment timestamps.
+    /// 2. The first segment's timestamp is greater than the start time and the last segment's timestamp
+    /// equals the end time.
+    /// 3. The segment timestamps are ordered chronologically.
```
smol-ninja, line 41:
```suggestion
    /// 3. The segment timestamps are monotonically increasing.
```

Saying "ordered chronologically" does not mean an increasing order.

src/libraries/VestingMath.sol
```diff
@@ -36,8 +36,10 @@ library VestingMath {
...
+    /// 4. There are no duplicate segment timestamps.
```
smol-ninja, line 42: We can remove this if we use "monotonically increasing"

src/libraries/VestingMath.sol
```diff
@@ -211,8 +214,10 @@ library VestingMath {
     ///
     /// Assumptions:
     /// 1. The sum of all tranche amounts does not overflow uint128.
-    /// 2. The tranche timestamps are ordered chronologically.
-    /// 3. There are no duplicate tranche timestamps.
+    /// 2. The first tranche's timestamp is greater than the start time and the last tranche's timestamp
+    /// equals the end time.
+    /// 3. The tranche timestamps are ordered chronologically.
```
smol-ninja, line 219: monotonically increasing

smol-ninja (changes requested):

src/libraries/VestingMath.sol
andreivladbrg: very correct, nice catch

src/libraries/VestingMath.sol
andreivladbrg: i agree that we can remove 4th point, but i would rather say: `are ordered in ascending order` 
wdyt?

smol-ninja: Sounds good

andreivladbrg:
@smol-ninja just pushed a new commit, lmk if it looks good

smol-ninja (approved):
LGTM

PR #1140, opened 2025-01-10
docs: nitpicks

PaulRBerg:
polishing https://github.com/sablier-labs/v2-core/pull/1136

### DIFF
--- a/src/libraries/VestingMath.sol
+++ b/src/libraries/VestingMath.sol
@@ -129,7 +129,7 @@ library VestingMath {
     /// - $c$ is the cliff unlock amount.
     ///
     /// Assumptions:
-    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128, and is less than or equal to
+    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128 and is less than or equal to
     /// the deposit amount.
     /// 2. The start time is before the end time.
     /// 3. If the cliff time is not zero, it is after the start time and before the end time.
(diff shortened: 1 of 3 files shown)


smol-ninja (approved):

src/libraries/VestingMath.sol
```diff
@@ -129,8 +129,8 @@ library VestingMath {
     /// - $c$ is the cliff unlock amount.
     ///
     /// Assumptions:
-    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128 and is less than or equal to
-    /// deposit amount.
+    /// 1. The sum of the unlock amounts (start and cliff) does not overflow uint128, and is less than or equal to
```
andreivladbrg, line 132: do we need comma here?

andreivladbrg (commented):
lgtm, just one question

src/libraries/VestingMath.sol
smol-ninja: [ChatGPT](https://chatgpt.com/share/67828fd9-c8b8-8000-9266-e4dd8b9730c1) says that don't since `and` in this context is not connecting to independent clauses.

smol-ninja: How do you decide where to keep the command after `and` and where to remove it? There are multiple instances through the repos where we have comma after `and` including License file.

Should we make a rule of thumb to not have comma after `and`?

PaulRBerg: My rules of thumb are:

- If there are two elements: no comma.
- If there are three or more elements: yes, comma
- If the clauses are independent: yes, comma
- If the clauses are NOT independent: no comma

PR #1141, opened 2025-01-10
test: move all `allowedToHook` calls in Integration contract

smol-ninja:
Closes https://github.com/sablier-labs/v2-core/issues/1129

### DIFF
--- a/tests/integration/Integration.t.sol
+++ b/tests/integration/Integration.t.sol
@@ -24,6 +24,8 @@ abstract contract Integration_Test is Base_Test {
     // Common stream IDs to be used across the tests.
     // Default stream ID.
     uint256 internal defaultStreamId;
+    // A stream with a recipient contract that is not allowed to hook.
+    uint256 internal notAllowedtoHookStreamId;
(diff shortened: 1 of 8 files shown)


andreivladbrg (approved):
👍

tests/integration/Integration.t.sol
```diff
@@ -24,6 +24,8 @@ abstract contract Integration_Test is Base_Test {
     // Common stream IDs to be used across the tests.
     // Default stream ID.
     uint256 internal defaultStreamId;
+    // A stream with a recipient contract that is not allowed to hook.
+    uint256 internal notAllowedtoHookStreamId;
```
PaulRBerg, line 28: It's quite difficult to see that this is a stream ID (due to the length).

Should we create a struct or a contract so that we can do `streamIds.notAllowedToHook`, `streamIds.default`, etc.?

If yes, let's create an issue, please.

cc @sablier-labs/solidity

smol-ninja: Good idea.

PR #1144, opened 2025-01-13
build: lower optimizer runs

andreivladbrg:


smol-ninja (approved):
I have pushed a new commit with the following changes:
- Changes runs to 570.
- Updates gas benchmark values, max segment values and the precompiles with the new runs.

If it looks good, feel free to merge it.

andreivladbrg:
> Changes runs to 570

I intentionally set it to a lower value to anticipate potential issues on certain chains, e.g. [Linea](https://github.com/sablier-labs/deployments/tree/main/lockup/v1.2.0/core#core-dao-and-linea). This way, we wouldn’t need to lower it later if problems arise—but I guess we eventually might.

> Updates gas benchmark values, max segment values and the precompiles with the new runs

thank you. surprized how mucht the optimizer runs affect the gas. i looks like a ~2-3k gas difference for the create functions

PR #1158, opened 2025-01-22
refactor: automate libraries linkage and remove precompiles

smol-ninja:
While working on docs for custom deployment docs, I found that its very complicated to make custom deployment if `foundry.toml` contains hard coded libraries' addresses. So I investigated and found a solution to remove them from the `foundry.toml` and automatically replace placeholders with deployed addresses of libraries.

This now looks neat to me, but may have an issue that I could not think of, so feedbacks are welcome.

Also, lets review and merge it before merging release PR.

--------------

Closes https://github.com/sablier-labs/v2-core/issues/1160.


smol-ninja:
I am converting it into draft as it requires more changes.

smol-ninja:
@andreivladbrg you can review it now.

tests/utils/DeployOptimized.sol
```diff
@@ -29,18 +32,38 @@ abstract contract DeployOptimized is StdCheats, CommonBase {
...
         // Deploy the libraries.
-        deployOptimizedLibraries();
+        (address helpers, address vestingMath) = deployOptimizedLibraries();
+
+        // Get the bytecode from {SablierLockup} artifact.
+        string memory artifactJson = vm.readFile("out-optimized/SablierLockup.sol/SablierLockup.json");
+        string memory rawBytecode = artifactJson.readString(".bytecode.object");
+
+        // Replace libraries placeholders.
+        rawBytecode = vm.replace({
```
andreivladbrg, line 45: can we add more explanatory comments about why we are doing these replacements please?

for a first time reader it would be very confusing

let's say smth like:

```
// The placeholder `__$<value>$__` is a unique marker generated during compilation to represent where the
// address of the library will be inserted.
// By replacing this placeholder, we "link" the library address into the contract's bytecode.
```

andreivladbrg: d23f6d2d96f1a2a6c8ab4149610ea2078c8fde6a

smol-ninja: Thanks looks good.

andreivladbrg:
gj, the with the `libraryPlaceholder` function it is better organized and aligns with how it actually works behind

PR #1162, opened 2025-01-28
Release v2.0.0

andreivladbrg:
Merge `main` --> `release`


smol-ninja:
Instead of merging, which would create a merge commit in `release` branch, shouldn't we just `git reset hard` release branch to reset it to main branch?

andreivladbrg:
> Instead of merging, which would create a merge commit in release branch, shouldn't we just git reset hard release branch to reset it to main branch?

we can "Rebase and merge" - same result

smol-ninja (approved):

PR #1165, opened 2025-02-01
test: use mainnet addresses in fork tests

smol-ninja:
### Changelog

- Remove `virtual` modifier from `run()` 
- Add `broadcast` to `_run` and remove it from `run` in DeployProtocol script
- Replace hard-coded `FORK_TOKEN_HOLDER` with locally created `forkTokenHolder` address
- Use mainnet addresses in fork tests
- Replace `defaults.START_TIME()` with `getBlockTimestamp() + 2 days` in fork test because `getBlockTimestamp()` returns timestamp of forked block


script/DeployBatchLockup.t.sol
andreivladbrg, line 1: just noticed that this file name is `.t.sol` instead of `.s.sol`

script/DeployDeterministicBatchLockup.t.sol
andreivladbrg, line 1: same here

tests/fork/Fork.t.sol
```diff
@@ -30,19 +33,24 @@ abstract contract Fork_Test is Base_Test {
...
+        forkTokenHolder = payable(makeAddr(string.concat(IERC20Metadata(address(FORK_TOKEN)).symbol(), "_HOLDER")));
 
-        // Label the contracts.
+        // Label the addresses.
         labelContracts();
 
-        // Make the forked token holder the caller in this test suite.
-        resetPrank({ msgSender: FORK_TOKEN_HOLDER });
+        // Deal token balance to the user.
+        initialHolderBalance = 10_000_000 * 10 ** IERC20Metadata(address(FORK_TOKEN)).decimals();
```
andreivladbrg: should we use the shorter version `1e7`?

andreivladbrg (approved):
pushed a commit 1b04b66ee9afcf0f08c9b175d995efba3a2e3bee
LGTM👍

tests/fork/Fork.t.sol
smol-ninja: Anything is fine by me. I like you changing it to `10e6` and we can even use `1e7`. The reason I didn't use `e` notation is because `e` is generally used to represent decimals.

andreivladbrg: > Anything is fine by me. I like you changing it to `10e6` and we can even use `1e7`. The reason I didn't use `e` notation is because `e` is generally used to represent decimals

fair enough, i don't have a strong preference, feel free to do as wish & merge

Issue #1157, opened 2025-01-22
Remove Broker functionality


andreivladbrg:
Ref https://github.com/sablier-labs/company-discussions/discussions/97#discussioncomment-11907866

PR #1166, opened 2025-02-04
refactor: remove broker

andreivladbrg:
Closes https://github.com/sablier-labs/lockup/issues/1157

I also decided to rename the `totalAmount` parameter to `depositAmount`, as there was no other amount in the `create` function. I’m also considering removing the `CreateEventCommon` struct, as I believe we will no longer encounter the "Stack too deep" error, the only problem is that it will cause more changes to the app. @smol-ninja wdyt?

I'm surprised by how much this change reduces the code! This would be advantageous for integrators as well.


PaulRBerg:
cc @sablier-labs/frontend

smol-ninja (approved):
Great work on the PR @andreivladbrg. I have pushed a small commit (https://github.com/sablier-labs/lockup/pull/1166/commits/d62a8d487093949673255131888d7f766ede13b5)- just some natspec fixes and moved `Base_Test.setUp()` under `TODO` tag to signal it clearly that it needs to be removed later.

If all looks good, you can go ahead and merge it.

Re `CreateEventCommon`, can we do it in another PR and as a separate issue for easier review?

I have also not reviewed `benchmarks` dir since it will be removed in the future (https://github.com/sablier-labs/lockup/issues/1170). I hope thats OK.

smol-ninja:
Re `CreateEventCommon`, can we do it in another PR and as a separate issue for easier review?

andreivladbrg:
> Re `CreateEventCommon`, can we do it in another PR and as a separate issue for easier review?

sounds good

> I have also not reviewed benchmarks dir since it will be removed in the future (https://github.com/sablier-labs/lockup/issues/1170). I hope thats OK.

that's ok, just keep in mind when you work on the new repo that there were made some changes in this branch

smol-ninja:
> that's ok, just keep in mind when you work on the new repo that there were made some changes in this branch

Yes we both will be co-authors of those changes. Also, in the future PR, in case it requires making changes to the benchmark code, then lets just remove it (https://github.com/sablier-labs/lockup/issues/1170).

> btw, pushed a commit to crease the coverage, don't know how we have missed it until now

Good catch.

PaulRBerg:
Agree that it was a good idea to rename `totalAmount` to `depositAmount`.

Issue #1168, opened 2025-02-04
Imports source contracts using relative path in `Modifiers` and `BatchLockupBuilder`


smol-ninja:
Both `BatchLockupBuilder` and `Modifiers` are not using relative path to import source contracts. As a a result, their imports cannot be resolved when importing Lockup using the npm package.

https://github.com/sablier-labs/lockup/blob/main/tests/utils/BatchLockupBuilder.sol#L4

https://github.com/sablier-labs/lockup/blob/main/tests/utils/Modifiers.sol#L4

PR #1169, opened 2025-02-05
tests: use relative path in utils contracts

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1168


andreivladbrg (commented):
didn't we decide to publish `2.0.1`? why no version bump in `package.json` file, and no `CHANGELOG` update?

also, i was looking at the `DeployOptimized` utils test contract, this file assumes there is a `out-optimized` profile under foundry configuration. which might be very unlikely, thus, should we mark it as `.t.sol` instead?

smol-ninja:
> didn't we decide to publish 2.0.1? why no version bump in package.json file, and no CHANGELOG update?

Will do. I missed it.

> DeployOptimized utils test contract, this file assumes there is a out-optimized profile under foundry configuration

Its not looking for the profile, but for the `out-optimized` directory, which we are already publishing under the `artifacts` name. Loading from `artifacts` will be complicated though so happy to rename it to `.t.sol`.

smol-ninja:
Instead of renaming it to `DeployOptimized.t.sol`, what if I add it to `.npmignore`? Base imports from `DeployOptimized.t.sol` and importing from a `.t.sol` file looks weird.

tests/utils/.npmignore
```diff
@@ -1 +1,2 @@
 *.t.sol
+DeployOptimized.sol
```
andreivladbrg, line 2: don't agree

smol-ninja: > why? :)) all the other files that are not under utils use this format, and they don’t look weird. (e.g. Base.t.sol)

Good point. Made that change.

smol-ninja:
made all the changes. LMK if it looks good now.

CHANGELOG.md
```diff
@@ -13,6 +14,17 @@ The format is based on [Common Changelog](https://common-changelog.org/).
 [1.0.1]: https://github.com/sablier-labs/lockup/compare/v1.0.0...v1.0.1
 [1.0.0]: https://github.com/sablier-labs/lockup/releases/tag/v1.0.0
 
+## [2.0.1] - 2025-02-05
+
+### Changed
+
+- Use relative path to import source contracts in `utils` files
```
andreivladbrg, line 21: one more thing, sorry

could we mention "test" `utils`?

andreivladbrg (approved):
👍

CHANGELOG.md
smol-ninja: Do you want it to be like

- Use relative path to import source contracts in "test" `utils` files 

or you mean 

- Use relative path to import source contracts in `tests/utils` files

andreivladbrg: first one: *Use relative path to import source contracts in test utils files*

smol-ninja: https://github.com/sablier-labs/lockup/pull/1169/commits/2ae61946d23cacea5c3593dc37f3f53c502e3c17

Issue #1134, opened 2025-01-06
Renamings for consistency


PaulRBerg:
It would be nice if we can get these refactors included in the current release (Lockup v2.0.0). I'll ask the auditors, but if they balk, I'd be OK with leaving this for v2.1.0

- [x] Rename `VestingMath` to `StreamingMath`
- [x] Shorten the model names to `LL`, `LT`, and `LD`
  - [x] In this function series: from `calculateLockupLinearStreamedAmount` to `calculateStreamedAmountLL`
  - [x] And from `checkCreateLockupLinear ` to `checkCreateLL`

There are two issues with the current names:

1. We maintain a duplicity of naming modalities for the models: (i) the longer form ("Lockup Linear"), and (ii) the short form ("LL")
2. There's an inconsistent ordering: some functions contain the model in the middle, while others contain it at as a suffix

smol-ninja:
Since you have assigned it to yourself @PaulRBerg, should I add it to the next Sprint or are you interested into doing it before the release?

smol-ninja:
Other than the pointers in the description, to confirm:

1. Do you want to refer to "Lockup Linear" as "LL" _everywhere_, or using "Lockup Linear" is fine in natspecs and comments? For example, comments like `This is used in Lockup Linear models` would become `This is used in LL models`. 
 
2. I talked to @razgraf and he recommended to keep the event names (`emit CreateLockupLinearStream`) as they are to avoid headaches for integrators. And to quote him, to which I also agree,

> I agree the reasoning behind them is sound with regards to naming schemas and cleaner architecture. But sometimes, they simply don't make sense when ran through a cost analysis - they give more headaches than they take aways.

3. We have `LockupLinear` library in DataTypes, are we OK with that?

4. There are contract names such as "LockupLinear.t.sol", we OK with that?

PaulRBerg:
1. Yes.

2. Makes sense to me

3. Yes

4. Yes - contract, library, and file names are good

PR #1177, opened 2025-02-13
docs: renaming of functions and variables

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1134

- [x] Rename `VestingMath` to `StreamingMath`
- [x] In this function series: from `calculateLockupLinearStreamedAmount` to `calculateStreamedAmountLL`
- [x] And from `checkCreateLockupLinear ` to `checkCreateLL`

### DIFF
--- a/script/Init.s.sol
+++ b/script/Init.s.sol
@@ -30,7 +30,7 @@ contract Init is MaxCountScript {
         // Approve the Lockup contracts to transfer the ERC-20 tokens from the sender.
         token.approve({ spender: address(lockup), value: type(uint256).max });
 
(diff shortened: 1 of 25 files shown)


PaulRBerg:
Had a brief look — how about renaming the folders from `lockup-linear` to `lockup/linear`? So using a nested hierarchy.

This would better match the unified `Lockup` architecture.

_Side note: low-preference, will let you decide_

andreivladbrg (approved):
shogun_lgtm

smol-ninja:
I will merge it now and we can discuss restructuring tests in a separate issue.

PR #1182, opened 2025-02-18
feat: erc20 recovery

andreivladbrg:


smol-ninja (approved):
Look good to me. Added a small commit, feel free to merge the PR after you have reviewed it.

I would like to float an idea (for future), we write assertions as the following:

```solidity
// It should not burn the NFT.
address actualNFTOwner = lockup.ownerOf({ tokenId: ids.recipientGoodStream });
address expectedNFTOwner = address(recipientGood);
assertEq(actualNFTOwner, expectedNFTOwner, "NFT owner");
```

When the right hand side expression does not include an arithmetic operation, imo, we should avoid assigning it to a new variable. For example, the following maintains the same readability and at the same time, makes it more convenient for readers to read the whole code with minimum scrolling.

```solidity
// It should not burn the NFT.
assertEq(lockup.ownerOf({ tokenId: ids.recipientGoodStream }), address(recipientGood), "NFT owner");
```

Only when right hand side includes some computation or arithmetic, we should consider assigning those value to a new variable.

andreivladbrg:
> Added a small commit, feel free to merge the PR after you have reviewed it.

looks good, thanks.

> Only when right hand side includes some computation or arithmetic, we should consider assigning those value to a new variable.

i agree with this. 
there is one more thing that we might consider, formatting? if its 1 line vs 3 lines

smol-ninja:
> there is one more thing that we might consider, formatting? if its 1 line vs 3 lines

Yes agree.

PR #1185, opened 2025-02-20
avb: tests dry fy

andreivladbrg:
this PR addresses my feedback on #1179 and includes some additional changes.

### DIFF
--- a/tests/fork/Fork.t.sol
+++ b/tests/fork/Fork.t.sol
@@ -49,10 +49,10 @@ abstract contract Fork_Test is Base_Test {
         initialHolderBalance = uint128(1e6 * (10 ** IERC20Metadata(address(FORK_TOKEN)).decimals()));
         deal({ token: address(FORK_TOKEN), to: forkTokenHolder, give: initialHolderBalance });
 
+        resetPrank({ msgSender: forkTokenHolder });
+
         // Approve {SablierLockup} to transfer the holder's tokens.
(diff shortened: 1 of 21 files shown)


tests/fork/Fork.t.sol
```diff
@@ -46,15 +46,13 @@ abstract contract Fork_Test is Base_Test {
...
 
-        resetPrank({ msgSender: forkTokenHolder });
-
         // Approve {SablierLockup} to transfer the holder's tokens.
-        // We use a low-level call to ignore reverts because the token can have the missing return value bug.
-        (bool success,) = address(FORK_TOKEN).call(abi.encodeCall(IERC20.approve, (address(lockup), MAX_UINT256)));
-        success;
+        approveContract({ token_: address(FORK_TOKEN), from: forkTokenHolder, spender: address(lockup) });
+
+        resetPrank({ msgSender: forkTokenHolder });
```
smol-ninja, line 55: Lets move it above `approveContract` because I assume `approveContract` is changing prank internally but its not intuitive by looking at it.

smol-ninja: and because the obvious approach is to reset prank first and then approve contract.

andreivladbrg: it has `from` as param

we could technically completely remove it, but i agree it improves readability if we keep it.
if it before or after, i really don't have a preference

smol-ninja: OK then lets move it above :))

smol-ninja (approved):
Thanks for the review.

PR #1189, opened 2025-02-24
correction SablierBatchLockup.sol

sky-coderay:
of all of stream 
of all the stream

of - deleted

### DIFF
--- a/script/Init.s.sol
+++ b/script/Init.s.sol
@@ -7,7 +7,7 @@ import { ud60x18 } from "@prb/math/src/UD60x18.sol";
 import { Solarray } from "solarray/src/Solarray.sol";
 
 import { ISablierLockup } from "../src/interfaces/ISablierLockup.sol";
-import { Broker, Lockup, LockupDynamic, LockupLinear, LockupTranched } from "../src/types/DataTypes.sol";
+import { Broker, Lockup, LockupDynamic, LockupLinear } from "../src/types/DataTypes.sol";
 
 import { BaseScript } from "./Base.s.sol";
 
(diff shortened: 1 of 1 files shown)


src/SablierBatchLockup.sol
```diff
@@ -52,7 +52,7 @@ contract SablierBatchLockup is ISablierBatchLockup {
             revert Errors.SablierBatchLockup_BatchSizeZero();
         }
 
-        // Calculate the sum of all of stream amounts. It is safe to use unchecked addition because one of the create
+        // Calculate the sum of all the stream amounts. It is safe to use unchecked addition because one of the create
```
PaulRBerg, line 55: We can just remove `of`. No need to add `the`.

src/SablierBatchLockup.sol
```diff
@@ -101,7 +101,7 @@ contract SablierBatchLockup is ISablierBatchLockup {
...
```
PaulRBerg, line 104: Ditto here

src/SablierBatchLockup.sol
```diff
@@ -162,7 +162,7 @@ contract SablierBatchLockup is ISablierBatchLockup {
...
```
PaulRBerg, line 165: Ditto here

src/SablierBatchLockup.sol
```diff
@@ -212,7 +212,7 @@ contract SablierBatchLockup is ISablierBatchLockup {
...
```
PaulRBerg, line 215: Ditto here

src/SablierBatchLockup.sol
```diff
@@ -267,7 +267,7 @@ contract SablierBatchLockup is ISablierBatchLockup {
...
```
PaulRBerg, line 270: Ditto here

PaulRBerg (changes requested):

PaulRBerg:
Thanks @sky-coderay, can you now please rebase from `staging` so we have a clean commit history.

You can just start afresh and add your changes on top of `staging`, and then run `git push --force` to your current branch.

PR #1209, opened 2025-03-14
Audit fixes

### LINKED ISSUE #1208: Audit fixes
## Non-informational fixes:

- For tokens like `POL` and `CELO`, we won't support them at the UI level.

### Informational fixes:

- Remove the unused `Errors.LockupNFTDescriptor_UnknownNFT`.
- Add more documentation about "double-entry" tokens, and update Assumptions to state that these tokens are not supported.

---

*(For further explanations, go through each finding.)*

**Findings:**
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/7  
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/8  
- https://cantina.xyz/code/c9980035-706a-4119-91fe-7ebddaf123e4/findings/11

andreivladbrg:
Closes #1208 and #1220


smol-ninja:
Before the review, I would like to wait for a comment on https://github.com/sablier-labs/company-discussions/discussions/115. 

Screenshot 2025-03-14 at 12 21 06

As I think we might make a decision restrict `collectFees` to only be callable by admin.

SECURITY.md
```diff
@@ -19,6 +19,7 @@ vulnerability, it must adhere to these assumptions as well:
   an approved address. This excludes rebase tokens, interest-bearing tokens, and permissioned tokens where the admin can
   arbitrarily change balances.
 - The token contract does not allow callbacks (e.g. ERC-777 is not supported).
+- The token contract must not implement multiple entry points for transfers (i.e., no double entry-point mechanisms).
```
smol-ninja, line 22: I simplified it as `- The token contract has only one entry point.` in airdrops.

smol-ninja:
Reminder to check contract size after implementing `setNativeToken` function.

SECURITY.md
andreivladbrg: will updated to the version from this commit https://github.com/sablier-labs/airdrops/commit/c4b155b0a0e3daa93455b95e8c82562ca576f166

smol-ninja:
> That’s very unfortunate—I had to lower the optimizer_runs to 50, since the contract sizes at this setting are

Yeah thats very unfortunate. Lets brainstorm Lockup design in the coming months to find solutions for this problem.

src/abstracts/SablierLockupBase.sol
```diff
@@ -57,14 +60,12 @@ abstract contract SablierLockupBase is
     }
 
     /*//////////////////////////////////////////////////////////////////////////
-                                      MODIFIERS
+                                     MODIFIERS
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Checks that `streamId` does not reference a null stream.
```
smol-ninja, line 66: @andreivladbrg we do use variable names with backtick in comments. I think both are fine, depends on the situation.

src/interfaces/ISablierLockup.sol
```diff
@@ -163,6 +163,7 @@ interface ISablierLockup is ISablierLockupBase {
     /// - `params.recipient` must not be the zero address.
     /// - `params.sender` must not be the zero address.
     /// - `msg.sender` must have allowed this contract to spend at least `params.depositAmount` tokens.
+    /// - `params.token` must not be the native token.
```
smol-ninja, line 166: variable name with `backtick`, again, fine from my side :)) just pointing out.

tests/integration/concrete/lockup-base/create-with-timestamps/createWithTimestamps.tree
```diff
@@ -17,5 +17,8 @@ CreateWithTimestamps_Integration_Concrete_Test
                ├── when start time zero
                │  └── it should revert
                └── when start time not zero
-                  └── when token not contract
-                     └── it should revert
+                  ├── when token native token
+                  │  └── it should revert
+                  └── when token not native token
+                     └── when token not contract
```
smol-ninja, line 23: The order should be:

1. when token not contract -> it should revert
2. when token contract
3. when token native token -> it should revert
4. when token not native token

foundry.toml
```diff
@@ -9,7 +9,7 @@
   ]
   gas_limit = 9223372036854775807
   optimizer = true
-  optimizer_runs = 500
+  optimizer_runs = 50
```
smol-ninja, line 12: 500 works too.

smol-ninja (changes requested):

foundry.toml
smol-ninja: I tested with 500 and the margin is 361. So we are in good hands.

andreivladbrg: as discussed on slack, i blundered here :))

src/abstracts/SablierLockupBase.sol
andreivladbrg: > depends on the situation.

Yeah, agree — natspec before the function logic starts vs inside the function logic. Also, like mentioned in airdrops, it was something Paul suggested. I want to stick to his advice and avoid future changes in his PRs, I guess :)) Personally, I don’t have a strong opinion on it.

tests/integration/concrete/lockup-base/create-with-timestamps/createWithTimestamps.tree
smol-ninja: For token to compare with a native token, it must be a contract address first. So the order is more helpful when the test checks for whether token is a contract or not and if its a contract, it checks if its a native token or not.

I understand that you followed the implementation execution here but I think sometimes we can divert away from how the implementation is in the source code and how the test flow should be.

That was my point. Either way it works.

smol-ninja (approved):

andreivladbrg:
@smol-ninja ik you've already approved, but i pushed a new commit a1e3a3e40b702a1880a1c8f511ecad2ad57032cf to aligned with the latest airdrops changes

smol-ninja:
yeah looks good.

PR #1210, opened 2025-03-14
docs: improve writing in NatSpec

PaulRBerg:
This PR contains mostly NatSpec changes.

I know we discussed we will have one mega PR for all refactors, but this PR ain't that. This is just for writing improvements.


src/interfaces/ISablierLockup.sol
```diff
@@ -49,24 +49,24 @@ interface ISablierLockup is ISablierLockupBase {
...
 
     /// @notice Retrieves the stream's cliff time, which is a Unix timestamp.  A value of zero means there is no cliff.
-    /// @dev Reverts if `streamId` references a null stream or a stream created using LD and LT model.
+    /// @dev Reverts if `streamId` references either a null stream or a non-LL stream.
     /// @param streamId The stream ID for the query.
     function getCliffTime(uint256 streamId) external view returns (uint40 cliffTime);
 
     /// @notice Retrieves the segments used to compose the dynamic distribution function.
-    /// @dev Reverts if `streamId` references a null stream or a stream created without using LD model.
+    /// @dev Reverts if `streamId` references either a null stream or a LD stream.
```
smol-ninja:
```suggestion
    /// @dev Reverts if `streamId` references either a null stream or a non-LD stream.
```

src/interfaces/ISablierLockup.sol
```diff
@@ -49,24 +49,24 @@ interface ISablierLockup is ISablierLockupBase {
...
     /// @param streamId The stream ID for the query.
     /// @return segments See the documentation in {DataTypes}.
     function getSegments(uint256 streamId) external view returns (LockupDynamic.Segment[] memory segments);
 
     /// @notice Retrieves the tranches used to compose the tranched distribution function.
-    /// @dev Reverts if `streamId` references a null stream or a stream created without using LT model.
+    /// @dev Reverts if `streamId` references either a null stream or a LT stream.
```
smol-ninja:
```suggestion
    /// @dev Reverts if `streamId` references either a null stream or a non-LT stream.
```

src/interfaces/ISablierLockup.sol
```diff
@@ -49,24 +49,24 @@ interface ISablierLockup is ISablierLockupBase {
...
     /// @param streamId The stream ID for the query.
     /// @return tranches See the documentation in {DataTypes}.
     function getTranches(uint256 streamId) external view returns (LockupTranched.Tranche[] memory tranches);
 
     /// @notice Retrieves the unlock amounts used to compose the linear distribution function.
-    /// @dev Reverts if `streamId` references a null stream or a stream created without using LL model.
+    /// @dev Reverts if `streamId` references either a null stream or a LL model.
```
smol-ninja:
```suggestion
    /// @dev Reverts if `streamId` references either a null stream or a non-LL stream.
```

tests/integration/fuzz/lockup-base/withdraw.t.sol
```diff
@@ -222,7 +222,7 @@ abstract contract Withdraw_Integration_Fuzz_Test is Integration_Test {
         address expectedNFTOwner = users.recipient;
         assertEq(actualNFTowner, expectedNFTOwner, "NFT owner");
 
-        // It should update the aggrate balance.
+        // It should update the aggregate balance.
```
smol-ninja, line 225: Since its fuzz test, shouldn't we use `Asset that` approach here similar to your refactor in cancel test?

I like the "asset that" approach in fuzz tests and "it should" approach in concrete tests.

tests/invariant/Invariant.t.sol
```diff
@@ -71,16 +71,32 @@ contract Invariant_Test is Base_Test, StdInvariant {
...
         assertGe(
             contractBalance,
-            depositedAmountsSum - refundedAmountsSum - withdrawnAmountsSum,
-            unicode"Invariant violation: contract balances < Σ deposited amounts - Σ refunded amounts - Σ withdrawn amounts"
+            aggregateBalance,
+            unicode"Invariant violation: contract balances < Σ deposits - Σ refunds - Σ withdrawals"
+        );
+    }
+
+    function invariant_DepositedAmountsSumGteAggregateBalance() external view {
```
smol-ninja, line 88: Why do we need this? We already have the following invariant tested above:

`Invariant violation: aggregate balance != Σ deposits - Σ refunds - Σ withdrawals`

This implies `aggregate balance <= Σ deposits`.

src/interfaces/ISablierLockup.sol
```diff
@@ -11,15 +11,15 @@ interface ISablierLockup is ISablierLockupBase {
                                        EVENTS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @notice Emitted when a stream is created using LD model.
+    /// @notice Emitted when a LD stream is created.
```
smol-ninja:
```suggestion 
    /// @notice Emitted when an LD stream is created.
```

Same with LL and LT below.

src/libraries/Helpers.sol
```diff
@@ -136,7 +136,7 @@ library Helpers {
                             PRIVATE CONSTANT FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @dev Checks the user-provided cliff, end times and unlock amounts of a stream with LL model.
+    /// @dev Checks the user-provided cliff, end times, and unlock amounts of a LL stream.
```
smol-ninja:
```suggestion
    /// @dev Checks the user-provided cliff, end times, and unlock amounts of an LL stream.
```

smol-ninja (changes requested):

src/interfaces/ISablierLockup.sol
PaulRBerg: Good catch

tests/integration/fuzz/lockup-base/withdraw.t.sol
PaulRBerg: It's up to you.

tests/invariant/Invariant.t.sol
PaulRBerg: Good point. We can remove it (feel free to do it yourself).

src/interfaces/ISablierLockup.sol
PaulRBerg: I've thought about this when I authored the commit, and I decided to stick with `a LD` for the following reasons:

1. Easier to make refactors
2. The text can be read as "A Lockup Dynamic" instead of the acronyms

tests/integration/fuzz/lockup-base/withdraw.t.sol
smol-ninja: I will leave it unchanged then.

PR #1233, opened 2025-03-31
refactor: move shell scripts under scripts/bash

andreivladbrg:
As discussed here: https://github.com/sablier-labs/airdrops/issues/121


smol-ninja (approved):
The diff looks good. I would assume that you have tested the changes yourself.

Issue #1237, opened 2025-04-09
Delete remappings.txt


smol-ninja:
As discussed in https://github.com/sablier-labs/flow/discussions/430, since foundry can handle autoremappings, delete `remappings.txt` file.

PR #1238, opened 2025-04-10
Remove remappings

andreivladbrg:
Closes https://github.com/sablier-labs/lockup/issues/1237 

**Note:** I didn't remove the section about `remappings` in the case of `forge install`, as it would still be required when using git submodules.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ typechain-types
 *.log
 .DS_Store
 .pnp.*
+bun.lockb
 deployments.md
 lcov.info
 package-lock.json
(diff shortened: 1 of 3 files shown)


smol-ninja:
>  it would still be required when using git submodules

Its not required for mapping only the dependencies in `lib` folder. It is required to map submodules dependencies. So, I would suggest to either remove it from the `README.md` to update it to match the actual remappings which is a bit complicated than what we had mentioned. You can follow the readme guide and try it yourself.

andreivladbrg:
i am ok removing the section from `README`

smol-ninja (approved):

Issue #1234, opened 2025-04-01
Add a "to" parameter in the "collectFees" function


PaulRBerg:
## Problem

As explained by Max [here](https://github.com/sablier-labs/command-center/discussions/119), there's a need for automating the fee management _after_ the fees are collected.

As of now, automating this process is difficult because the fees can only be collected to the admin contract, which is a multisig on some chains. On these chains, we will need to manually submit a multisig tx for bridging or selling the fees for USDC.

## Solution

Add a new `to` parameter in the `collectFees` function, which would behave similarly to the `to` parameter in the stream withdraw functions:

- If the `msg.sender` is the admin, allow `to` to be any address.
- Otherwise, `to` must be the admin. If `to` is not the admin, the function should revert.

smol-ninja:
Given that we are going to use the collect fee script with a separate EOA, the fee will always be collected to the admin address. So in practice, I think the problem will not be solved just by this.

What we need is a new role to authorize collectFee(address to) on behalf of the admin. Thats the only possible solution. Though admin would still have to authorize the EOA once per contract per chain.

Thoughts?

smol-ninja:
Nvm I just saw the other discussion.

Issue #1231, opened 2025-03-31
Remove `indexed` from `feeAmount` parameter in `CollectFees` event


smol-ninja:
[ISablierLockupBase.sol#L51](https://github.com/sablier-labs/lockup/blob/staging/src/interfaces/ISablierLockupBase.sol#L51): `CollectFees` event should not index `feeAmount`. 

Change

```solidity
 event CollectFees(address indexed admin, uint256 indexed feeAmount);
```

to

```solidity
 event CollectFees(address indexed admin, uint256 feeAmount);
```

PaulRBerg:
Why??

andreivladbrg:
Why is that? Whats the problem with fee amount?

smol-ninja:
The indexed parameters are used as filters to search for these amounts. Since `feeAmount` would vary all the time in `CollectFees` event, why should it be indexed?

Do you think we will be searching event logs using fee amount value, such as get me all the events with 0.001 ETH? I don't think so but let me know otherwise.

smol-ninja:
I think you guys assumed that I am talking about removing `feeAmount` from `CollectFees`. No, I am talking about not "indexing" it.

So, the suggestion is to have

```solidity
 event CollectFees(address indexed admin, uint256 feeAmount);
```

instead of

```solidity
 event CollectFees(address indexed admin, uint256 indexed feeAmount);
```

andreivladbrg:
> I think you guys assumed that I am talking about removing feeAmount from CollectFees

Oh, yeah, thats what I thought :))

Then, yes, should be fine to remove it

smol-ninja:
Thanks. Updated description for expciciteness.

PaulRBerg:
Hahhaa yes I have also thought you wanted to remove it

PR #1239, opened 2025-04-11
feat: collect fee with fee recipient

smol-ninja:
This PR is required prior to integrating RBAC and is same as https://github.com/sablier-labs/flow/pull/435. After RBAC is integrated, I will move the common code into the `evm-utils`.

### Changelog
- Closes https://github.com/sablier-labs/lockup/issues/1234
- Closes https://github.com/sablier-labs/lockup/issues/1231


andreivladbrg (approved):
LGTM

Issue #1240, opened 2025-04-11
Role-Based Access Control


smol-ninja:
Similar to https://github.com/sablier-labs/airdrops/issues/141

PR #1242, opened 2025-04-14
feat: access based control over collect fee function

smol-ninja:
### Changelog

- Closes https://github.com/sablier-labs/lockup/issues/1240
- Prefix events with associated interface contract

### Important note
- Depends on [this discussion](https://github.com/sablier-labs/command-center/discussions/135#discussioncomment-12822384)

### Blocked by
- https://github.com/sablier-labs/evm-utils/pull/20

### DIFF
--- a/src/abstracts/SablierLockupBase.sol
+++ b/src/abstracts/SablierLockupBase.sol
@@ -1,16 +1,16 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.8.22;
 
+import { AccessControl } from "@openzeppelin/contracts/access/AccessControl.sol";
 import { IERC4906 } from "@openzeppelin/contracts/interfaces/IERC4906.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { ERC721 } from "@openzeppelin/contracts/token/ERC721/ERC721.sol";
-import { AccessControl } from "@openzeppelin/contracts/access/AccessControl.sol";
 import { IERC721Metadata } from "@openzeppelin/contracts/token/ERC721/extensions/IERC721Metadata.sol";
 import { IERC165 } from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
 import { Batch } from "@sablier/evm-utils/src/Batch.sol";
-import { RoleAdminable } from "@sablier/evm-utils/src/RoleAdminable.sol";
 import { NoDelegateCall } from "@sablier/evm-utils/src/NoDelegateCall.sol";
+import { RoleAdminable } from "@sablier/evm-utils/src/RoleAdminable.sol";
 
 import { ILockupNFTDescriptor } from "./../interfaces/ILockupNFTDescriptor.sol";
 import { ISablierLockupBase } from "./../interfaces/ISablierLockupBase.sol";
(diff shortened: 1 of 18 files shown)


src/abstracts/SablierLockupBase.sol
```diff
@@ -20,10 +21,10 @@ import { Lockup } from "./../types/DataTypes.sol";
 /// @notice See the documentation in {ISablierLockupBase}.
 abstract contract SablierLockupBase is
     Batch, // 1 inherited components
+    ERC721, // 6 inherited components
+    ISablierLockupBase, // 7 inherited components
     NoDelegateCall, // 0 inherited components
-    Adminable, // 1 inherited components
-    ISablierLockupBase, // 6 inherited components
-    ERC721 // 6 inherited components
+    RoleAdminable // 6 inherited components
```
andreivladbrg, line 27: i think its 3 inherited components for `RoleAdminable`

src/abstracts/SablierLockupBase.sol
```diff
@@ -349,9 +350,11 @@ abstract contract SablierLockupBase is
 
     /// @inheritdoc ISablierLockupBase
     function collectFees(address feeRecipient) external override {
-        // Check: if `msg.sender` is not the admin, `feeRecipient` must be the admin address.
-        if (msg.sender != admin && feeRecipient != admin) {
-            revert Errors.SablierLockupBase_FeeRecipientNotAdmin(feeRecipient, admin);
+        // Check: if `msg.sender` has neither the {RoleAdminable.FEE_COLLECTOR_ROLE} role nor is the contract admin,
+        // then `feeRecipient` must be the admin address.
+        bool hasRoleOrIsAdmin = hasRoleOrIsAdmin({ role: FEE_COLLECTOR_ROLE, account: msg.sender });
```
andreivladbrg, line 355: let's use the internal `_hasRoleOrIsAdmin` as its slight more efficient

andreivladbrg: ohh - its private - does making it `internal` would increase the gas cost?

tests/Base.t.sol
```diff
@@ -104,6 +96,26 @@ abstract contract Base_Test is Assertions, Calculations, DeployOptimized, Modifi
...
+        users.accountant = createUser("Accountant", spenders);
+        users.alice = createUser("Alice", spenders);
+        users.eve = createUser("Eve", spenders);
+        users.operator = createUser("Operator", spenders);
+        users.recipient = createUser("Recipient", spenders);
+        users.sender = createUser("Sender", spenders);
+
+        // Assign fee collector role to the accountant user.
+        setMsgSender(users.admin);
+        lockup.grantRole(FEE_COLLECTOR_ROLE, users.accountant);
```
andreivladbrg, line 116: since the function is called `createTestUsers` - i believe moving these lines in the `setUp` function would be more appropriate

andreivladbrg (commented):
generally looks good, left a few comments

src/abstracts/SablierLockupBase.sol
smol-ninja: Yeah.

src/abstracts/SablierLockupBase.sol
smol-ninja: No it wont increase the contract size but reduce the gas cost so thats a good idea.

smol-ninja: https://github.com/sablier-labs/evm-utils/commit/a03370133f3ca44cfaf3f7d8c4953f440b03b890

tests/Base.t.sol
smol-ninja: IMO its more appropriate to have it inside `createTestUsers`. Assigning correct roles and approving contracts (which happens inside `createUser` function) should be seen as a part of the create test user process.

andreivladbrg: good point about approvals - lets leave it like this

andreivladbrg (approved):
👍

Issue #1241, opened 2025-04-14
Remove `code.length` check from `allowToHook`


smol-ninja:
`recipient.code.length == 0` check is redundant because it is followed by `ISablierLockupRecipient(recipient).supportsInterface(interfaceId)`. `code.length` is implicitly checked by checking that `recipient` implements `supportsInterface`. 

cc @sablier-labs/evm for objection.

andreivladbrg:
nice catch - if the recipient is not a contract the call would fail 👍

PaulRBerg:
LGTM

IIRC that was check was added for better dev UX, not because it was necessary, but I agree that we should remove it to lower the contract size

smol-ninja:
Thanks for the confirmation. 

> for better dev UX

We can bother less about DX when caller is admin.

Issue #1221, opened 2025-03-17
Get rid of "MAX_COUNT"


PaulRBerg:
## Task

- [x] Delete `MAX_COUNT` and the associated checks in the create function
- [ ] Delete the `EstimateMaxCount` script
- [x] Delete all other logic related to the `MAX_COUNT`
- [ ] Write invariant tests to show that the gas consumption for `create` is always greater than `cancel` and/or `withdraw` (for the same `streamId`)

## Rationale + Benefits

- The estimation script takes into account the block gas limit at the time when the contract is deployed .. but the block gas limit changes all the time due to ERC-1559, and some chains like Base change their gas limit for [business reasons](https://github.com/sablier-labs/interfaces/issues/1473#issue-2898127008).
- It follows that:
     - There is no guarantee that the `MAX_COUNT` will remain accurate over time.
     - This is an artificial limitation that harms business for Sablier (by arbitrarily limiting how many segments/tranches there can be a in stream, when in fact there may be more).
- The UI will switch to an [optimistic approach](https://github.com/sablier-labs/interfaces/issues/1473#issuecomment-2729865961) that simulates the stream creation before submitting the tx. The same approach will enable us to check if the user-provided segment count

> [!NOTE]
> It's safe to do this because the gas cost of storing segments outweighs the gas cost of traversing them. So, we know the the gas cost of the `create` function is always larger than the gas cost of the `withdraw` or `cancel` functions when the number of segments/tranches is large.

## Side Note

- The estimation script remain still useful in the frontend, so let's move it to the [scripts](github.com/sablier-labs/scripts) repo, or re-implement it in the frontend code.

## Related

cc @sablier-labs/evm and @sablier-labs/frontend

Also see https://github.com/sablier-labs/interfaces/issues/1473#issuecomment-2718466401

smol-ninja:
Agree with everything you said. 

1. Re. the scripts, I think it could be useful in the future. So we should move it into `scripts` repo.
2. Re. invariant tests, wouldn't it be unnecessary? Its very clear why `create` is more expensive than `withdraw` or `cancel` which is due to the amount of `SSTORE` that it uses.

PaulRBerg:
1. Agree
2. It might be very clear but this is the purpose of tests. To criticize guesses like this. Also, tests also ensure regression safety, i.e., as the `cancel` and `withdraw` functions continue to evolve, the test will continuously show that the gas usage stays below `create`.

[...]

PR #1243, opened 2025-04-14
refactor: remove `MAX_COUNT`

smol-ninja:
### Changelog
- Closes https://github.com/sablier-labs/lockup/issues/1221
- `EstimateMaxCount` has been kept.
- Closes https://github.com/sablier-labs/lockup/issues/1241
- `test_RevertWhen_ProvidedAddressNotContract` test has been kept.

### Note

The invariant test will be handled separately via https://github.com/sablier-labs/lockup/issues/1246.

### Blocked by
- https://github.com/sablier-labs/lockup/pull/1242
- https://github.com/sablier-labs/evm-utils/pull/20


SECURITY.md
```diff
@@ -10,7 +10,7 @@ continuous process. If you believe you have found a security vulnerability, plea
 Sablier Lockup has been developed with a number of technical assumptions in mind. For a disclosure to qualify as a
 vulnerability, it must adhere to these assumptions as well:
 
-- The immutable `MAX_COUNT` has value that cannot lead to an overflow of the block gas limit.
+- The number of segments/tranches should be such that `create` cannot lead to an overflow of the block gas limit.
```
andreivladbrg, line 13: such that _creating a stream_
wdyt?

src/interfaces/ISablierLockup.sol
```diff
@@ -156,7 +152,7 @@ interface ISablierLockup is ISablierLockupBase {
     /// - Must not be delegate called.
     /// - `params.depositAmount` must be greater than zero.
     /// - `params.timestamps.start` must be greater than zero and less than the first segment's timestamp.
-    /// - `segments` must have at least one segment, but not more than `MAX_COUNT`.
+    /// - `segments` must have at least one segment.
```
andreivladbrg, line 155: should we mention that its size should not be too big?  
the same question applies to tranches.

tests/utils/Defaults.sol
```diff
@@ -25,9 +25,8 @@ contract Defaults {
     uint128 public constant DEPOSIT_AMOUNT_6D = 10_000e6;
     uint40 public immutable END_TIME;
     uint40 public constant FEB_1_2025 = 1_738_368_000;
-    uint256 public constant MAX_COUNT = 10_000;
+    uint256 public constant MAX_COUNT = 100;
```
andreivladbrg, line 28: we remove it

tests/utils/EstimateMaxCount.sol
andreivladbrg, line 1: do we still need it?

andreivladbrg (commented):
generally looks good

tests/utils/EstimateMaxCount.sol
andreivladbrg: oor, i think its still relevant for the `docs`, right?

tests/utils/Defaults.sol
andreivladbrg: 34442ea24f2fab8ae511d35ef70ccfce2941725a

src/interfaces/ISablierLockup.sol
smol-ninja: Its already added to to the Security. So, I don't think its necessary to mention here. Also, note that "too big" is very vague since it can vary from chain to chain. And this applies everywhere for example, in Safe's batch contract, batch size should never be too large etc.

tests/utils/EstimateMaxCount.sol
smol-ninja: Not sure if we still need it or not but it would be useful in case we decide to include a table in the docs about allowed number of segments on each network. So, lets wait for a few months and if we don't use it, we will remove it.

src/interfaces/ISablierLockup.sol
andreivladbrg: > this applies everywhere for example, in Safe's batch

good point

tests/utils/EstimateMaxCount.sol
andreivladbrg: agree

andreivladbrg (approved):
👍

PR #1244, opened 2025-04-15
perf: refactor for code size optimization

smol-ninja:
This PR focuses on minimising the contract size. So far, a reduction of 991 Bytes (using OZ's code) has been achieved and further reductions may be possible upon a thorough review of all events.

If there are any parameter in events (including the entire event itself) that will hardly be used, we should remove it. Events contribute non-trivial to the contract size.

## Changelog

1. (**-305 B**) Removes `renounceMultiple`

As confirmed with @razgraf, this function isn't integrated in the UI because it has never been requested by a user. If a user ever requests it, the UI can implement it through `batch`.

2. ~(**-205 B**) Moves the entire `calculateStreamedAmount` logic into `LockupMath`~ See [comment](https://github.com/sablier-labs/lockup/pull/1244#discussion_r2058515393)

3. (**-64 B**) Removes unnecessary checks from admin functions such as `surplus == 0` and `newNativeToken == address(0)`

4. (**-84 B**) Removes `isStream` from storage and replace it with `_streams[streamId].token != address(0)` check

5. (**-105 B**) Merges internal `_calculateStreamedAmount` into `_streamedAmountOf` as having two functions are redundant

6. (**-14 B**) Replaced `_isCallerStreamSender` with `msg.sender != _streams[streamId].sender`

7. (**-10 B**) Moves `_renounce` logic into `renounce`

7. (**-68 B**) Uses `_isAuthorized` inside `_isCallerStreamRecipientOrApproved`

8. (**-49 B**) Allows `_createLL` to set 0 values. The gas cost increases only by 200-300 which is very low but it impacts contract size.

9. Removes events that will be used very less frequently
  - Removes (**-47 B**) `SetNativeToken`, (**-90 B**) `Recover`, ~(**-59 B**) `AllowToHook` and (**-72 B**) `SetNFTDescriptor`~ events
  - (**-48 B**) Removes `funder` parameter from create events
  - (**-39 B**) Index `streamId` in `InvalidStreamInCancelMultiple` and `InvalidWithdrawalInWithdrawMultiple` events as it reduces contract size. Also index parameters in `AllowToHook` and `SetNFTDescriptor` events

## Blocked by
- (**-188 B**) https://github.com/sablier-labs/lockup/pull/1243
- (**+1598 B**) https://github.com/sablier-labs/lockup/pull/1242
- (**-716 B**) https://github.com/sablier-labs/evm-utils/pull/20


smol-ninja:
According to so-called [deep research with ChatGPT](https://chatgpt.com/c/680121cc-cbd4-8000-b6f9-7e36084ba3a8) (which I also verified onchain), null address is only used as a symbolic address to burn tokens on all the chains. It couldn't even find a null address that has deployed bytecode on it. Therefore, some smart contracts use null address to refer to native token (which is just a placeholder) [such as this](https://docs.unizen.io/api-get-started/utxo-assets-and-cosmos-swap/get-trade-v1-chainid-quote-cross-1). So, if a chain deploys a token on the null address, it would break several protocols and the UI in the industry.

So, it seems to me that its quite safe to use it for the `isStream` status.

PaulRBerg:
I don't think LLMs are a good source to quote when performing security analysis like this (because the training data for EVM development is small). Posting a question on StackExchange would be a much more reliable method.

At any rate, I agree that it should be safe to use zero address as the `isStream` status, but we should still write tests for this.

smol-ninja:
> I don't think LLMs are a good source to quote when performing security analysis like this

I completely agree with you. I only use deep research to gather sources that I can personally verify. I have also [asked on X](https://x.com/sy100x/status/1912980004444733521) as well as [posted a question on Stackoverflow](https://ethereum.stackexchange.com/questions/168639/what-are-known-uses-of-null-address-other-than-for-burning-tokens). So far, I’ve found no use for the null address beyond burning tokens. Chains which have assigned addresses to the native tokens do not use the null address.

If a chain were to use the null address to represent its native token, it would already be excluded because of `setNativeToken` function. Its also not possible to deterministically deploy an ERC20 token to the null address.

>  we should still write tests for this — to show what happens when the zero address is a token

I’m not entirely sure what kind of test you want here. If we were to test the null address as a token, the create functions would pass, but all other functions would fail.

-------------

Also, this PR is ready for review. As discussed yesterday, I have brought back `AllowToHook` and `SetNFTDescriptor` events. However, I’m still not sure about emitting `funder` in the create events. `funder` can be tracked using ERC20 transfer event. And in case of `BatchLockup`, since `funder` is the batch lockup contract, the information of actual funder is missing from Create events anyways. So, I do not think emitting `funder` adds much value here.

src/abstracts/SablierLockupBase.sol
```diff
@@ -451,6 +428,7 @@ abstract contract SablierLockupBase is
         });
 
         // Refresh the NFT metadata for all streams.
+
```
andreivladbrg, line 431: typo

src/abstracts/SablierLockupBase.sol
```diff
@@ -651,7 +609,7 @@ abstract contract SablierLockupBase is
     /// @dev See the documentation for the user-facing functions that call this internal function.
     function _cancel(uint256 streamId) internal returns (uint128 senderAmount) {
```
andreivladbrg, line 610: can't we move this function logic inside the `external` one `cancel`?
wouldn't it bring a small optimization?

src/abstracts/SablierLockupBase.sol
```diff
@@ -397,44 +387,31 @@ abstract contract SablierLockupBase is
...
-
     /// @inheritdoc ISablierLockupBase
     function setNativeToken(address newNativeToken) external override onlyAdmin {
-        // Check: provided token is not zero address.
-        if (newNativeToken == address(0)) {
-            revert Errors.SablierLockupBase_NativeTokenZeroAddress();
-        }
-
         // Check: native token is not set.
         if (nativeToken != address(0)) {
```
andreivladbrg, line 409: what if we remove this check as well?

ik that from the user's pov it can be viewed as less secure, but i believe that, in general, admin-related functions should be trusted by them

(btw, i'm not myself convinced that's the right call here)

src/abstracts/SablierLockupBase.sol
```diff
@@ -588,23 +566,12 @@ abstract contract SablierLockupBase is
...
     function _isCallerStreamRecipientOrApproved(uint256 streamId, address recipient) internal view returns (bool) {
-        return msg.sender == recipient || isApprovedForAll({ owner: recipient, operator: msg.sender })
-            || getApproved(streamId) == msg.sender;
-    }
-
-    /// @notice Checks whether `msg.sender` is the stream's sender.
-    /// @param streamId The stream ID for the query.
-    function _isCallerStreamSender(uint256 streamId) internal view returns (bool) {
-        return msg.sender == _streams[streamId].sender;
+        return _isAuthorized({ owner: recipient, spender: msg.sender, tokenId: streamId });
```
andreivladbrg, line 574: sheesh, why didn't we make this change earlier?

src/abstracts/SablierLockupBase.sol
andreivladbrg: adc90d19cb4588021a741a362260d15b861c8dcc

smol-ninja:
Thank you so much Andrei for the review. 

> IMO the better option would be to check the sender

hmm I didn't notice that. In that case, checking only `sender` sounds better. There is no need to check for both `token` and `sender`. Wdyt?

> it would only read the first slot in the stream struct

I don't think gas efficiency depends on which slot you are accessing.

andreivladbrg:
> There is no need to check for both token and sender. Wdyt?

ofc not. my point is to:

```solidity
function _notNull(uint256 streamId) private view {
    if (_streams[streamId].sender == address(0)) {
        revert Errors.SablierLockupBase_Null(streamId);
    }
}
```

> I don't think gas efficiency depends on which slot you are accessing.

hmm, maybe. i remember something about different gas costs depending on (not sure read or write operation) which slots are accessed but not sure what.

src/SablierLockup.sol
```diff
@@ -452,20 +425,14 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
...
+        // Effect: update cliff time.
+        _cliffs[streamId] = cliffTime;
 
-            // Effect: set the cliff unlock amount if it is non-zero.
-            if (unlockAmounts.cliff > 0) {
-                _unlockAmounts[streamId].cliff = unlockAmounts.cliff;
-            }
-        }
+        // Effect: set the cliff unlock amount.
+        _unlockAmounts[streamId].cliff = unlockAmounts.cliff;
```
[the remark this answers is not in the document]
smol-ninja: Good catch. It reduced the contract size by 61 B.

src/SablierLockup.sol
```diff
@@ -288,47 +288,22 @@ contract SablierLockup is ISablierLockup, SablierLockupBase {
...
-            streamedAmount = LockupMath.calculateStreamedAmountLT({
-                depositedAmount: depositedAmount,
-                blockTimestamp: blockTimestamp,
-                timestamps: timestamps,
-                tranches: _tranches[streamId]
-            });
+    function _streamedAmountOf(uint256 streamId) internal view override returns (uint128) {
+        Lockup.Amounts memory amounts = _streams[streamId].amounts;
+
+        if (_streams[streamId].isDepleted) {
```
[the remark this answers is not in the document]
smol-ninja: I have reverted this change so no need for this now.

src/abstracts/SablierLockupBase.sol
smol-ninja: Yes we should. I will push a commit soon but I was also able to drop gas cost with the changes.

src/abstracts/SablierLockupBase.sol
smol-ninja: That will be a major concern from security pov. If admin is able to overwrite the native token, it can block even the USDC. Its not about trust, but what happens if admin gets compromised.

andreivladbrg: > it can block even the USDC.

that can happen no matter if we remove this (when the native token is not configured yet)

> Its not about trust, but what happens if admin gets compromised

i agree with you, but the vast majority of chains won't have the native token configured, so this concern will still exist.

src/abstracts/SablierLockupBase.sol
```diff
@@ -169,7 +169,8 @@ abstract contract SablierLockupBase is
 
     /// @inheritdoc ISablierLockupBase
     function isStream(uint256 streamId) external view override returns (bool result) {
-        result = _streams[streamId].isStream;
+        // If a stream exists, the token address is not zero.
+        result = address(_streams[streamId].token) != address(0);
```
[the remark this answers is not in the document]
andreivladbrg: actually this was the previous method to check if a stream is null (before adding the boolean value) https://github.com/sablier-labs/lockup/commit/68bfd9031247da96b0f7db51f34970524045728f

src/abstracts/SablierLockupBase.sol
smol-ninja: Thats a fair point.. I agree with you.

andreivladbrg: the only diff between removing that check or not is whether the native token acts like a "late" immutable variable.

src/abstracts/SablierLockupBase.sol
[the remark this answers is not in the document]
smol-ninja: Then why did we change it?

I remember Paul mentioning that a valid stream can have null address as the sender but since we have null address check for the sender, I think we should definitely bring it back.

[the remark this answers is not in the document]
andreivladbrg: > Then why did we change it?

IIRC it was related to this post: https://ethereum.stackexchange.com/questions/13167/are-there-well-solved-and-simple-storage-patterns-for-solidity

or another one from Rob Hitchens

src/abstracts/SablierLockupBase.sol
andreivladbrg: > , I am not sure tbh

me neither

> What if we only allow this function on chains where native token has an address? That would solve all these problems.

that would indeed be a solution — but the problem with this is that it would require constantly maintaining both branches (remember the blast one — it wasn't that friendly)

src/abstracts/SablierLockupBase.sol
smol-ninja: It had no effect on the contract size. So, I didn't make this change. 

I also tried moving `_withdraw` inside `withdraw` but it increases the contract size by almost 400B which is weird.

src/abstracts/SablierLockupBase.sol
smol-ninja: You are right. I think it's best to leave it as it is for now since the contract size isn't an issue. I have another idea which is to keep transfer `msg.value` to another address on each tx. That way, we won't have to deal with native token anymore. We can even support native token in streams.

The problem with set-native-token is that we may not always know the native token's address when a new chain is supported. By the time we identify it, users might have already created streams using the native token and that would be a bigger problem.

smol-ninja: > transfer msg.value to another address on each tx.

This would be more expensive though.

andreivladbrg: > The problem with set-native-token is that we may not always know the native token's address when a new chain is supported

good point

>  I think it's best to leave it as it is for now since the contract size isn't an issue

agree

smol-ninja:
One of the major contributor to code size was using structs within internal functions. I changed it in https://github.com/sablier-labs/lockup/pull/1244/commits/4964c97d4b64e594a849dca2dfb4d2c6f652f55d and it reduced contract size by 700 Bytes. Now the new contract size is 1,040 Bytes below the limit 🎉🎉🎉

andreivladbrg:
> One of the major contributor to code size was using structs within internal functions. I changed it in [4964c97](https://github.com/sablier-labs/lockup/commit/4964c97d4b64e594a849dca2dfb4d2c6f652f55d) and it reduced contract size by 700 Bytes. Now the new contract size is 1,040 Bytes below the limit 🎉🎉🎉

oh wow, very nice finding, congrats 👏🏻
so now we have room to play

src/abstracts/SablierLockupBase.sol
andreivladbrg: > but it increases the contract size by almost 400B which is weird

interesting. it means calling public functions inside other public functions increases the size a lot. i think that's the same way it works with modifiers — it copies that logic into the function.

thus, the current version of the `multiple` functions is much less efficient in terms of size

smol-ninja: _withdraw is internal and withdraw is public. So public function is calling internal.

src/abstracts/SablierLockupBase.sol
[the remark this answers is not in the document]
smol-ninja: I read that post but couldn't understand how it influenced the decision. Was isStream introduced for extreme clarity? I guess so.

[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/lockup/pull/1244/commits/26551eb07c1e586d213e963d419aad27a43547c6

andreivladbrg (approved):
looks good on my end

smol-ninja:
Thanks I will merge it then. I think @PaulRBerg would like to have a look at it during PRB's internal audit.

PR #1245, opened 2025-04-15
ci: new cache mechanism

PaulRBerg:


smol-ninja:
This did not trigger any CI

PaulRBerg:
We might have to merge it first and then push commits on `staging`

smol-ninja:
May be. But it should work in the PR as well (thats how I tested the `paths-filter` changes - https://github.com/sablier-labs/flow/pull/432/checks)

PaulRBerg:
I've originally targeted `main` and then changed it to `staging`, could be related to that

PR #1247, opened 2025-04-25
feat: state contract

### LINKED ISSUE #1212: Change the type of the constructor params to `address` instead of interfaces
Recently, when we implemented the [Chainlink oracle PR](https://github.com/sablier-labs/airdrops/pull/75) in Airdrops, we decided not to use an interface/contract type in the constructor parameters (see [here](https://github.com/sablier-labs/airdrops/blob/459f8f4b0f61ae9cafa195c81894c1a9e35d4f60/src/SablierMerkleFactoryLT.sol#L25)) in order to avoid the need to import the type in multiple places where it was only necessary to cast and pass it.

We should do the same here for `initialNFTDescriptor`. (similar in `flow`)

cc @sablier-labs/evm — lmk if you have any objections.

### LINKED ISSUE #1222: Reclassify functions under new categories
As discussed in https://github.com/sablier-labs/airdrops/discussions/110, reclassify the functions as the following:

USER-FACING / CONTRACT-INTERNAL + READ-ONLY / STATE-CHANGING + FUNCTIONS

### LINKED ISSUE #1224: Refactor "LockupBase" into "LockupState"
As discussed in https://github.com/sablier-labs/lockup/discussions/1213, restructure the contract as the following:

- [x] Rename `SablierLockupBase` into `SablierLockupState`.
- [x] Move all state variables, associated getters and `notNull` modifier into `SablierLockupState`.
- [x] Everything else into `SablierLockup`.

#### Note

`SablierLockup` should inherit from `Batch`, `NoDelegateCall`, `Adminable`, `SablierLockupState` and `ERC721`.

If it looks good after implementation, we can make the same changes in other repos.

andreivladbrg:
### Changelog

- Closes #1224
- Closes #1212 
- Closes #1222


andreivladbrg:
@smol-ninja the PR is pretty much finished, but i have one thing i'm not sure about.

currently, the `setNativeToken` and `setNFTDescriptor` functions are implemented in the `SablierLockup` contract, tho they are declared in the `SablierLockupState` contract — IMO we should move them into the state contract, wdyt?

smol-ninja:
Why not have them in `SablierLockup`? As per the OP they should be in `SablierLockup`.  If the argument is that because they are setters then `allowToHook` is also a setter, which would again take us back to the original problem.

andreivladbrg:
> Why not have them in `SablierLockup`? As per the OP they should be in `SablierLockup`. If the argument is that because they are setters then `allowToHook` is also a setter, which would again take us back to the original problem.

changed my mind because they are admin-related setters — it would require inheriting `RoleAdminable` in the state contract, so it wouldn't work

tests/integration/fuzz/lockup-base/cancel.t.sol
smol-ninja, line 1: the directory name should be `lockup-state`

src/interfaces/ISablierLockup.sol
```diff
@@ -40,40 +76,167 @@ interface ISablierLockup is ISablierLockupBase {
         uint256 indexed streamId, Lockup.CreateEventCommon commonParams, LockupTranched.Tranche[] tranches
     );
 
+    /// @notice Emitted when the accrued fees are collected.
+    /// @param admin The address of the current contract admin.
+    /// @param feeRecipient The address where the fees will be collected.
+    /// @param feeAmount The amount of collected fees.
+    event CollectFees(address indexed admin, address indexed feeRecipient, uint256 feeAmount);
```
smol-ninja: Should be above Create events.

src/abstracts/SablierLockupState.sol
```diff
@@ -0,0 +1,272 @@
...
+        view
+        override
+        notNull(streamId)
+        returns (uint128 refundedAmount)
+    {
+        refundedAmount = _streams[streamId].amounts.refunded;
+    }
+
+    /// @inheritdoc ISablierLockupState
+    function getCliffTime(uint256 streamId) external view override notNull(streamId) returns (uint40 cliffTime) {
```
smol-ninja: Should be above `getDepositedAmount`

smol-ninja (approved):
Thanks for the PR. See the below comments which I have addressed in https://github.com/sablier-labs/lockup/pull/1247/commits/342bfe09ee0012e02ec7c0bcb6f102cbe5baa69f. If you agree, feel free to merge it.

tests/integration/fuzz/lockup-base/cancel.t.sol
andreivladbrg: nice catch

src/interfaces/ISablierLockup.sol
andreivladbrg: correct

src/abstracts/SablierLockupState.sol
andreivladbrg: correct

andreivladbrg:
> [342bfe0](https://github.com/sablier-labs/lockup/commit/342bfe09ee0012e02ec7c0bcb6f102cbe5baa69f). If you agree, feel free to merge

thanks, looks good, merging

Issue #1235, opened 2025-04-04
Refactor `SablieLockup` create functions to allow array of parameters


smol-ninja:
Discussed in https://github.com/sablier-labs/lockup/discussions/1114#discussioncomment-12714607.

- [x] Remove `BatchLockup` contract and associated strucuts.
- [x] Refactor all create functions to accept an array of parameters except for `token`.
- [x] Have single ERC20 transfer
- [x] Update `aggregateAmount` only once 
- [x] Verify contract size within limit

<details><summary>Click to see demo code</summary>

```solidity
 function createWithDurationsLL(
        IERC20 token,
        BatchLockup.CreateWithDurationsLL[] calldata batch
    )
        external
        payable
        override
        noDelegateCall
        returns (uint256[] memory streamIds)
    {
        // Check that the batch size is not zero.
        uint256 batchSize = batch.length;
        if (batchSize == 0) {
            revert Errors.SablierBatchLockup_BatchSizeZero();
        }

        uint128 amountDeposited;

        // Use the block timestamp as the start time.
        uint40 startTime = uint40(block.timestamp);

        streamIds = new uint256[](batchSize);

        for (uint256 i = 0; i < batchSize; ++i) {
            // Set the current block timestamp as the stream's start time.
            Lockup.Timestamps memory timestamps = Lockup.Timestamps({ start: startTime, end: 0 });

            uint40 cliffTime;

            // Calculate the cliff time and the end time.
            if (batch[i].durations.cliff > 0) {
                cliffTime = timestamps.start + batch[i].durations.cliff;
            }
            timestamps.end = timestamps.start + batch[i].durations.total;

            // Checks, Effects and Interactions: create the stream.
            streamIds[i] = _createLL(
                Lockup.CreateWithTimestamps({
                    sender: batch[i].sender,
                    recipient: batch[i].recipient,
                    depositAmount: batch[i].depositAmount,
                    token: token,
                    cancelable: batch[i].cancelable,
                    transferable: batch[i].transferable,
                    timestamps: timestamps,
                    shape: batch[i].shape
                }),
                batch[i].unlockAmounts,
                cliffTime
            );

            amountDeposited += batch[i].depositAmount;
        }

        // Transfer the total amount to the contract.
        token.safeTransferFrom({ from: msg.sender, to: address(this), value: amountDeposited });

        aggregateAmount[token] += amountDeposited;
    }
```
</details>

razgraf:
[...]

PR #1251, opened 2025-05-14
refactor: batchification of create functions

smol-ninja:
## Changelog
- Closes https://github.com/sablier-labs/lockup/issues/1235.
- [ ] Write tests after the changes have been agreed upon.

## Preamble

With one leg into "stack too deep" error and another leg into contract size limit, I tried several iterations and this is the best I could come up with to avoid both stack too deep error and contract size issue, without making large refactors (such as delegating more common logic to the libraries).

_Now with 300 runs, the contract size is within limits (28B margin) and there are no stack too deep issues as well._

#### Other approaches I had taken:

- Use of `transient` storage to store interim `streamId` to prevent stack too deep error but it didn't help. One reason is that `forge fmt` throws error with `transient` keyword. Another reason is that `cancun` throws error when `shaghai` doesn't. For ex, if you change evm to `cancun` on the `staging` branch, it would throw stack too deep error with `NFTDescriptorMock` contract.
- I also wanted to explore if we could only read `nextStreamId` once and update it only once at the end of the execution. But it causes stack error due to more local variables.
- Use of local variables (not memory structs) can potentially reduce contract size but causes stack error whereas using memory structs solves stack error but increases contract size.

I also have a feeling that somehow we can DRY'ify the following lines since they are repeated in each of the functions but couldn't figure out how without increasing the contract size or causing stack too deep error.

```solidity
unchecked {
    // Update the total deposit amount.
    totalDepositAmount += batchParams[i].commonParams.depositAmount;
}
```

I am also open to more ideas.


smol-ninja:
Closing this PR as decided in https://github.com/sablier-labs/lockup/discussions/1114#discussioncomment-13218843.

PR #1254, opened 2025-05-22
chore: just scripts

PaulRBerg:
Implements https://github.com/sablier-labs/command-center/issues/151

lmk if you like this. If yes, we can implement the same in the other repos

See https://github.com/sablier-labs/devkit


package.json
```diff
@@ -13,6 +13,7 @@
   "dependencies": {
     "@openzeppelin/contracts": "5.3.0",
     "@prb/math": "4.1.0",
+    "@sablier/devkit": "github:sablier-labs/devkit#main",
```
smol-ninja, line 16: Why isn't it a part of `devDependencies`?

CONTRIBUTING.md
```diff
@@ -5,35 +5,36 @@
...
 $ git clone git@github.com:sablier-labs/lockup.git
 ```
 
-Then, inside the project's directory, run this to install the Node.js dependencies and build the contracts:
+Then, inside the project's directory, run these commands to install the Node.js dependencies and build the contracts:
 
 ```shell
-$ bun install
-$ bun run build
+$ just install
```
smol-ninja, line 36: This command is throwing error: https://app.warp.dev/block/k0sefPceWxeXcWRfYdChru

Should it be `bun install`?

package.json
```diff
@@ -62,23 +63,7 @@
   },
   "repository": "github.com/sablier-labs/lockup",
   "scripts": {
-    "build": "forge build",
-    "build:optimized": "FOUNDRY_PROFILE=optimized forge build",
```
smol-ninja, line 66: Even though I like `Just`, I am of the opinion that having these commands here gives us more flexibility over development. For ex. `just --list` prints list of available commands but doesn't provide info on what they will be executing without looking up the other repo.

What if we only move styling related commands to devkit and keep forge related commands in a separate Justfile that can be placed in this repo? This would give us more control over adjusting those commands during development.

Screenshot 2025-05-27 at 11 51 35

smol-ninja (changes requested):

PaulRBerg:
CI error is because we need to install [ni](https://github.com/antfu-collective/ni) in CI:

> bash: line 1: na: command not found

smol-ninja:
Understand. In that case, we should add it under `Pre Requisites`.

smol-ninja:
Bumping this up, in case you have missed my review @PaulRBerg, ignore otherwise.

package.json
PaulRBerg: Huh. Mistake!

CONTRIBUTING.md
PaulRBerg: It is working but you need to have [`@antfu/ni`](https://github.com/antfu-collective/ni) package installed globally on your machine.

package.json
PaulRBerg: Just is super powerful and flexible.

- You can run `just --show RECIPE` to see the recipe's script **without** running it.
- You can run `just --explain RECIPE` to see the recipe's docs **while also** running it.
- You can override all recipes because I set [`allow-duplicate-recipes`](https://just.systems/man/en/settings.html#settings) to `true`. So if you want to quickly prototype something you can either override `forge-check` and the `devkit` recipe will not be run at all, or you can extend it by defining a new recipe called `my-awesome-check`.

PaulRBerg:
The PR is ready for re-review — and perhaps merge this time! @sablier-labs/evm

PaulRBerg:
Merging now — make sure you have `just` and `@antfu/ci` on your machines, @sablier-labs/evm 🫡

https://github.com/sablier-labs/team-setup/blob/94103a1b664019ee275de4758e2823ca542adf1c/install-globals.sh

PR #1255, opened 2025-05-22
fix: sophscan verification

lutr0:


smol-ninja (approved):
Thank you ser. Appreciate it.

pikachu

PR #1263, opened 2025-06-22
build: remove optimized profile

andreivladbrg:


andreivladbrg:
@smol-ninja I'm pretty sure there's a bug in Foundry, as running `forge build --sizes` returns:

```
| SablierLockup  | 40,971  | 42,654 | -16,395 | 6,498 |
```

but it shouldn't.

The issue is that it inherits the Lite profile (`via_ir = false`).

If I remove the lite profile as well, then `forge build --sizes` correctly returns:

```
| SablierLockup | 22,633 | 23,677 | 1,943 | 25,475   |
```

--- 

 and re the compilation time — I was wrong — it actually increases if we remove the lite profile as well (which is normal):

image

smol-ninja:
I see. What do you propose then since both are not looking good?

andreivladbrg:
postpone or give up on this idea? 😆

andreivladbrg:
> Oh no, that sounds bad too 😅.

i'm not sure if you're being sarcastic 🙈

having only one profile is a good thing since it's easier to understand the setup, but it comes with longer compilation time, which isn't great

i'd say let's postpone and just check from time to time if the bug is fixed?

PR #1284, opened 2025-08-27
build: use routeme rpcs for all chains

smol-ninja:
Addresses https://github.com/sablier-labs/command-center/issues/170


smol-ninja:
@sablier-labs/evm I would like to confirm if this is OK before I make similar changes in other repos?

### Motivation

Having an explicit `ROUTEME_API_KEY` variable guarantees that we are using ROUTEME rpc endpoint. This is important to reduce failing fork tests. 

I am working on other PR that will auto rerun a failed fork test but I believe having this guarantee is important for routeme team to provide us the best RPC infra.

smol-ninja:
@andreivladbrg because the failing fork tests have been PITA on `main` branch so its more relevant to `main` than `staging`. If you are concerned about the conflicts, I will provide all the help you need to resolve those conflicts.

My objective here is to fully migrate to routeme, add the re-run job and then chase Route me guy to fix failing issues for us.

smol-ninja:
Also @andreivladbrg, to support my argument, I don't think that conflicts in `foundry.toml` would be any issue at all. Its one of the simplest file we have.

smol-ninja:
> but why bother at all when we can avoid it completely?

Because its better for fork tests. Therefore, I don't agree that we should avoid it.

PR #1289, opened 2025-09-25
build: bump evm-utils for deployments

andreivladbrg:
we need the comptroller address for deployments


smol-ninja:
I say lets merge that PR first and then let me release `1.0.0` right away. So we wont have to change it again, wdyt?

smol-ninja (changes requested):

PR #1321, opened 2025-11-10
chore: correct ignore paths in codecov

andreivladbrg:
Closes https://github.com/sablier-labs/lockup/issues/1319

The coverage is now >97%.

image

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,10 +5,9 @@ coverage:
   status:
     patch: off
 ignore:
-  - "scripts"
-  - "scripts/solidity"
-  - "src/libraries/NFTSVG.sol"
-  - "src/libraries/SVGElements.sol"
-  - "src/mocks"
-  - "src/tests"
-  - "tests"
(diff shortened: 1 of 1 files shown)


smol-ninja:
For Flow and Airdrops, the coverage data matches but any idea why it is not 100% for `utils` and `SablierLockup`? I ran `just coverage utils` locally and its 100% but not on the codecov UI.

andreivladbrg:
> For Flow and Airdrops, the coverage data matches but any idea why it is not 100% for `utils` and `SablierLockup`? I ran `just coverage utils` locally and its 100% but not on the codecov UI.

might be a bug in foundry/codecov? 

image

it seems like calling `streamedAmountOf` in case of linear/dynamic streams is marked as "partially tested"

smol-ninja (approved):
Could be. May be codecov requires more data through a few more CI runs before it can correctly show the coverage. We can merge it now and check it later.

PR #1322, opened 2025-11-10
ci: run ci on PRs against main and staging

andreivladbrg:


.github/workflows/ci-airdrops.yml
```diff
@@ -8,6 +8,9 @@ concurrency:
 on:
   workflow_dispatch:
   pull_request:
+    branches:
+      - "main"
```
smol-ninja, line 12: Wouldn't it trigger this CI when a commit is pushed to current `main` branch?

andreivladbrg: wdym?

these are declared under the `pull_request` which means the CI should run on both:
1. commits pushed to `main` & `staging`
2. PRs open on top of `main` & `staging`

smol-ninja: nvm. I think it wont run.

My question, however, was that since `main` is not a monorepo right now, can this CI be triggered when we push to `main`? I guess no because this CI is not part of `main` yet.

smol-ninja: But, why do we need `main` and `staging` under `pull_request`? IMO we should run CI on every `pull_request` and not just on `main` or `staging`.

I agree having `main` and `staging` under `push` through.

andreivladbrg: > My question, however, was that since `main` is not a monorepo right now, can this CI be triggered when we push to `main`? I guess no because this CI is not part of `main` yet.

oh, yeah, it won't run, for the reason you mentioned, this code doesn't exist on `main`

> why do we need main and staging under pull_request? IMO we should run CI on every pull_request and not just on main or staging.

i agree on running the CI on all PRs, but before this change - on this PR: https://github.com/sablier-labs/lockup/pull/1321 - the CI was not running unless i added these specifically. i believe the reason the CI didn't run is that the CI workflows are not part of main yet and exist only on staging (it was not even running the cached version)

smol-ninja: May be it didn't run in that PR for other reasons. I just checked https://github.com/sablier-labs/lockup/pull/1320, and CI ran there.

smol-ninja: I created a new PR (https://github.com/sablier-labs/lockup/pull/1323) and it triggered CI so I think its working as expected. We should close this PR then.

andreivladbrg: right, it didn't run because of these paths declared:

https://github.com/sablier-labs/lockup/blob/4446bf1a972a025be42130b6a1b6485bf29bb0b3/.github/workflows/ci-lockup.yml#L21-L26

we can close it

PR #1324, opened 2025-11-12
test(lockup): fix tokenURI tests

smol-ninja:


lockup/tests/integration/concrete/lockup/token-uri/tokenURI.t.sol
```diff
@@ -17,14 +17,23 @@ import { Integration_Test } from "../../../Integration.t.sol";
 /// 0x1804c8AB1F12E6bbf3894d4083f33e07309d1f38 so that the deployed contracts have the same addresses as
 /// the values hard coded in the tests below
 contract TokenURI_Lockup_Integration_Concrete_Test is Integration_Test {
-    address internal constant LOCKUP = 0x923b5Ab3714FD343316aF5A5434582Fd16722523;
+    address internal constant LOCKUP = 0x3D7Ebc40AF7092E3F1C81F2e996cbA5Cae2090d7;
```
andreivladbrg, line 21: should we add a comment?
 `// TODO: Update anytime we have a new release`

smol-ninja: That would be a forever `TODO` 😂. I don't prefer forever TODOs but if you like it, I am OK adding it.

andreivladbrg: yeah, a forever one 
if not a `TODO`, i'd like a comment that's highlighted in the IDE so it draws your attention
do you know any other alternative?

smol-ninja: Yeah its possible. Make sure to add the following to VSCode's `settings.json`

```json
  "todo-tree.general.tags": [
    "TODO",
    "FIXME",
    "BUG",
    "NOTE",
    "REVIEW"
  ],
```

By default, todo tree only supports the first three so we can extend it with `NOTE` and `REVIEW` (I added both because sometimes we can use `REVIEW` such as `REVIEW: this address after staging is merge`

https://github.com/sablier-labs/lockup/pull/1324/commits/5eed01f90c0304d7399c679c73b75c3d9a0497ae

andreivladbrg (approved):

PR #1326, opened 2025-11-15
chore: Fix mathematical formula notation in LL stream documentation

sunnyraindy:
change "=>" to ">=" for proper inequality representation


smol-ninja:
Good catch @sunnyraindy

PR #1352, opened 2025-11-27
chore: fix solhint configurations

smol-ninja:
### Changelog

- This PR adds rules to individual solhint files since solhint does not support inheritance from root folder. 
- Also fixes solhint warnings.


airdrops/tests/.solhint.json
```diff
@@ -1,13 +1,21 @@
 {
+  "extends": "solhint:recommended",
```
andreivladbrg, line 2: do we need this?

can't we remove it so we can also remove more of the `off` configs below

andreivladbrg (approved):
looks good, just one question

airdrops/tests/.solhint.json
smol-ninja: If we remove this then we would have to add many useful rules from recommended config. Previously, it didn't have `extends` config and was ignoring a few important rules.

Ideally, we should be able to extend from the root config file and then override some of the rules but because its not an npm workspace but individual npm packages, it does not work.

So when use `just full-check lockup`, because it `cd` to lockup dir, the solhint in `lockup` dir does not use rules from the solhint in the root dir.

andreivladbrg: > then we would have to add many useful rules from recommended config

so the question is if we need more than 50% of the rules from the "recommended config" or not

lets leave it like this for now

smol-ninja: > so the question is if we need more than 50% of the rules from the "recommended config" or not

may be or may be not. 

> lets leave it like this for now

Yes that would be better. Thanks.

smol-ninja:
The failing fork test is fixed in https://github.com/sablier-labs/lockup/pull/1366 (see description which includes https://github.com/sablier-labs/lockup/commit/2539c05c7e792073b0b14d2b5cb769c448561c0f)

PR #1353, opened 2025-11-27
test(airdrops): change type of `MerkleBuilder` from library to contract

smol-ninja:
This PR addresses refactors required for the invariants PR. 

### Changelog

- Change type of `MerkleBuilder` from `library` to `contract`. Why? Because this allowed me to add all merkle tree related functions to `MerkleBuilder` such as `computeMerkleProof` and `constructMerkleTree`. Previosuly these were part of `Fuzzers`.
- Keep only fuzzing related functions in Fuzzers.
- So now we have `MerkleBuilder` which deals with merkle tree related functions and it makes it easy to import into invariant handlers.


andreivladbrg (approved):
LGTM 👍

PR #1354, opened 2025-11-27
test(airdrops): add invariant tests for merkle contracts

### LINKED ISSUE #7: Write invariant tests

smol-ninja:
Closes https://github.com/sablier-labs/airdrops/issues/7

This is blocked by https://github.com/sablier-labs/lockup/pull/1353

Feel free to suggest more invariants. I only added what I could think of as making sense.


airdrops/tests/invariant/README.md
```diff
@@ -0,0 +1,10 @@
+### List of Invariants
+
+#### For all campaigns:
+
+1. token.balanceOf(campaign) = total deposit - $\sum$ claimed - $\sum$ clawbacked
+2. `hasClaimed` should never change its value from `true` to `false`
```
andreivladbrg, line 6: ideas for more:
- `minFeeUSD` never increases
- similar to the second one 2. - `hasExpired` should never change its value from `true` to `false`
- `firstClaimTime == 0`→ clawback succeeds

wdyt?

airdrops/tests/invariant/README.md
```diff
@@ -0,0 +1,10 @@
...
+
+#### For VCA campaign:
+
+1. For MerkleVCA campaigns, total forgone = $\sum$ claim requested - $\sum$ claimed
```
andreivladbrg: one more idea for VCA: `forgone monotonicity`

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,218 @@
...
+
+    uint256 internal aggregateAmount;
+
+    /// @dev The campaign token being used in the handler.
+    IERC20 public campaignToken;
+
+    address public comptroller;
+
+    /// @dev Store leaves as `uint256` in storage so that we can use OpenZeppelin's {Arrays.findUpperBound}.
+    uint256[] internal leaves;
```
andreivladbrg, line 26: wouldn't it make more sense to have these in store? (mapped by campaign addresses)

`leavesData` as well

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,218 @@
...
+        uint256 timeJump,
+        uint256 amount
+    )
+        external
+        adjustTimestamp(timeJump)
+        instrument("fundCampaign")
+        isDeployed
+    {
+        // Bound amount to be less than aggregate amount.
+        amount = bound(amount, 1, aggregateAmount);
```
andreivladbrg: bounding this starting from 1 will result in a lot of discarded claim tests IMO here:

```solidity
vm.assume(leafData.amount <= campaignToken.balanceOf(address(campaign)));
``` 

wdys about changing it to half or a quarter of `aggregateAmount`?


```solidity
uint256 lowerBound = aggregateAmount / 2;
amount = bound(amount, lowerBound, aggregateAmount);
``` 

and at the same time, restrict how many times `fundCampaign` is called in favor of more `claim` calls

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,218 @@
...
+    modifier adjustTimestamp(uint256 timeJump) {
+        timeJump = bound(timeJump, 0, 30 days);
+        skip(timeJump);
+        _;
+    }
+
+    /// @dev Assume common deployment parameters.
+    modifier assumeDeployParams(address campaignCreator, LeafData[] memory rawLeavesData) {
+        // Deploy campaign only if it hasn't been deployed yet.
+        vm.assume(totalCalls["deployCampaign"] == 0);
```
andreivladbrg, line 57: why restrict the invariants test to a _single_ campaign per model?

i would allow up to 10, or at least a campaign per unique token

airdrops/tests/invariant/handlers/MerkleInstantHandler.sol
```diff
@@ -0,0 +1,57 @@
...
+        params.campaignName = CAMPAIGN_NAME;
+        params.campaignStartTime = getBlockTimestamp();
+        params.expiration = getBlockTimestamp() + 365 days;
+        params.initialAdmin = campaignCreator;
+        params.ipfsCID = IPFS_CID;
+        params.merkleRoot = merkleRoot;
+        params.token = campaignToken;
+
+        // Deploy and return the campaign address
+        campaign = ISablierMerkleBase(new SablierMerkleInstant(params, campaignCreator, comptroller));
```
andreivladbrg: why not use factories in the invariant tests? using them is good to also test the create2 salts

airdrops/tests/invariant/stores/Store.sol
andreivladbrg, line 1: why not call it `<protocol>Store.sol`, as we do in the other ones?

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
```diff
@@ -0,0 +1,79 @@
...
+        MerkleVCA.ConstructorParams memory params;
+
+        params.campaignName = CAMPAIGN_NAME;
+        params.campaignStartTime = getBlockTimestamp();
+        params.expiration = getBlockTimestamp() + 365 days;
+        params.initialAdmin = campaignCreator;
+        params.ipfsCID = IPFS_CID;
+        params.merkleRoot = merkleRoot;
+        params.token = campaignToken;
+        params.unlockPercentage = VCA_UNLOCK_PERCENTAGE;
```
andreivladbrg, line 58: unlike LL and LT campaigns (which use lockup) i think it would be a good idea to fuzz the unlock pct here

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,218 @@
...
+
+        // Ensure raw leaves data has more than one leaf.
+        vm.assume(rawLeavesData.length > 1);
+
+        // Ensure campaign creator is not the zero address.
+        vm.assume(campaignCreator != address(0));
+
+        // Ensure campaign creator is not blacklisted.
+        assumeNoBlacklisted({ token: address(campaignToken), addr: campaignCreator });
```
andreivladbrg: we don't need this in the invariant tests, but only in the fork ones

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,218 @@
...
+    /// @dev Simulates the passage of time.
+    /// @param timeJump A fuzzed value for time warps.
+    modifier adjustTimestamp(uint256 timeJump) {
+        timeJump = bound(timeJump, 0, 30 days);
+        skip(timeJump);
+        _;
+    }
+
+    /// @dev Assume common deployment parameters.
+    modifier assumeDeployParams(address campaignCreator, LeafData[] memory rawLeavesData) {
```
andreivladbrg, line 55: were we getting the stack too deep error?

since its used only in the `deployCampaign` function, can't we move the logic just there?

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,218 @@
...
+        setMsgSender(campaign.admin());
+
+        // Clawback funds.
+        campaign.clawback(address(this), amount);
+
+        // Update clawback amount in store.
+        store.updateTotalClawbackAmount(address(campaign), amount);
+    }
+
+    /// @notice Helper function to deploy a Merkle Instant campaign with fuzzed leaves data and token.
```
andreivladbrg: "Merkle campaign" without "Instant"

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
```diff
@@ -0,0 +1,79 @@
...
+    /*//////////////////////////////////////////////////////////////////////////
+                                      HELPERS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function _claim(LeafData memory leafData, bytes32[] memory merkleProof) internal override {
+        SablierMerkleVCA merkleVCA = SablierMerkleVCA(address(campaign));
+
+        // Calculate the claim amount.
+        uint128 claimAmount =
+            merkleVCA.calculateClaimAmount({ fullAmount: leafData.amount, claimTime: getBlockTimestamp() });
```
andreivladbrg: what would you say about moving `BaseTest.calculateMerkleVCAAmounts()` to `Utils` and using it here instead of the campaign logic itself?

the rationale is that the store values are updated according to the campaign logic itself, so it won’t revert in case of issues

airdrops/tests/invariant/stores/Store.sol
```diff
@@ -0,0 +1,109 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { BaseUtils } from "@sablier/evm-utils/src/tests/BaseUtils.sol";
```
andreivladbrg: unneeded import and inheritance

andreivladbrg (commented):
leaving my review bellow:

airdrops/tests/invariant/README.md
smol-ninja: - Since `newFeeUSD` is a state changed by [a state setter function](https://github.com/sablier-labs/lockup/blob/airdrops-invariant-tests/airdrops/src/abstracts/SablierMerkleBase.sol#L181) and given that it has [a concrete test](https://github.com/sablier-labs/lockup/blob/airdrops-invariant-tests/airdrops/tests/integration/concrete/campaign/shared/lower-min-fee-usd/lowerMinFeeUSD.tree#L5) for upper bound, I do not think its a good invariant. Otherwise any state variable with upper/lower bounds would have an invariant.

- We can have `hasExpired` invariant. 

- Wouldn't having [this test](https://github.com/sablier-labs/lockup/blob/airdrops-invariant-tests/airdrops/tests/integration/concrete/campaign/shared/clawback/clawback.tree#L5) be sufficient for this?

airdrops/tests/invariant/README.md
smol-ninja: Sounds good. Would you like to add it?

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: Store contains variables used by the Invariants. So based on that, `leaves` and `leavesData` make more sense here.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: Sounds good.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: IMO having a single campaign is sufficient because each campaign would have a very large leaves data. And the invariants focus on the accuracy of those states.

Plus there will be several invariant runs over time (on every PR etc.).

airdrops/tests/invariant/handlers/MerkleInstantHandler.sol
smol-ninja: isn't `create2` tested sufficiently in fuzz tests? I don't see a point of testing create2 here until there is a create2 related invariant. Wdyt?

airdrops/tests/invariant/stores/Store.sol
smol-ninja: May be we should call it `Store.sol` there too? Its in the airdrops folder so I didn't think it needed protocol prefix.

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: Sounds good.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: Yeah you are right.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: I don't remember by I guess yes.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: Yup. Good catch as always.

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: OK sounds good to me.

airdrops/tests/invariant/README.md
andreivladbrg: > * Otherwise any state variable with upper/lower bounds would have an invariant.
> *

in case of airdrops campaign, `minFeeUSD` its the only state mutable variable that has a setter


> * Wouldn't having [this test](https://github.com/sablier-labs/lockup/blob/airdrops-invariant-tests/airdrops/tests/integration/concrete/campaign/shared/clawback/clawback.tree#L5) be sufficient for this?

fair enough, i guess it is

smol-ninja: > in case of airdrops campaign, minFeeUSD its the only state mutable variable that has a setter

You are right. And I think it should have a fuzz test then. Feel free to add an invariant or a fuzz test instead of invariant if you like. I will let you decide.

airdrops/tests/invariant/handlers/MerkleInstantHandler.sol
andreivladbrg: yeah, its enough

airdrops/tests/invariant/stores/Store.sol
andreivladbrg: thats a good idea, i am in favor of renaming them to `Store.sol` - as we have the `Invariant.t.sol`, `Integration.t.sol` etc.

smol-ninja: https://github.com/sablier-labs/lockup/issues/1387

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: good point, you are correct, lets keep them here

airdrops/tests/invariant/README.md
andreivladbrg: > And I think it should have a fuzz test then

how can it be a fuzz test? it requires a random sequence of calls

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: fair enough, lets keep it

airdrops/tests/invariant/README.md
smol-ninja: Yeah I guess I am overthinking on this. Lets have the invariant as you suggested.

andreivladbrg: thanks

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
andreivladbrg: 8b66291dbd8027ffcb522f5081b7a0cfbeb2e7a1

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: 8b66291dbd8027ffcb522f5081b7a0cfbeb2e7a1

airdrops/tests/invariant/README.md
andreivladbrg: 8b66291dbd8027ffcb522f5081b7a0cfbeb2e7a1

airdrops/tests/invariant/README.md
andreivladbrg: 8b66291dbd8027ffcb522f5081b7a0cfbeb2e7a1

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
andreivladbrg: i decided not to implement this - if you are also ok, feel free to resolve this

andreivladbrg (approved):
if my commit looks good, feel free to mark them as resolved and merge it

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: After looking at the diff, I do not agree that we should replace  `merkleVCA.calculateClaimAmount` with `calculateMerkleVCAAmounts`. Invariants should apply on the contract states and the values returned by the public functions. 

So even if they both mirrors the same logic, we should still be using `calculateClaimAmount` here.

And sorry, I misunderstood your original comment.

airdrops/tests/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,188 @@
...
+                uint256 claimedIndex = claimedIndexes[j];
+                assertTrue(
+                    ISablierMerkleBase(campaign).hasClaimed(claimedIndex),
+                    unicode"Invariant violation: claim status changed from true to false"
+                );
+            }
+        }
+    }
+
+    /// @dev Once a campaign has expired, it should never become unexpired.
```
smol-ninja:
```suggestion
    /// @dev Once a campaign has expired, it should never become active again.
```

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,246 @@
...
+
+    /// @dev Ensures the campaign has been deployed.
+    modifier isDeployed() {
+        vm.assume(address(campaign) != address(0));
+        _;
+    }
+
+    /// @dev Limits the number of calls to a specific function.
+    modifier limitNumberOfCalls(string memory functionName, uint256 max) {
+        vm.assume(totalCalls[functionName] < max);
```
smol-ninja: Because you are calling `instrument` before `limitNumberOfCalls`, it should be `<=`. `max` is the maximum value.

airdrops/tests/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,188 @@
...
+    function invariant_ExpirationStatusTransition() external view {
+        address[] memory campaigns = store.getCampaigns();
+
+        for (uint256 i = 0; i < campaigns.length; ++i) {
+            ISablierMerkleBase campaign = ISablierMerkleBase(campaigns[i]);
+
+            // If the campaign has an expiration time set and current time is past it, it must be expired.
+            if (campaign.EXPIRATION() > 0 && block.timestamp > campaign.EXPIRATION()) {
+                assertTrue(
+                    campaign.hasExpired(), unicode"Invariant violation: expiration status changed from true to false"
```
smol-ninja: Because `hasExpired` is just a proxy for `return EXPIRATION > 0 && EXPIRATION <= block.timestamp`, I now think this is not a valid invariant. Invariants are for contract states. `EXPIRATION` is a constant here.

So, I think we should revert it now. Wdyt?

airdrops/tests/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,188 @@
...
+    function invariant_MinFeeUSDNeverIncreases() external view {
+        address[] memory campaigns = store.getCampaigns();
+
+        for (uint256 i = 0; i < campaigns.length; ++i) {
+            address campaign = campaigns[i];
+
+            assertLe(
+                ISablierMerkleBase(campaign).minFeeUSD(),
+                store.minFeeUSD(campaign),
+                unicode"Invariant violation: minFeeUSD increased from deployment value"
```
smol-ninja: This is now true if there are more than 1 call to `lowerMinFeeUSD` function. So it should say

"Invariant violation: minFeeUSD increased"

airdrops/tests/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,188 @@
...
+    function invariant_ForgoneMonotonicity() external view {
+        address vcaCampaign = store.vcaCampaign();
+
+        // Skip if no VCA campaign is deployed.
+        if (vcaCampaign == address(0)) return;
+
+        assertLe(
+            store.vcaTotalForgoneAmount(),
+            ISablierMerkleVCA(vcaCampaign).totalForgoneAmount(),
+            unicode"Invariant violation: store forgone amount > campaign total forgone amount"
```
smol-ninja: "Invariant violation: total forgone amount decreases" similar to other statements. We dont mention `store`

smol-ninja (commented):
Thanks for the commit. Left few comments below:

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: Yeah I am OK with this. I don't find it necessary since its just initial unlock which does not impact the VCA logic.

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: correct

airdrops/tests/invariant/Invariant.t.sol
andreivladbrg: good point about the constant, i am ok to revert it

airdrops/tests/invariant/Invariant.t.sol
andreivladbrg: we don't need this if we remove it

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
andreivladbrg: yes, the invariants are about the state  
this logic is in the handler, which IMO should have its own mechanism for bounding fuzzing and checking that the inputs are valid for the function call

airdrops/tests/invariant/Invariant.t.sol
andreivladbrg: e98ceb7761ddff990359c56ec25c44f16ab6fbb0

airdrops/tests/invariant/Invariant.t.sol
andreivladbrg: e98ceb7761ddff990359c56ec25c44f16ab6fbb0

airdrops/tests/invariant/Invariant.t.sol
andreivladbrg: e98ceb7761ddff990359c56ec25c44f16ab6fbb0

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: e98ceb7761ddff990359c56ec25c44f16ab6fbb0

airdrops/tests/invariant/Invariant.t.sol
smol-ninja: Thanks

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: Well, yes, you make sense. Good reminder about this.

smol-ninja: Some more thoughts on this:

The values are returned by `calculateMerkleVCAAmounts`, so they are neither handler's inputs nor fuzzed (there is just a non-zero check). Plus, they are also stored in the `Store` and then used in the invariant: `total forgone amount = $\\sum$ claim amount requested - $\\sum$ claimed amount`. 

That means the invariant is checked on the values returned by a local function and not the contract's function. But, it should be on the values returned by the contract. Otherwise, we end up comparing some local values against the `totalClaimAmountRequested` (an user input).

So IMO if we want to keep `calculateMerkleVCAAmounts`, we should take a difference of claimer's balance and update the difference in `store.updateTotalClaimAmount` and similarly for `store.updateTotalForgoneAmount`, or we use contract functions here.

Let me know what you think about this point.

andreivladbrg: ok, gonna stop debating in order to move further. will revert the change

andreivladbrg (approved):

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: OK. I will merge the PR now but I would still appreciate if you can share your thoughts on this whenever you have time.

PR #1355, opened 2025-11-28
feat: add chainlink oracle for Monad and HyperEVM

andreivladbrg:
re-implements the PR https://github.com/sablier-labs/evm-utils/pull/71

i did not add the exact commits because we have a different setup here, i.e. the `foundry.base.toml` file (added you as author)

i think it would be a good choice not to push to those repos

### DIFF
--- a/airdrops/bun.lock
+++ b/airdrops/bun.lock
@@ -1,5 +1,6 @@
 {
   "lockfileVersion": 1,
+  "configVersion": 0,
   "workspaces": {
     "": {
       "name": "@sablier/airdrops",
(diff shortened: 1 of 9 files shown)


andreivladbrg:
hmm, not sure it formated diferently the `bun.lock` files 
i just wanted to add the `"configVersion": 0,`

smol-ninja (approved):
Lets not worry about bun lockfile as long as tests are passing. I think they made some changes recently. I am on bun 1.3.3 and you might be on 1.3.2

smol-ninja:
>i did not add the exact commits because we have a different setup here, i.e. the foundry.base.toml file (added you as author) i think it would be a good choice not to push to those repos

What do you mean by "would be a good choice not to push to those repos"? We will be pushing to them when making deployment to a new chain. And we need these tests because we run fork tests on those repos.

andreivladbrg:
> What do you mean by

i meant that we should make changes only to this repo on staging branch (not airdrops/flow/utils)

> We will be pushing to them when making deployment to a new chain. And we need these tests because we run fork tests on those repos.

you are right, only for these cases

PR #1361, opened 2025-12-09
docs: fix logical mistake in Helpers.sol

IaroslavMazur:


smol-ninja (approved):
nice catch.

andreivladbrg (approved):
nice catch

PR #1363, opened 2025-12-16
feat(airdrops): redistribution of forgone tokens in MerkleVCA contract

### LINKED ISSUE #1339: Implement redistribution in MerkleVCA contract
Implement redistribution logic in MerkleVCA contract. For specs, please refer to [this discussion](https://github.com/sablier-labs/airdrops/discussions/117).

Feel free to take decision on function names as well as other necessary logic required for implementation.

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1339

The actual implementation is different from what was actually proposed due to its simplicity.

### DIFF
--- a/airdrops/src/SablierMerkleVCA.sol
+++ b/airdrops/src/SablierMerkleVCA.sol
@@ -3,6 +3,7 @@ pragma solidity >=0.8.22;
 
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { ud, UD60x18 } from "@prb/math/src/UD60x18.sol";
 
 import { SablierMerkleBase } from "./abstracts/SablierMerkleBase.sol";
@@ -34,11 +35,15 @@ contract SablierMerkleVCA is
     ISablierMerkleVCA, // 2 inherited components
     SablierMerkleBase // 3 inherited components
 {
+    using SafeCast for uint256;
     using SafeERC20 for IERC20;
 
     /*//////////////////////////////////////////////////////////////////////////
                                   STATE VARIABLES
     //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierMerkleVCA
+    uint128 public immutable override AGGREGATE_AMOUNT;
 
     /// @inheritdoc ISablierMerkleVCA
     UD60x18 public immutable override UNLOCK_PERCENTAGE;
@@ -50,7 +55,13 @@ contract SablierMerkleVCA is
     uint40 public immutable override VESTING_START_TIME;
 
     /// @inheritdoc ISablierMerkleVCA
-    uint256 public override totalForgoneAmount;
+    bool public override isRedistributionEnabled;
+
+    /// @inheritdoc ISablierMerkleVCA
+    uint128 public override totalForgoneAmount;
+
+    /// @dev Tracks the full amount allocated to the recipients who claimed before the vesting end time.
+    uint128 private _fullAmountAllocatedToEarlyClaimers;
 
     /*//////////////////////////////////////////////////////////////////////////
                                     CONSTRUCTOR
@@ -75,9 +86,13 @@ contract SablierMerkleVCA is
         )
     {
         // Effect: set the immutable variables.
+        AGGREGATE_AMOUNT = campaignParams.aggregateAmount;
         UNLOCK_PERCENTAGE = campaignParams.unlockPercentage;
         VESTING_END_TIME = campaignParams.vestingEndTime;
         VESTING_START_TIME = campaignParams.vestingStartTime;
+
+        // Effect: enable redistribution if true.
+        if (campaignParams.enableRedistribution) isRedistributionEnabled = true;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -85,7 +100,7 @@ contract SablierMerkleVCA is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @inheritdoc ISablierMerkleVCA
-    function calculateClaimAmount(uint128 fullAmount, uint40 claimTime) external view returns (uint128) {
+    function calculateClaimAmount(uint128 fullAmount, uint40 claimTime) external view override returns (uint128) {
         // Zero is a sentinel value for `block.timestamp`.
         if (claimTime == 0) {
             claimTime = uint40(block.timestamp);
@@ -96,7 +111,7 @@ contract SablierMerkleVCA is
     }
 
     /// @inheritdoc ISablierMerkleVCA
-    function calculateForgoneAmount(uint128 fullAmount, uint40 claimTime) external view returns (uint128) {
+    function calculateForgoneAmount(uint128 fullAmount, uint40 claimTime) external view override returns (uint128) {
         // Zero is a sentinel value for `block.timestamp`.
         if (claimTime == 0) {
             claimTime = uint40(block.timestamp);
@@ -111,6 +126,17 @@ contract SablierMerkleVCA is
         }
 
         return fullAmount - _calculateClaimAmount(fullAmount, claimTime);
+    }
+
+    /// @inheritdoc ISablierMerkleVCA
+    function calculateRedistributionRewards(uint128 fullAmount) external view override returns (uint128) {
+        // Check: redistribution is enabled.
+        if (!isRedistributionEnabled) {
+            revert Errors.SablierMerkleVCA_RedistributionNotEnabled();
+        }
+
+        // Calculate and return the redistribution rewards.
+        return _calculateRedistributionRewards(fullAmount);
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -161,6 +187,20 @@ contract SablierMerkleVCA is
         _postProcessClaim({ index: index, recipient: recipient, to: to, fullAmount: fullAmount, viaSig: true });
     }
 
+    /// @inheritdoc ISablierMerkleVCA
+    function enableRedistribution() external override onlyAdmin {
+        // Check: the redistribution is not already enabled.
+        if (isRedistributionEnabled) {
+            revert Errors.SablierMerkleVCA_RedistributionAlreadyEnabled();
+        }
+
+        // Effect: set the value to true.
+        isRedistributionEnabled = true;
+
+        // Log the event.
+        emit RedistributionEnabled();
+    }
+
     /*//////////////////////////////////////////////////////////////////////////
                             PRIVATE READ-ONLY FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
@@ -197,6 +237,29 @@ contract SablierMerkleVCA is
         }
     }
 
+    /// @notice Calculates the redistribution rewards for a given full amount.
+    function _calculateRedistributionRewards(uint256 fullAmount) private view returns (uint128 rewards) {
+        // Return zero if total forgone amount is zero.
+        if (totalForgoneAmount == 0) {
+            return 0;
+        }
+
+        // Return zero if aggregate amount does not exceed the amount allocated to early claimers.
+        if (AGGREGATE_AMOUNT <= _fullAmountAllocatedToEarlyClaimers) {
+            return 0;
+        }
+
+        // Calculate the total amount allocated to the remaining claimers.
+        uint128 fullAmountAllocatedToRemainingClaimers;
+        unchecked {
+            // Safe to use unchecked because it cannot overflow due to above check.
+            fullAmountAllocatedToRemainingClaimers = AGGREGATE_AMOUNT - _fullAmountAllocatedToEarlyClaimers;
+        }
+
+        // Calculate the rewards.
+        rewards = ((fullAmount * totalForgoneAmount) / fullAmountAllocatedToRemainingClaimers).toUint128();
+    }
+
     /*//////////////////////////////////////////////////////////////////////////
                           PRIVATE STATE-CHANGING FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
@@ -212,21 +275,38 @@ contract SablierMerkleVCA is
         }
 
         uint128 forgoneAmount;
+        uint128 rewardAmount;
+        uint128 transferAmount = claimAmount;
 
-        // Effect: update the total forgone amount.
+        // Effect: update the total forgone amount and the total amount claimed by early claimers.
         if (claimAmount < fullAmount) {
+            // Its safe to use unchecked because none of these values can overflow.
             unchecked {
                 forgoneAmount = fullAmount - claimAmount;
                 totalForgoneAmount += forgoneAmount;
+                _fullAmountAllocatedToEarlyClaimers += fullAmount;
             }
         } else {
             // Although the claim amount should never exceed the full amount, this assertion prevents excessive claiming
             // in case of a calculation error.
             assert(claimAmount == fullAmount);
+
+            if (isRedistributionEnabled) {
+                // Calculate the reward amount.
+                rewardAmount = _calculateRedistributionRewards(fullAmount);
+
+                // Update the transfer amount if there are rewards to distribute.
+                if (rewardAmount > 0) {
+                    transferAmount += rewardAmount;
+
+                    // Log the event.
+                    emit RedistributionReward(index, recipient, rewardAmount, to);
+                }
+            }
         }
 
         // Interaction: transfer the tokens to the recipient.
-        TOKEN.safeTransfer({ to: to, value: claimAmount });
+        TOKEN.safeTransfer({ to: to, value: transferAmount });
 
         // Emit claim event.
         emit ClaimVCA(index, recipient, claimAmount, forgoneAmount, to, viaSig);
--- a/airdrops/src/SablierMerkleVCA.sol
+++ b/airdrops/src/SablierMerkleVCA.sol
@@ -91,8 +91,8 @@ contract SablierMerkleVCA is
         VESTING_END_TIME = campaignParams.vestingEndTime;
         VESTING_START_TIME = campaignParams.vestingStartTime;
 
-        // Effect: enable redistribution if true.
-        if (campaignParams.enableRedistribution) isRedistributionEnabled = true;
+        // Effect: set the enable redistribution flag.
+        isRedistributionEnabled = campaignParams.enableRedistribution;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -280,12 +280,12 @@ contract SablierMerkleVCA is
 
         // Effect: update the total forgone amount and the total amount claimed by early claimers.
         if (claimAmount < fullAmount) {
-            // Its safe to use unchecked because none of these values can overflow.
+            // Its safe to use unchecked because the value can't underflow.
             unchecked {
                 forgoneAmount = fullAmount - claimAmount;
-                totalForgoneAmount += forgoneAmount;
-                _fullAmountAllocatedToEarlyClaimers += fullAmount;
             }
+            totalForgoneAmount += forgoneAmount;
+            _fullAmountAllocatedToEarlyClaimers += fullAmount;
         } else {
             // Although the claim amount should never exceed the full amount, this assertion prevents excessive claiming
             // in case of a calculation error.
--- a/airdrops/src/interfaces/ISablierMerkleVCA.sol
+++ b/airdrops/src/interfaces/ISablierMerkleVCA.sol
@@ -7,7 +7,7 @@ import { ISablierMerkleBase } from "./ISablierMerkleBase.sol";
 /// @title ISablierMerkleVCA
 /// @notice VCA stands for Variable Claim Amount, and is an airdrop model where the claim amount increases linearly
 /// until the airdrop period ends. Claiming early results in forgoing the remaining amount, whereas claiming after the
-/// period grants the full amount that was allocated.
+/// period grants the full amount that was allocated, plus redistribution rewards if enabled.
 interface ISablierMerkleVCA is ISablierMerkleBase {
     /*//////////////////////////////////////////////////////////////////////////
                                        EVENTS
@@ -98,11 +98,9 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
     /// @dev It emits a {ClaimVCA} event, and a {RedistributionReward} event if the redistribution is enabled.
     ///
     /// Notes:
-    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, it can either cause a
-    /// race condition among the recipients or rewards would be calculated as 0 if its too low depending on how low the
-    /// value is set.
-    /// - If the campaign creator does not sufficiently fund the campaign with the actual total allocations (or
-    /// `AGGREGATE_AMOUNT`), it can cause a race condition among the recipients.
+    /// - There can be a race condition among recipients if:
+    ///   1. `AGGREGATE_AMOUNT` is set lower than the actual total allocations in the Merkle tree.
+    ///   2. The campaign is not sufficiently funded with the actual total allocations.
     /// - The rewards are transferred to the recipients at the time of claiming. If the campaign creator turns the
     /// redistribution on after the vesting end time, the recipients who have already claimed the full amount would miss
     /// on the rewards while subsequent recipients would get them.
--- a/airdrops/src/interfaces/ISablierMerkleVCA.sol
+++ b/airdrops/src/interfaces/ISablierMerkleVCA.sol
@@ -29,9 +29,23 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
         bool viaSig
     );
 
+    /// @notice Emitted when the redistribution is enabled.
+    event RedistributionEnabled();
+
+    /// @notice Emitted when a recipient receives rewards from the forgone tokens pool.
+    /// @dev Only emitted when redistribution is enabled.
+    /// @param index The index of the airdrop recipient in the Merkle tree.
+    /// @param recipient The address of the airdrop recipient.
+    /// @param amount The amount of ERC-20 tokens distributed as a reward.
+    /// @param to The address receiving the reward tokens.
+    event RedistributionReward(uint256 index, address indexed recipient, uint128 amount, address to);
+
     /*//////////////////////////////////////////////////////////////////////////
                                 READ-ONLY FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Retrieves the total amount of ERC-20 tokens allocated to the campaign.
+    function AGGREGATE_AMOUNT() external view returns (uint128);
 
     /// @notice Retrieves the percentage of the full amount that will unlock immediately at the start time. The
     /// value is denominated as a fixed-point number where 1e18 is 100%.
@@ -58,17 +72,40 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
     /// @return The amount that would be forgone, denominated in the token's decimals.
     function calculateForgoneAmount(uint128 fullAmount, uint40 claimTime) external view returns (uint128);
 
+    /// @notice Calculates the redistribution rewards for a given full amount.
+    /// @dev Notes:
+    /// - Reverts if redistribution is not enabled.
+    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, this might return 0
+    /// rather than reverting.
+    /// @param fullAmount The amount of tokens that the redistribution rewards are to be calculated for.
+    function calculateRedistributionRewards(uint128 fullAmount) external view returns (uint128);
+
+    /// @notice Retrieves a bool indicating whether the redistribution of forgone tokens is enabled or not.
+    function isRedistributionEnabled() external view returns (bool);
+
     /// @notice Retrieves the total amount of tokens forgone by early claimers.
-    function totalForgoneAmount() external view returns (uint256);
+    function totalForgoneAmount() external view returns (uint128);
 
     /*//////////////////////////////////////////////////////////////////////////
                               STATE-CHANGING FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @notice Claim airdrop. If the vesting end time is in the future, it calculates the claim amount to transfer to
-    /// the `to` address, otherwise it transfers the full amount.
+    /// the `to` address, otherwise it transfers the full amount. If the redistribution is enabled, it calculates the
+    /// reward amount based on the total amount of tokens forgone by early claimers and transfers it to the recipients
+    /// claiming after the vesting end time.
     ///
-    /// @dev It emits a {ClaimVCA} event.
+    /// @dev It emits a {ClaimVCA} event, and a {RedistributionReward} event if the redistribution is enabled.
+    ///
+    /// Notes:
+    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, it can either cause a
+    /// race condition among the recipients or rewards would be calculated as 0 if its too low depending on how low the
+    /// value is set.
+    /// - If the campaign creator does not sufficiently fund the campaign with the actual total allocations (or
+    /// `AGGREGATE_AMOUNT`), it can cause a race condition among the recipients.
+    /// - The rewards are transferred to the recipients at the time of claiming. If the campaign creator turns the
+    /// redistribution on after the vesting end time, the recipients who have already claimed the full amount would miss
+    /// on the rewards while subsequent recipients would get them.
     ///
     /// Requirements:
     /// - The current time must be greater than or equal to the campaign start time.
@@ -150,4 +187,18 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
     )
         external
         payable;
+
+    /// @notice Enable the redistribution of forgone tokens among recipients claiming after the vesting end time,
+    /// proportional to their allocation amount. Once enabled, it cannot be disabled.
+    ///
+    /// @dev Notes while calling this function:
+    /// - If the function is called after the vesting end time, the recipients who have already claimed the full amount
+    /// would miss on the rewards while subsequent recipients would get them.
+    /// - It is also recommended to fund the campaign with the actual total allocation in the Merkle tree (ideally
+    /// equivalent to `AGGREGATE_AMOUNT`) to avoid race conditions among the recipients.
+    ///
+    /// Requirements:
+    /// - `msg.sender` must be the admin.
+    /// - Redistribution must not be already enabled.
+    function enableRedistribution() external;
 }
--- a/airdrops/src/libraries/Errors.sol
+++ b/airdrops/src/libraries/Errors.sol
@@ -99,7 +99,7 @@ library Errors {
     /// @notice Thrown when the claim amount is zero.
     error SablierMerkleVCA_ClaimAmountZero(address recipient);
 
-    /// @notice Thrown when trying to switch to REDISTRIBUTE strategy when already using it.
+    /// @notice Thrown when trying to switch to redistribute strategy when already using it.
     error SablierMerkleVCA_RedistributionAlreadyEnabled();
 
     /// @notice Thrown when trying to calculate the rewards amount without redistribution enabled.
--- a/airdrops/src/libraries/Errors.sol
+++ b/airdrops/src/libraries/Errors.sol
@@ -30,6 +30,9 @@ library Errors {
                              SABLIER-FACTORY-MERKLE-VCA
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @notice Thrown when trying to create a Merkle VCA campaign with zero aggregate amount.
+    error SablierFactoryMerkleVCA_AggregateAmountZero();
+
     /// @notice Thrown if expiration time is within 1 week from the vesting end time.
     error SablierFactoryMerkleVCA_ExpirationTooEarly(uint40 vestingEndTime, uint40 expiration);
 
@@ -93,9 +96,15 @@ library Errors {
                                  SABLIER-MERKLE-VCA
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @notice Thrown when calculating the forgone amount with claim time less than the vesting start time.
-    error SablierMerkleVCA_VestingNotStarted(uint40 claimTime, uint40 vestingStartTime);
-
     /// @notice Thrown when the claim amount is zero.
     error SablierMerkleVCA_ClaimAmountZero(address recipient);
+
+    /// @notice Thrown when trying to switch to REDISTRIBUTE strategy when already using it.
+    error SablierMerkleVCA_RedistributionAlreadyEnabled();
+
+    /// @notice Thrown when trying to calculate the rewards amount without redistribution enabled.
+    error SablierMerkleVCA_RedistributionNotEnabled();
+
+    /// @notice Thrown when calculating the forgone amount with claim time less than the vesting start time.
+    error SablierMerkleVCA_VestingNotStarted(uint40 claimTime, uint40 vestingStartTime);
 }
--- a/airdrops/src/types/DataTypes.sol
+++ b/airdrops/src/types/DataTypes.sol
@@ -129,8 +129,13 @@ library MerkleLT {
 library MerkleVCA {
     /// @notice Struct encapsulating the constructor parameters of Merkle VCA campaigns.
     /// @dev The fields are arranged alphabetically.
+    /// @param aggregateAmount The total amount of ERC-20 tokens to be distributed to all recipients. If its value is
+    /// set lower than actual total allocations in the Merkle tree, it can either cause a race condition among the
+    /// recipients or rewards would be calculated as 0 if its too low. As a campaign creator, it is recommended to set
+    /// the value to the actual total allocations.
     /// @param campaignName The name of the campaign.
     /// @param campaignStartTime The start time of the campaign, as a Unix timestamp.
+    /// @param enableRedistribution Enable redistribution of forgone tokens at deployment.
     /// @param expiration The expiration of the campaign, as a Unix timestamp.
     /// @param initialAdmin The initial admin of the campaign.
     /// @param ipfsCID The content identifier for indexing the contract on IPFS. An empty value may break certain UI
@@ -142,8 +147,10 @@ library MerkleVCA {
     /// @param vestingEndTime Vesting end time, as a Unix timestamp.
     /// @param vestingStartTime Vesting start time, as a Unix timestamp.
     struct ConstructorParams {
+        uint128 aggregateAmount;
         string campaignName;
         uint40 campaignStartTime;
+        bool enableRedistribution;
         uint40 expiration;
         address initialAdmin;
         string ipfsCID;
--- a/airdrops/tests/fork/merkle-campaign/MerkleBase.t.sol
+++ b/airdrops/tests/fork/merkle-campaign/MerkleBase.t.sol
@@ -25,7 +25,7 @@ abstract contract MerkleBase_Fork_Test is Fork_Test {
         LeafData[] leavesData;
         LeafData leafToClaim;
         uint256 initialAdminBalance;
-        uint256 aggregateAmount;
+        uint128 aggregateAmount;
         uint128 clawbackAmount;
         address expectedMerkleCampaign;
         bytes32[] merkleProof;
--- a/airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
+++ b/airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
@@ -18,13 +18,25 @@ contract ClaimTo_MerkleVCA_Integration_Test is
     ClaimTo_Integration_Test,
     MerkleVCA_Integration_Shared_Test
 {
-    function setUp()
-        public
-        virtual
-        override(MerkleVCA_Integration_Shared_Test, ClaimTo_Integration_Test, Integration_Test)
-    {
+    function setUp() public override(MerkleVCA_Integration_Shared_Test, ClaimTo_Integration_Test, Integration_Test) {
         MerkleVCA_Integration_Shared_Test.setUp();
         ClaimTo_Integration_Test.setUp();
+
+        // Warp to campaign start time.
+        vm.warp({ newTimestamp: CAMPAIGN_START_TIME });
+
+        // Claim early for a user so that there is some forgone amount.
+        setMsgSender(users.unknownRecipient);
+        claimTo({
+            msgValue: AIRDROP_MIN_FEE_WEI,
+            index: getIndexInMerkleTree(users.unknownRecipient),
+            to: users.unknownRecipient,
+            amount: VCA_FULL_AMOUNT,
+            merkleProof: getMerkleProof(users.unknownRecipient)
+        });
+
+        // Set the recipient as the caller for this test.
+        setMsgSender(users.recipient);
     }
 
     function test_RevertWhen_VestingStartTimeInFuture() external whenMerkleProofValid {
@@ -32,6 +44,7 @@ contract ClaimTo_MerkleVCA_Integration_Test is
         MerkleVCA.ConstructorParams memory params = merkleVCAConstructorParams();
         params.vestingStartTime = getBlockTimestamp() + 1 seconds;
         merkleVCA = createMerkleVCA(params);
+        merkleBase = merkleVCA;
 
         // It should revert.
         vm.expectRevert(abi.encodeWithSelector(Errors.SablierMerkleVCA_ClaimAmountZero.selector, users.recipient));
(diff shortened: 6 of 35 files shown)


airdrops/src/SablierMerkleVCA.sol
```diff
@@ -75,17 +86,21 @@ contract SablierMerkleVCA is
         )
     {
         // Effect: set the immutable variables.
+        AGGREGATE_AMOUNT = params.aggregateAmount;
         UNLOCK_PERCENTAGE = params.unlockPercentage;
         VESTING_END_TIME = params.vestingEndTime;
         VESTING_START_TIME = params.vestingStartTime;
+
+        // Effect: enable redistribution if true.
+        if (params.enableRedistribution) isRedistributionEnabled = true;
```
andreivladbrg: we don't need the check, we can simply: `isRedistributionEnabled = params.enableRedistribution `

airdrops/src/types/DataTypes.sol
```diff
@@ -139,8 +144,10 @@ library MerkleVCA {
     /// @param vestingEndTime Vesting end time, as a Unix timestamp.
     /// @param vestingStartTime Vesting start time, as a Unix timestamp.
     struct ConstructorParams {
+        uint128 aggregateAmount;
```
andreivladbrg, line 150: why is the `aggregateAmount` declared as `uint128 `?

it should be `uint256`

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -34,12 +35,16 @@ contract SablierMerkleVCA is
...
 {
+    using SafeCast for uint256;
     using SafeERC20 for IERC20;
 
     /*//////////////////////////////////////////////////////////////////////////
                                   STATE VARIABLES
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @inheritdoc ISablierMerkleVCA
+    uint128 public immutable override AGGREGATE_AMOUNT;
```
andreivladbrg, line 46: `AGGREGATE_AMOUNT` can actually be greater than `uint128`

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -50,7 +55,13 @@ contract SablierMerkleVCA is
     uint40 public immutable override VESTING_START_TIME;
 
     /// @inheritdoc ISablierMerkleVCA
-    uint256 public override totalForgoneAmount;
+    bool public override isRedistributionEnabled;
+
+    /// @inheritdoc ISablierMerkleVCA
+    uint128 public override totalForgoneAmount;
```
andreivladbrg, line 61: why was it changed to `uin128` ?

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -50,7 +55,13 @@ contract SablierMerkleVCA is
...
+
+    /// @dev Tracks the full amount allocated to the recipients who claimed before the vesting end time.
+    uint128 private _fullAmountAllocatedToEarlyClaimers;
```
andreivladbrg, line 64: similar to `AGGREGATE_AMOUNT` - it should be `uint256`

airdrops/src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -58,17 +72,40 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
     /// @return The amount that would be forgone, denominated in the token's decimals.
     function calculateForgoneAmount(uint128 fullAmount, uint40 claimTime) external view returns (uint128);
 
+    /// @notice Calculates the redistribution rewards for a given full amount.
+    /// @dev Notes:
+    /// - Reverts if redistribution is not enabled.
+    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, this might return 0
+    /// rather than reverting.
+    /// @param fullAmount The amount of tokens that the redistribution rewards are to be calculated for.
+    function calculateRedistributionRewards(uint128 fullAmount) external view returns (uint128);
```
andreivladbrg, line 81: IMO it would be useful to add a view `calculateRewardAPY` function

airdrops/src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -150,4 +187,18 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
...
         payable;
+
+    /// @notice Enable the redistribution of forgone tokens among recipients claiming after the vesting end time,
+    /// proportional to their allocation amount. Once enabled, it cannot be disabled.
+    ///
+    /// @dev Notes while calling this function:
+    /// - If the function is called after the vesting end time, the recipients who have already claimed the full amount
+    /// would miss on the rewards while subsequent recipients would get them.
+    /// - It is also recommended to fund the campaign with the actual total allocation in the Merkle tree (ideally
+    /// equivalent to `AGGREGATE_AMOUNT`) to avoid race conditions among the recipients.
```
andreivladbrg, line 196: saying ``equivalent to `AGGREGATE_AMOUNT` ideally`` might incorrect information if the aggregate amount from state is set incorrectly

so, i'd keep it only with "actual total allocation in the Merkle tree"

airdrops/src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -58,17 +72,40 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
...
+    ///
+    /// Notes:
+    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, it can either cause a
+    /// race condition among the recipients or rewards would be calculated as 0 if its too low depending on how low the
+    /// value is set.
+    /// - If the campaign creator does not sufficiently fund the campaign with the actual total allocations (or
+    /// `AGGREGATE_AMOUNT`), it can cause a race condition among the recipients.
+    /// - The rewards are transferred to the recipients at the time of claiming. If the campaign creator turns the
+    /// redistribution on after the vesting end time, the recipients who have already claimed the full amount would miss
+    /// on the rewards while subsequent recipients would get them.
```
andreivladbrg, line 106: i believe this point is more relevant to the `enableRedistribution` function 
for the claimer, it’s simple: if the rewards are enabled, they get some extra funds, if not they don’t

airdrops/src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -58,17 +72,40 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
...
     ///
-    /// @dev It emits a {ClaimVCA} event.
+    /// @dev It emits a {ClaimVCA} event, and a {RedistributionReward} event if the redistribution is enabled.
+    ///
+    /// Notes:
+    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, it can either cause a
+    /// race condition among the recipients or rewards would be calculated as 0 if its too low depending on how low the
+    /// value is set.
+    /// - If the campaign creator does not sufficiently fund the campaign with the actual total allocations (or
+    /// `AGGREGATE_AMOUNT`), it can cause a race condition among the recipients.
```
andreivladbrg:
```diff
-   /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, it can either cause a
-   /// race condition among the recipients or rewards would be calculated as 0 if its too low depending on how low -- the
-   /// value is set.
-   /// - If the campaign creator does not sufficiently fund the campaign with the actual total allocations (or
-   /// `AGGREGATE_AMOUNT`), it can cause a race condition among the recipients.
+   /// - There can be a race condition among recipients if:                                                                                                                
+   ///   1. `AGGREGATE_AMOUNT` is set lower than the actual total allocations in the Merkle tree.                                                                          
+   ///   2. The campaign is not sufficiently funded with the actual total allocations.                                                                                     
```

airdrops/src/interfaces/ISablierMerkleVCA.sol
```diff
@@ -58,17 +72,40 @@ interface ISablierMerkleVCA is ISablierMerkleBase {
...
+    /// the `to` address, otherwise it transfers the full amount. If the redistribution is enabled, it calculates the
+    /// reward amount based on the total amount of tokens forgone by early claimers and transfers it to the recipients
+    /// claiming after the vesting end time.
     ///
-    /// @dev It emits a {ClaimVCA} event.
+    /// @dev It emits a {ClaimVCA} event, and a {RedistributionReward} event if the redistribution is enabled.
+    ///
+    /// Notes:
+    /// - If `AGGREGATE_AMOUNT` is set lower than actual total allocations in the Merkle tree, it can either cause a
+    /// race condition among the recipients or rewards would be calculated as 0 if its too low depending on how low the
```
andreivladbrg: "rewards would be calculated as 0 if its too low depending on how low the value is set"

is more relevant to the `calculateRedistributionRewards` function

airdrops/src/libraries/Errors.sol
```diff
@@ -93,9 +96,15 @@ library Errors {
                                  SABLIER-MERKLE-VCA
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @notice Thrown when calculating the forgone amount with claim time less than the vesting start time.
-    error SablierMerkleVCA_VestingNotStarted(uint40 claimTime, uint40 vestingStartTime);
-
     /// @notice Thrown when the claim amount is zero.
     error SablierMerkleVCA_ClaimAmountZero(address recipient);
+
+    /// @notice Thrown when trying to switch to REDISTRIBUTE strategy when already using it.
```
andreivladbrg: why caps on "REDISTRIBUTE" ? 
is the natspec outdated based on the initial implementation?

airdrops/src/SablierMerkleVCA.sol
andreivladbrg, line 46: instead of marking this as immutable wdys about allowing to edit it in case of mistakes - but only **before** the vesting end time or if its in within the grace period

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -212,21 +275,38 @@ contract SablierMerkleVCA is
...
+        uint128 transferAmount = claimAmount;
 
-        // Effect: update the total forgone amount.
+        // Effect: update the total forgone amount and the total amount claimed by early claimers.
         if (claimAmount < fullAmount) {
+            // Its safe to use unchecked because none of these values can overflow.
             unchecked {
                 forgoneAmount = fullAmount - claimAmount;
                 totalForgoneAmount += forgoneAmount;
+                _fullAmountAllocatedToEarlyClaimers += fullAmount;
```
andreivladbrg: these two (`totalForgoneAmount & _fullAmountAllocatedToEarlyClaimers`) now, can actually overflow:
- incorrect aggregate amount 
- lots of claims before end time with high amounts

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg, line 1: lets edit the last part of the `notice` natspec:

```diff
- /// @notice VCA stands for Variable Claim Amount, and is an airdrop model where the claim amount increases linearly
- /// until the airdrop period ends. Claiming early results in forgoing the remaining amount, whereas claiming after the
- /// period grants the full amount that was allocated.
+ /// @notice VCA stands for Variable Claim Amount, and is an airdrop model where the claim amount increases linearly                                                     
+ /// until the airdrop period ends. Claiming early results in forgoing the remaining amount, whereas claiming after the                                                  
+ /// period grants the full amount that was allocated, plus redistribution rewards if enabled.
```

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -197,6 +237,29 @@ contract SablierMerkleVCA is
         }
     }
 
+    /// @notice Calculates the redistribution rewards for a given full amount.
+    function _calculateRedistributionRewards(uint256 fullAmount) private view returns (uint128 rewards) {
```
andreivladbrg, line 241:
```suggestion
    function _calculateRedistributionRewards(uint128 fullAmount) private view returns (uint128 rewards) {
```

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -50,7 +55,13 @@ contract SablierMerkleVCA is
     uint40 public immutable override VESTING_START_TIME;
 
     /// @inheritdoc ISablierMerkleVCA
-    uint256 public override totalForgoneAmount;
+    bool public override isRedistributionEnabled;
```
andreivladbrg, line 58: wouldn't it make sense to have a redistribution end time in case this is `true`?

my rationale is that otherwise the user would be confused if the campaign creator suddenly clawed back funds after it was "finished". with a state variable (which can also be edited, but not only during the reward period), IMO it would be cleaner for both parties. also, `clawback` can be called only after this end time

airdrops/tests/fork/merkle-campaign/MerkleBase.t.sol
```diff
@@ -25,7 +25,7 @@ abstract contract MerkleBase_Fork_Test is Fork_Test {
         LeafData[] leavesData;
         LeafData leafToClaim;
         uint256 initialAdminBalance;
-        uint256 aggregateAmount;
+        uint128 aggregateAmount;
```
andreivladbrg, line 28: only the VCA's aggregate amount was updated to `uint128` - the other still use `uint256`

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
```diff
@@ -8,13 +8,20 @@ ClaimTo_MerkleVCA_Integration_Test
...
+      │  ├── it should update total forgone amount
       │  └── it should emit a {ClaimVCA} event
-      └── when vesting end time not in past
-         ├── it should transfer a portion of the amount to Eve
-         ├── it should update the claimed status
-         ├── it should update total forgone amount
-         └── it should emit a {ClaimVCA} event
+      └── when vesting end time not in future
+         ├── given redistribution not enabled
+         │  ├── it should transfer the full amount to Eve
```
andreivladbrg, line 18: why "Eve" specifically? :))

usually "Eve" is used for the reverting cases

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
```diff
@@ -8,13 +8,20 @@ ClaimTo_MerkleVCA_Integration_Test
...
+         │  ├── it should update the claimed status
```
andreivladbrg: why not "it should mark the index as Claimed"

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
andreivladbrg, line 1: i think it would be useful to have more cases in the `claimTo` tests, not only the "given redistribution not enabled/given redistribution enabled"

- claim _before_ vesting end time: (i) redistribution not enabled (ii) redistribution enabled
- claim _after_ vesting end time: (i) redistribution not enabled (ii) redistribution enabled

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
```diff
@@ -45,48 +58,101 @@ contract ClaimTo_MerkleVCA_Integration_Test is
...
+        // If current claim is made after vesting end time, the value of total forgone amount remains the same.
+        uint256 expectedTotalForgoneAmount = getBlockTimestamp() >= VESTING_END_TIME
+            ? initialTotalForgoneAmount
+            : initialTotalForgoneAmount + forgoneAmount;
+
         uint256 previousFeeAccrued = address(comptroller).balance;
 
+        // It should emit a {RedistributionReward} event for claims made after the vesting end time only if
+        // redistribution is enabled.
+        if (isRedistributionEnabled && getBlockTimestamp() >= VESTING_END_TIME) {
```
andreivladbrg:
```suggestion
        if (isRedistributionEnabled) {
```

we can remove the the time check

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
```diff
@@ -45,48 +58,101 @@ contract ClaimTo_MerkleVCA_Integration_Test is
...
         // Cast the {MerkleVCA} contract as {ISablierMerkleBase}.
         merkleBase = merkleVCA;
 
         uint256 index = getIndexInMerkleTree();
 
-        uint128 forgoneAmount = VCA_FULL_AMOUNT - claimAmount;
+        // If vesting end time is in the future, calculate the forgone amount.
+        uint128 forgoneAmount;
+        if (getBlockTimestamp() < VESTING_END_TIME) {
+            forgoneAmount = VCA_FULL_AMOUNT - expectedTransferAmount;
```
andreivladbrg: why not pass it as parameter to `_test_ClaimTo` instead of calculating it? (similar to `expectedTransferAmount`)

airdrops/tests/integration/fuzz/MerkleVCA.t.sol
```diff
@@ -226,13 +240,40 @@ contract MerkleVCA_Fuzz_Test is Shared_Fuzz_Test {
...
+        if (getBlockTimestamp() < merkleVCA.VESTING_END_TIME()) {
+            amountClaimedEarly += leafData.amount;
+        }
+
+        // Calculate the reward amount if vesting has ended and redistribution is enabled.
+        uint128 expectedRewardAmount;
+        if (getBlockTimestamp() >= merkleVCA.VESTING_END_TIME() && merkleVCA.isRedistributionEnabled()) {
+            uint128 amountExpectedToClaimAfterVestingEndTime = merkleVCA.AGGREGATE_AMOUNT() - amountClaimedEarly;
+            expectedRewardAmount =
+                uint128((uint256(leafData.amount) * totalForgoneAmount) / amountExpectedToClaimAfterVestingEndTime);
```
andreivladbrg, line 263: why not move these calculations in the `Base.calculateMerkleVCAAmounts()`?

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -173,9 +180,12 @@ abstract contract BaseHandler is Fuzzers, StdCheats, Utils {
     }
 
     /// @notice Helper function to deploy a Merkle Instant campaign with fuzzed leaves data and token.
-    /// @dev This will be called only once.
+    /// @dev Notes:
+    /// - This will be called only once.
+    /// - `enableRedistribution` is only relevant for Merkle VCA campaigns.
     function deployCampaign(
         address campaignCreator,
+        bool vcaRedistributionEnabled,
```
andreivladbrg: why not deploy campaigns with this boolean as true?

it would cover a wider range of cases

airdrops/tests/invariant/README.md
```diff
@@ -2,9 +2,16 @@
...
 
 #### For VCA campaign:
 
-1. For MerkleVCA campaigns, total forgone = $\sum$ claim requested - $\sum$ claimed
+1. total forgone = $\sum$ full amount requested - $\sum$ claimed
+2. If vesting has ended, total forgone amount should never change.
+3. If redistribution is enabled and aggregate amount is correctly set,
+   - Redistribution rewards for a fixed amount should never decrease.
+   - If vesting has ended, redistribution rewards for a fixed amount should never change.
+   - `calculateRedistributionRewardsPerToken` should never revert.
```
andreivladbrg: tbh i don't find this so useful 

its implicit when calling `merkleVCA.calculateRedistributionRewards()`

airdrops/tests/invariant/README.md
```diff
@@ -2,9 +2,16 @@
...
 2. `hasClaimed` should never change its value from `true` to `false`
 
 #### For VCA campaign:
 
-1. For MerkleVCA campaigns, total forgone = $\sum$ claim requested - $\sum$ claimed
+1. total forgone = $\sum$ full amount requested - $\sum$ claimed
+2. If vesting has ended, total forgone amount should never change.
+3. If redistribution is enabled and aggregate amount is correctly set,
+   - Redistribution rewards for a fixed amount should never decrease.
+   - If vesting has ended, redistribution rewards for a fixed amount should never change.
```
andreivladbrg, line 17: for "if vesting has ended", i think we can test that all these values remain constant:
- `totalForgoneAmount`
- `_fullAmountAllocatedToEarlyClaimers`
- `rewardAmount` per user

airdrops/tests/invariant/README.md
```diff
@@ -2,9 +2,16 @@
...
+   - `calculateRedistributionRewardsPerToken` should never revert.
+   - Rewards distributed should never exceed total forgone amount.
```
andreivladbrg, line 18: wdyt about: `sum(rewards) <= aggregateAmount - _fullAmountAllocatedToEarlyClaimers`

airdrops/tests/invariant/stores/Store.sol
```diff
@@ -16,16 +16,22 @@ contract Store is BaseUtils {
     /// @dev List of addresses to be excluded from being fuzzed as `msg.sender`.
     address[] internal excludedAddresses;
 
+    /// @dev Track redistribution rewards per 1e18 for VCA campaign before a claim is made.
+    uint256 public previousVcaRedistributionRewardsPer1e18;
```
andreivladbrg: i don't really understand this, 

whats up with the `Per1e18`?

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
```diff
@@ -30,8 +30,18 @@ contract MerkleVCAHandler is BaseHandler {
...
+
+        // Update redistribution rewards per 1e18 before claiming.
+        if (merkleVCA.isRedistributionEnabled()) {
+            store.updatePreviousVcaRedistributionRewardsPer1e18(
+                merkleVCA.calculateRedistributionRewards({ fullAmount: 1e18 })
+            );
+        }
+
+        // Update forgone amount for VCA campaign in store before calling the claim.
+        store.updatePreviousVcaTotalForgoneAmount(leafData.amount - claimAmount);
```
andreivladbrg: let's move these store updates after the claim call

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
```diff
@@ -30,8 +30,18 @@ contract MerkleVCAHandler is BaseHandler {
...
 
-        // Calculate the forgone amount.
-        uint128 forgoneAmount = leafData.amount - claimAmount;
+        // Get initial balance of recipient for rewards calculation.
+        uint256 initialRecipientBalance = campaignToken.balanceOf(leafData.recipient);
+
+        // Update redistribution rewards per 1e18 before claiming.
+        if (merkleVCA.isRedistributionEnabled()) {
+            store.updatePreviousVcaRedistributionRewardsPer1e18(
+                merkleVCA.calculateRedistributionRewards({ fullAmount: 1e18 })
```
andreivladbrg: this one can be grouped together with the one below

andreivladbrg (changes requested):
great work Shub, this new design is indeed much simpler. but with the aggregate amount being an input param, i think we should add some points to the protocol assumptions to prevent further discussions. something like: "in the case of merkle VCA campaigns, ensuring that the aggregate amount is `sum(leaves.amounts)` is the campaign creator’s responsibility"

leaving my review below

airdrops/src/SablierMerkleVCA.sol
smol-ninja: Right.

airdrops/src/types/DataTypes.sol
smol-ninja: Not in this case. In other campaigns, `aggregateAmount` is declared as `uint256` because its only emitted in the event. In case of VCA, `aggregateAmount` is a storage variable used in `_calculateRedistributionRewards` and therefore should follow the type we use for amounts i.e. `uint128`.

This is why I also set type for `AGGREGATE_AMOUNT` as `uint128` to avoid casting during calculations.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: In case of VCA, it should not be because of the assumption: The total supply of any ERC-20 token remains below `type(uint128).max`.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: Because its no more a returning value but is used in the calculation of rewards so it must be treated like other amounts.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: See my comments above. `uint128` is the best practice here imo.

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: I didn't add it because the value returned by `calculateRewardAPY` could be very different than actual apy due to rounding error.

In case of `calculateRedistributionRewards`, the rewards are calculated as `rewards = ((fullAmount * totalForgoneAmount) / fullAmountAllocatedToRemainingClaimers).toUint128();` which uses `mul` followed by `div`. If you were to calculate `calculateRewardAPY`, the value would depend on what you choose for `fullAmount`. 

We should let the UI do the APY calculation.

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: > saying ``equivalent to `AGGREGATE_AMOUNT` ideally`` might incorrect information if the aggregate amount from state is set incorrectly

Wdym? If the aggregate amount from state is set incorrectly, then redistribution wont work correctly.

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: I didn't get your point. lets say redistribution is enabled 5 days after vesting end time, the users who claim in the first 5 days wont receive the rewards and thats what the comment says.

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: Makes sense.

airdrops/src/libraries/Errors.sol
smol-ninja: enum, but we can go with lower cap.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: It can also be misused, no?

It can be used both ways: - a user incorrectly sets it, and then corrects it in good faith vs a user correctly sets it and then adjusts it in bad faith.

IMO if its value is incorrectly set, it can be detected within the grace period and the campaign creator can launch a new campaign. So having immutable + grace period allows creators to correct their mistakes.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: How can it overflow if both are within `uint128`?

airdrops/src/SablierMerkleVCA.sol
smol-ninja: I did that to avoid casting in `rewards = ((fullAmount * totalForgoneAmount) / fullAmountAllocatedToRemainingClaimers).toUint128();`

airdrops/src/SablierMerkleVCA.sol
smol-ninja: There is end time and then there is expiry. Between end time and expiry, campaign owner cannot clawback the funds. Isn't end time same as what you are proposing?

airdrops/tests/fork/merkle-campaign/MerkleBase.t.sol
smol-ninja: Answered above.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
smol-ninja: Yeah it should be recipient.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
smol-ninja: Can do. No preference from my side.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
smol-ninja: - claim before vesting end time: how will the test change here? The claim here does not depend on the value of `isRedistributionEnabled`. Also, fuzz tests already test these edge cases.
- claim after vesting end time: its already covered.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
smol-ninja: If we remove the time check then wont test fail because it would expect `RedistributionReward` event but it wont be emitted until end time has been passed?

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
smol-ninja: Can do if that makes code cleaner. `expectedTransferAmount` is an input parameter and `VCA_FULL_AMOUNT` is a constant so I thought it makes sense to compute `forgoneAmount` here.

airdrops/tests/integration/fuzz/MerkleVCA.t.sol
smol-ninja: Its only used here right?

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: but it wont cover a case where its not enabled at all. Wdyt?

And what new range of cases would it cover if we pass it as true?

airdrops/tests/invariant/README.md
smol-ninja: Sounds good. We can remove it.

airdrops/tests/invariant/README.md
smol-ninja: Yeah we can. However, how would you test `_fullAmountAllocatedToEarlyClaimers`?

airdrops/tests/invariant/README.md
smol-ninja: if "Rewards distributed should never exceed total forgone amount" is true, then what you are saying is also true. Isn't it?

airdrops/tests/invariant/stores/Store.sol
smol-ninja: Same reason as mentioned in https://github.com/sablier-labs/lockup/pull/1363#discussion_r2709808110

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: Why? 

By updating it prior to claim allows us to calculate the difference between the current value and the previous value (the value before the call).

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: you mean with `updatePreviousVcaTotalForgoneAmount`?

smol-ninja:
Rebased and resolved conflicts.

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: right, i forgot about this check:

```solidity
if (expiration < vestingEndTime + 1 weeks) {
    revert Errors.SablierFactoryMerkleVCA_ExpirationTooEarly(vestingEndTime, expiration);
}
```

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: good point  - let's keep the `uint128`

andreivladbrg: yes, it can be misused, but there is the trust assumption anyway

my counterargument, besides the good-faith scenario, is that we can also achieve a new feature by allowing the campaign creator to increase the redistribute amount for loyal recipients, in case they later decide they want to do it

> IMO if its value is incorrectly set, it can be detected within the grace period and the campaign creator can launch a new campaign

thats a good point, but wdyt think about the new feature?

airdrops/src/types/DataTypes.sol
andreivladbrg: based on the total supply assumption, lets keep the `uint128`

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: based on the total supply assumption, lets keep the `uint128`

airdrops/src/SablierMerkleVCA.sol
smol-ninja: > campaign creator to increase the redistribute amount for loyal recipients

Then its not "redistribution". The original idea was to distribute the forgone amount only. What you are suggesting is different kind of rewards (potentially a separate airdrop).

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: > I didn't add it because the value returned by calculateRewardAPY could be very different than actual apy > > due to rounding error.
> ...
> We should let the UI do the APY calculation.

we can add explanatory comments about its precision (similar to flow’s `depletionTime`). and since it’s just for informational purposes, that’s not an issue IMO

isn't the UI going to face the same rounding problems?

---
> If you were to calculate `calculateRewardAPY`, the value would depend on what you choose for `fullAmount`.

that’s not true. the only things that matter are: `totalForgoneAmount` and `fullAmountAllocatedToRemainingClaimers`:

```solidity
fullAmountAllocatedToRemainingClaimers = AGGREGATE_AMOUNT - _fullAmountAllocatedToEarlyClaimers
// reward pct
return ud(totalForgoneAmount).div(ud(fullAmountAllocatedToRemainingClaimers));
```

for example:

- alice: 500 tokens
- bob: 1000 tokens
- eve: 1500 tokens
- AGGREGATE_AMOUNT = 3000

alice claims at 50% vesting:

```
claimAmount   = 500 × 50% = 250 tokens  
forgoneAmount = 500 - 250 = 250 tokens
```

state after alice’s claim:

```
totalForgoneAmount                  = 250  
_fullAmountAllocatedToEarlyClaimers = 500  
remainingAllocation                 = 3000 - 500 = 2500
```

---

apy calculation:

```
apy = totalForgoneAmount / remainingAllocation  
    = 250 / 2500  
    = 0.10 (10%)
```

---

bob and eve claim after vesting ends:

| recipient | full amount | reward (10%) | total received |
|-----------|-------------|--------------|----------------|
| bob       | 1000        | 100          | 1100           |
| eve       | 1500        | 150          | 1650           |

airdrops/src/libraries/Errors.sol
andreivladbrg: what enum? there isn't one

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: if we keep the aggregateAmount as `uint128`, how can it overflow then if we have `mul` and then `div`?

the max value one can get is his `fullAmount * 2`

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: Because `.div` multiplies the value with `1e18`,

```solidity
ud(totalForgoneAmount).div(ud(fullAmountAllocatedToRemainingClaimers));
```

Its technically same as `calculateRedistributionRewards({ fullAmount: 1e18 })`. Because there are extra 0s, I wonder how this value is useful to any user. A user will largely be interested into knowing how much he would receive rather than seeing this value, divide it by 1e18 and multiply by his full amount.

Also, calling it APY would be inappropriate as APY stands for Annual Percentage Yield. Here, the rewards are fixed.

Your thoughts?

airdrops/src/libraries/Errors.sol
smol-ninja: Right. I confused with the original discussion where an enum was proposed. I agree it should be all lower cap.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: In case of `fullAmount * totalForgoneAmount`, if both are `uint128`, the product can overflow `uint128` and therefore, the correct method would be `uint256(fullAmount) * totalForgoneAmount` so its within `uint256` and then divide it by `fullAmountAllocatedToRemainingClaimers`.

By having `uint256 fullAmount` in the function argument, it automatically cast `fullAmount` from `uint128` into `uint256`.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
andreivladbrg: > claim before vesting end time: how will the test change here? The claim here does not depend on the value of isRedistributionEnabled. Also, fuzz tests already test these edge cases.

my idea is to verify that no matter if its enabled or not - the states are updated in the same manner
but, lets do it only in fuzz tests

> * claim after vesting end time: its already covered.

correct, overlooked it

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: I have one more argument. Lets say the `totalForgoneAmount` is so small compared to `fullAmountAllocatedToRemainingClaimers` that `ud(totalForgoneAmount).div(ud(fullAmountAllocatedToRemainingClaimers))` returns 0 (because it only multiplies by `1e18`). But a user with `1000e18` full amount can receive non-zero tokens. In that case, the APY function would return 0 but the actual reward would not be zero for users > `1000e18`. So this is also possible in theory.

andreivladbrg: > Its technically same as calculateRedistributionRewards({ fullAmount: 1e18 }). Because there are extra 0s, I wonder how this value is useful to any user. A user will largely be interested into knowing how much he would receive rather than seeing this value, divide it by 1e18 and multiply by his full amount.

as an absolute value, it would be true, but the difference here is that `calculateRedistributionRewards({ fullAmount: 1e18 })` returns an amount, while this new function returns a **percentage** of rewards scaled (normalized) to 18 decimals.  
1e18 = 100%, 0.1e18 = 10%, 0.01e18 = 1%, and so on

you are right about the name. we can call it `redistributeRewardPct` or any other name

--- 

>Lets say the totalForgoneAmount is so small compared to fullAmountAllocatedToRemainingClaimers that ud(totalForgoneAmount).div(ud(fullAmountAllocatedToRemainingClaimers)) returns 0 

this is theoretically true, but it’s also true for the `calculateRedistributionRewards` function and what a certain user would get. 
if `totalForgoneAmount` is small and there are different recipients, with some having a large amount airdropped and some with a smaller portion, the large-amount recipients can get a reward > 0 while the small ones get 0 redistributed: basically when `fullAmount * totalForgoneAmount < fullAmountAllocatedToRemainingClaimers`

which is more problematic actually than just a view (informational) function

smol-ninja: > returns an amount, while this new function returns a percentage of rewards scaled (normalized) to 18 decimals

How are these two different? Both functions would return `0.1e18`. So, from the user point of view, one is just another with `fullAmount = 1e18`.  The advantage with `calculateRedistributionRewards` is that user can check the exact amount they may receive. 

> but it’s also true for the calculateRedistributionRewards function and what a certain user would get

This is true. But they will see what they will receive with their full amount. If a user sees 0 with his full amount, that means he will receive nothing. On the other hand, because `redistributeRewardPct` returns a fixed value, it does not reflect the fact that some user will receive rewards while others wont.

> which is more problematic actually than just a view (informational) function

Isn't it the reality?

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: the tree might be incorrect and include higher amounts  
i believe that, from the perspective of the contract’s security (even tho we have the relevant checks in the UI & server) it’s safer not to use `unchecked` for them

airdrops/tests/integration/fuzz/MerkleVCA.t.sol
andreivladbrg: i thought its used in fork tests as well

but there we have this:
```solidity
expectedRewardAmount = merkleVCA.calculateRedistributionRewards({ fullAmount: vars.leafToClaim.amount });
``` 

whats the reason for using internal logic in the fuzz tests and there calling the campaign function?

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: > And what new range of cases would it cover if we pass it as true?

more claims with redistribute amount > 0

airdrops/tests/invariant/README.md
andreivladbrg: i changed my mind about this because these values are implicitly constant if “ redistribution rewards for a fixed amount should never change”

airdrops/src/SablierMerkleVCA.sol
smol-ninja: Yes thats a good point. It shouldn't have `unchecked`

airdrops/tests/integration/fuzz/MerkleVCA.t.sol
smol-ninja: Fork tests should also use internal logic as well. I think I made a mistake here. `expectedRewardAmount` should be calculated locally and then compared against the contract value.

So, feel free to move these into `Utils`.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: And what about the case when `vcaRedistributionEnabled` is not enabled? Are you suggesting to have a separate function for that?

airdrops/tests/invariant/README.md
smol-ninja: Yeah

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: right, closing

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: my ultimate problem is that within an invariant test run, the tests would be run against a single vca campaign, which is either with `vcaRedistributionEnabled` true or false, and IMO we should have both

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
andreivladbrg: i am not sure i understand your point

`leafData.amount - claimAmount` is the same before and after the `claimTo` call

my idea with the comment was to group all store updates next to each other

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: OK lets have both. I am up for it.

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
andreivladbrg: i meant with this:
```solidity
if (merkleVCA.isRedistributionEnabled()) {
    // Get final balance of recipient for rewards calculation.
    uint256 finalRecipientBalance = campaignToken.balanceOf(leafData.recipient);

    // Calculate the rewards transferred to the recipient and update in store.
    uint256 rewardsTransferred = finalRecipientBalance - initialRecipientBalance - claimAmount;
    store.updateTotalRewardsDistributed(rewardsTransferred);
}
``` 

but this is a mistake as `merkleVCA.calculateRedistributionRewards({ fullAmount: 1e18 })` would return a different value

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
smol-ninja: Yeah you are right. I misunderstood you.

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: well, i can argue that redistribution just means giving an incentive to the loyal recipients that didn’t claim
no matter whether that rewarded amount is the initial aggregate amount of the merkle tree or greater

in the current design, a campaign creator can actually set up a higher aggregate amount from the start so that the redistributed reward amount is > 0 even tho there is no early claimer - so i don't think thats a strong point

smol-ninja: > a campaign creator can actually set up a higher aggregate amount 

In that case, `fullAmountAllocatedToRemainingClaimers` will be higher and rewards will be less for each user. 

```solidity
fullAmountAllocatedToRemainingClaimers = AGGREGATE_AMOUNT - _fullAmountAllocatedToEarlyClaimers;
rewards = ((fullAmount * totalForgoneAmount) / fullAmountAllocatedToRemainingClaimers).toUint128();
```

So if campaign creator sets aggregate amount more than the true aggregate amount, it penalize each user's redistribution rewards.

If a campaign creator wants to increase rewards, then he must set a low value for aggregate amount, and then each user will receive more amount but this will become a first come first serve case. 

You see how complicated it can become.

airdrops/tests/integration/fuzz/MerkleVCA.t.sol
andreivladbrg: actually, it ok to keep the fork tests as it is now because there are no prior claims, thus the forgone amount is 0

andreivladbrg: resolving this

airdrops/tests/invariant/README.md
andreivladbrg: yeah, thats right

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: yeah, i got that, but i my point was that its not needed to have it both functions `claimTo` and `enableRedistribution`

but i am now ok to keep it both places

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: good point

yeah, i agree, lets not overcomplicate things

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: > How are these two different? Both functions would return `0.1e18`. So, from the user point of view, one is just another with `fullAmount = 1e18`. The advantage with `calculateRedistributionRewards` is that user can check the exact amount they may receive

because they represent different things - one is a pct and the other an amount

it’s like saying that, in the case of an LL stream, at the midpoint of the stream, the `withdrawable` and `refundable` functions return the same value, and ask why, when they represent different things

> This is true. But they will see what they will receive with their full amount. If a user sees 0 with his full amount, that means he will receive nothing

but wouldn’t that be problematic (compared to a view function) if users see that redistribution is enabled on the campaign, consider themselves "loyal", and then see that they receive 0 surplus tokens?

if you’re worried about the precision of the pct function, we can scale it to `1e27`, which would address your concern for small values, given that tokens have `1e18` decimals or less

and again, we can add notes about its precision, like in `flow.depletionTimeOf()`

> Isn't it the reality?

i am not sure i understand what you mean

airdrops/tests/invariant/handlers/MerkleVCAHandler.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/tests/invariant/README.md
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: > but wouldn’t that be problematic (compared to a view function) if users see that redistribution is enabled on the campaign, consider themselves "loyal", and then see that they receive 0 surplus tokens?

if they see that they will receive 0 token, they will actually receive 0 tokens when they claim. isn't it? So how is that problematic? They receive what they see.

I am not worried about the precision. What I am worried is the following:

1. In case of `pct function`, they may see some non-zero value and assume they will receive something but when they claim, they receive nothing.
2. In case of `calculateRedistributionRewards`, they see what they will actually receive.

This is the whole point of my arguments. Let me know your thoughts.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.tree
andreivladbrg: apparently in case of `claimTo` `users.eve` is used , i kept it like this

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/src/libraries/Errors.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: 45d212f751e45e43095674e57a31c790d6dda2f0

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: i’m going to give up on the idea. i just took a look at aave’s contracts and didn’t find an "apy" function at the contract level either

sorry for the long thread

airdrops/tests/invariant/handlers/BaseHandler.sol
andreivladbrg: 330ab3419aa58ec087af5f80aecfd9a1c7a9db07

airdrops/src/interfaces/ISablierMerkleVCA.sol
andreivladbrg: nvm, the total allocation in the Merkle tree can also be configured incorrectly

andreivladbrg (approved):
if the last commit pushed looks good to you, feel free to merge it

airdrops/src/interfaces/ISablierMerkleVCA.sol
smol-ninja: No worries. It was a good discussion.

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -54,7 +54,7 @@ abstract contract BaseHandler is Fuzzers, StdCheats, Utils {
     /// @dev Assume common deployment parameters.
     modifier assumeDeployParams(address campaignCreator, LeafData[] memory rawLeavesData) {
         // Deploy campaign only if it hasn't been deployed yet.
```
smol-ninja: This should be updated.

airdrops/tests/invariant/handlers/BaseHandler.sol
```diff
@@ -109,6 +109,11 @@ abstract contract BaseHandler is Fuzzers, StdCheats, Utils {
     /// @dev An internal deploy campaign function that must be overridden by the handler contract.
     function _deployCampaign(address campaignCreator, bytes32 merkleRoot) internal virtual returns (address);
 
+    /// @dev Overridden by VCA handler to allow two campaigns.
```
smol-ninja: Here it should say

"Returns maximum number of campaigns to deploy during invariant tests. It may be overridden by the handler contract."

The note about VCA should be in the VCA Handler.

smol-ninja (commented):
LGTM. Just few comments.

airdrops/tests/invariant/handlers/BaseHandler.sol
smol-ninja: I am pushing the commit

PR #1366, opened 2025-12-31
feat(lockup,airdrops): add unlock granularity to Lockup Linear streams

### LINKED ISSUE #1362: Add unlock granularity to Lockup Linear streams
### Task

As proposed in https://github.com/sablier-labs/command-center/discussions/228, add a new unlock granularity storage variable that will track the smallest step in time between two consecutive token unlocks.

As of today, the unlock granularity is 1 second.

### Pseudo Implementation

- [x] `SablierLockupState.sol`

```solidity
/// @dev Unlock granularity mapped by stream ID, used in LL streams. The value cannot be zero.
mapping(uint256 streamId => uint64 granularity) internal _granularities;
```

- [x] `LockupMath.sol`

```solidity
UD60x18 elapsedTimeInGranularPercentage;
if (elapsedTimeGranular > 0) {
  // Step 1: Calculate elapstedTimeOnGranularAxis = elapsedTime / granularity <---- we need the floor value
  // Step 2: Calculate totalPeriodOnGranularAxis = streamableRange / granularity <----- we need the floor value
  // Step 3: Calculate elapsedTimePercentageOnGranularAxis = elapstedTimeOnGranularAxis.div(totalPeriodOnGranularAxis)
} else {
    elapsedTimePercentageOnGranularAxis = elapsedTimePercentage;
}

uint128 streamedAmount = unlockAmountsSum + (elapsedTimePercentageOnGranularAxis.mul(streamableAmount)).intoUint128();
```

smol-ninja:
### Changelog

- Closes https://github.com/sablier-labs/lockup/issues/1362
- Adds changes from https://github.com/sablier-labs/lockup/commit/4068680944d596cfd143f6c90f3cc98ad5e5ade0, https://github.com/sablier-labs/lockup/commit/2539c05c7e792073b0b14d2b5cb769c448561c0f, https://github.com/sablier-labs/lockup/commit/ff46bb947571570146089c7eb610b6637cdbb9c3, https://github.com/sablier-labs/lockup/commit/f8f2f7929d8fff600034782da47c8a8fe92d29de and https://github.com/sablier-labs/flow/commit/54bdeb7301adf603e27543df71fe9330acd0944f

### DIFF
--- a/airdrops/src/SablierMerkleLL.sol
+++ b/airdrops/src/SablierMerkleLL.sol
@@ -88,10 +88,10 @@ contract SablierMerkleLL is
         // Effect: set the immutable variables.
         VESTING_CLIFF_DURATION = campaignParams.cliffDuration;
         VESTING_CLIFF_UNLOCK_PERCENTAGE = campaignParams.cliffUnlockPercentage;
+        VESTING_GRANULARITY = campaignParams.granularity;
         VESTING_START_TIME = campaignParams.vestingStartTime;
         VESTING_START_UNLOCK_PERCENTAGE = campaignParams.startUnlockPercentage;
         VESTING_TOTAL_DURATION = campaignParams.totalDuration;
-        VESTING_GRANULARITY = campaignParams.granularity;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
--- a/airdrops/src/SablierMerkleLL.sol
+++ b/airdrops/src/SablierMerkleLL.sol
@@ -55,6 +55,9 @@ contract SablierMerkleLL is
     /// @inheritdoc ISablierMerkleLL
     uint40 public immutable override VESTING_TOTAL_DURATION;
 
+    /// @inheritdoc ISablierMerkleLL
+    uint40 public immutable override VESTING_UNLOCK_GRANULARITY;
+
     /*//////////////////////////////////////////////////////////////////////////
                                     CONSTRUCTOR
     //////////////////////////////////////////////////////////////////////////*/
@@ -88,6 +91,7 @@ contract SablierMerkleLL is
         VESTING_START_TIME = campaignParams.vestingStartTime;
         VESTING_START_UNLOCK_PERCENTAGE = campaignParams.startUnlockPercentage;
         VESTING_TOTAL_DURATION = campaignParams.totalDuration;
+        VESTING_UNLOCK_GRANULARITY = campaignParams.unlockGranularity;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -207,7 +211,8 @@ contract SablierMerkleLL is
                     shape: streamShape
                 }),
                 unlockAmounts,
-                cliffTime
+                cliffTime,
+                VESTING_UNLOCK_GRANULARITY
             );
 
             // Effect: push the stream ID into the claimed streams array.
--- a/airdrops/src/SablierMerkleLL.sol
+++ b/airdrops/src/SablierMerkleLL.sol
@@ -47,6 +47,9 @@ contract SablierMerkleLL is
     UD60x18 public immutable override VESTING_CLIFF_UNLOCK_PERCENTAGE;
 
     /// @inheritdoc ISablierMerkleLL
+    uint40 public immutable override VESTING_GRANULARITY;
+
+    /// @inheritdoc ISablierMerkleLL
     uint40 public immutable override VESTING_START_TIME;
 
     /// @inheritdoc ISablierMerkleLL
@@ -54,9 +57,6 @@ contract SablierMerkleLL is
 
     /// @inheritdoc ISablierMerkleLL
     uint40 public immutable override VESTING_TOTAL_DURATION;
-
-    /// @inheritdoc ISablierMerkleLL
-    uint40 public immutable override VESTING_UNLOCK_GRANULARITY;
 
     /*//////////////////////////////////////////////////////////////////////////
                                     CONSTRUCTOR
@@ -91,7 +91,7 @@ contract SablierMerkleLL is
         VESTING_START_TIME = campaignParams.vestingStartTime;
         VESTING_START_UNLOCK_PERCENTAGE = campaignParams.startUnlockPercentage;
         VESTING_TOTAL_DURATION = campaignParams.totalDuration;
-        VESTING_UNLOCK_GRANULARITY = campaignParams.unlockGranularity;
+        VESTING_GRANULARITY = campaignParams.granularity;
     }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -211,7 +211,7 @@ contract SablierMerkleLL is
                     shape: streamShape
                 }),
                 unlockAmounts,
-                VESTING_UNLOCK_GRANULARITY,
+                VESTING_GRANULARITY,
                 cliffTime
             );
 
--- a/airdrops/src/SablierMerkleLL.sol
+++ b/airdrops/src/SablierMerkleLL.sol
@@ -211,8 +211,8 @@ contract SablierMerkleLL is
                     shape: streamShape
                 }),
                 unlockAmounts,
-                cliffTime,
-                VESTING_UNLOCK_GRANULARITY
+                VESTING_UNLOCK_GRANULARITY,
+                cliffTime
             );
 
             // Effect: push the stream ID into the claimed streams array.
--- a/lockup/scripts/solidity/Init.s.sol
+++ b/lockup/scripts/solidity/Init.s.sol
@@ -56,7 +56,11 @@ contract Init is BaseScript {
                     shape: "Cliff Linear"
                 }),
                 LockupLinear.UnlockAmounts({ start: 0, cliff: 0 }),
-                LockupLinear.Durations({ cliff: cliffDurations[i], total: totalDurations[i] })
+                LockupLinear.Durations({
+                    cliff: cliffDurations[i],
+                    total: totalDurations[i],
+                    unlockGranularity: 1 seconds
+                })
             );
         }
 
--- a/lockup/scripts/solidity/Init.s.sol
+++ b/lockup/scripts/solidity/Init.s.sol
@@ -45,8 +45,8 @@ contract Init is BaseScript {
         uint40[] memory totalDurations =
             Solarray.uint40s(1 seconds, 1 hours, 24 hours, 1 weeks, 4 weeks, 12 weeks, 48 weeks);
         for (uint256 i = 0; i < totalDurations.length; ++i) {
-            lockup.createWithDurationsLL(
-                Lockup.CreateWithDurations({
+            lockup.createWithDurationsLL({
+                params: Lockup.CreateWithDurations({
                     sender: sender,
                     recipient: recipient,
                     depositAmount: depositAmounts[i],
@@ -55,10 +55,10 @@ contract Init is BaseScript {
                     transferable: true,
                     shape: "Cliff Linear"
                 }),
-                LockupLinear.UnlockAmounts({ start: 0, cliff: 0 }),
-                1 seconds,
-                LockupLinear.Durations({ cliff: cliffDurations[i], total: totalDurations[i] })
-            );
+                unlockAmounts: LockupLinear.UnlockAmounts({ start: 0, cliff: 0 }),
+                granularity: 1 seconds,
+                durations: LockupLinear.Durations({ cliff: cliffDurations[i], total: totalDurations[i] })
+            });
         }
 
         // Renounce the 5th stream.
--- a/lockup/scripts/solidity/Init.s.sol
+++ b/lockup/scripts/solidity/Init.s.sol
@@ -56,11 +56,8 @@ contract Init is BaseScript {
                     shape: "Cliff Linear"
                 }),
                 LockupLinear.UnlockAmounts({ start: 0, cliff: 0 }),
-                LockupLinear.Durations({
-                    cliff: cliffDurations[i],
-                    total: totalDurations[i],
-                    unlockGranularity: 1 seconds
-                })
+                1 seconds,
+                LockupLinear.Durations({ cliff: cliffDurations[i], total: totalDurations[i] })
             );
         }
 
--- a/lockup/src/abstracts/SablierLockupLinear.sol
+++ b/lockup/src/abstracts/SablierLockupLinear.sol
@@ -169,8 +169,8 @@ abstract contract SablierLockupLinear is
                 shape: shape
             }),
             cliffTime: cliffTime,
-            unlockAmounts: unlockAmounts,
-            granularity: granularity
+            granularity: granularity,
+            unlockAmounts: unlockAmounts
         });
     }
 }
--- a/lockup/src/abstracts/SablierLockupLinear.sol
+++ b/lockup/src/abstracts/SablierLockupLinear.sol
@@ -25,6 +25,7 @@ abstract contract SablierLockupLinear is
     function createWithDurationsLL(
         Lockup.CreateWithDurations calldata params,
         LockupLinear.UnlockAmounts calldata unlockAmounts,
+        uint40 unlockGranularity,
         LockupLinear.Durations calldata durations
     )
         external
@@ -56,7 +57,7 @@ abstract contract SablierLockupLinear is
             token: params.token,
             transferable: params.transferable,
             unlockAmounts: unlockAmounts,
-            unlockGranularity: durations.unlockGranularity
+            unlockGranularity: unlockGranularity
         });
     }
 
@@ -64,8 +65,8 @@ abstract contract SablierLockupLinear is
     function createWithTimestampsLL(
         Lockup.CreateWithTimestamps calldata params,
         LockupLinear.UnlockAmounts calldata unlockAmounts,
-        uint40 cliffTime,
-        uint40 unlockGranularity
+        uint40 unlockGranularity,
+        uint40 cliffTime
     )
         external
         payable
--- a/lockup/src/abstracts/SablierLockupLinear.sol
+++ b/lockup/src/abstracts/SablierLockupLinear.sol
@@ -25,7 +25,7 @@ abstract contract SablierLockupLinear is
     function createWithDurationsLL(
         Lockup.CreateWithDurations calldata params,
         LockupLinear.UnlockAmounts calldata unlockAmounts,
-        uint40 unlockGranularity,
+        uint40 granularity,
         LockupLinear.Durations calldata durations
     )
         external
@@ -57,7 +57,7 @@ abstract contract SablierLockupLinear is
             token: params.token,
             transferable: params.transferable,
             unlockAmounts: unlockAmounts,
-            unlockGranularity: unlockGranularity
+            granularity: granularity
         });
     }
 
@@ -65,7 +65,7 @@ abstract contract SablierLockupLinear is
     function createWithTimestampsLL(
         Lockup.CreateWithTimestamps calldata params,
         LockupLinear.UnlockAmounts calldata unlockAmounts,
-        uint40 unlockGranularity,
+        uint40 granularity,
         uint40 cliffTime
     )
         external
@@ -86,7 +86,7 @@ abstract contract SablierLockupLinear is
             token: params.token,
             transferable: params.transferable,
             unlockAmounts: unlockAmounts,
-            unlockGranularity: unlockGranularity
+            granularity: granularity
         });
     }
 
@@ -106,40 +106,40 @@ abstract contract SablierLockupLinear is
         IERC20 token,
         bool transferable,
         LockupLinear.UnlockAmounts memory unlockAmounts,
-        uint40 unlockGranularity
+        uint40 granularity
     )
         private
         returns (uint256 streamId)
     {
-        // If sentinel value of zero is used for unlock granularity, change it to one second.
-        if (unlockGranularity == 0) {
-            unlockGranularity = 1;
+        // If sentinel value of zero is used for granularity, change it to one second.
+        if (granularity == 0) {
+            granularity = 1;
         }
 
         // Check: validate the user-provided parameters.
         Helpers.checkCreateLL({
             cliffTime: cliffTime,
             depositAmount: depositAmount,
+            granularity: granularity,
             nativeToken: nativeToken,
             sender: sender,
             shape: shape,
             timestamps: timestamps,
             token: address(token),
-            unlockAmounts: unlockAmounts,
-            unlockGranularity: unlockGranularity
+            unlockAmounts: unlockAmounts
         });
 
         // Load the stream ID in a variable.
         streamId = nextStreamId;
 
-        // Effect: set the start and cliff unlock amounts.
-        _unlockAmounts[streamId] = unlockAmounts;
-
         // Effect: update cliff time.
         _cliffs[streamId] = cliffTime;
 
-        // Effect: set the unlock granularity.
-        _unlockGranularities[streamId] = unlockGranularity;
+        // Effect: set the granularity.
+        _granularities[streamId] = granularity;
+
+        // Effect: set the start and cliff unlock amounts.
+        _unlockAmounts[streamId] = unlockAmounts;
 
         // Effect: create the stream, mint the NFT and transfer the deposit amount.
         _create({
@@ -170,7 +170,7 @@ abstract contract SablierLockupLinear is
             }),
             cliffTime: cliffTime,
             unlockAmounts: unlockAmounts,
-            unlockGranularity: unlockGranularity
+            granularity: granularity
         });
     }
 }
--- a/lockup/src/abstracts/SablierLockupLinear.sol
+++ b/lockup/src/abstracts/SablierLockupLinear.sol
@@ -55,7 +55,8 @@ abstract contract SablierLockupLinear is
             timestamps: timestamps,
             token: params.token,
             transferable: params.transferable,
-            unlockAmounts: unlockAmounts
+            unlockAmounts: unlockAmounts,
+            unlockGranularity: durations.unlockGranularity
         });
     }
 
@@ -63,7 +64,8 @@ abstract contract SablierLockupLinear is
     function createWithTimestampsLL(
         Lockup.CreateWithTimestamps calldata params,
         LockupLinear.UnlockAmounts calldata unlockAmounts,
-        uint40 cliffTime
+        uint40 cliffTime,
+        uint40 unlockGranularity
     )
         external
         payable
@@ -82,7 +84,8 @@ abstract contract SablierLockupLinear is
             timestamps: params.timestamps,
             token: params.token,
             transferable: params.transferable,
-            unlockAmounts: unlockAmounts
+            unlockAmounts: unlockAmounts,
+            unlockGranularity: unlockGranularity
         });
     }
 
@@ -101,21 +104,28 @@ abstract contract SablierLockupLinear is
         Lockup.Timestamps memory timestamps,
         IERC20 token,
         bool transferable,
-        LockupLinear.UnlockAmounts memory unlockAmounts
+        LockupLinear.UnlockAmounts memory unlockAmounts,
+        uint40 unlockGranularity
     )
         private
         returns (uint256 streamId)
     {
-        // Check: validate the user-provided parameters and cliff time.
+        // If sentinel value of zero is used for unlock granularity, change it to one second.
+        if (unlockGranularity == 0) {
+            unlockGranularity = 1;
+        }
+
+        // Check: validate the user-provided parameters.
         Helpers.checkCreateLL({
-            sender: sender,
-            timestamps: timestamps,
             cliffTime: cliffTime,
             depositAmount: depositAmount,
-            unlockAmounts: unlockAmounts,
-            token: address(token),
             nativeToken: nativeToken,
-            shape: shape
+            sender: sender,
+            shape: shape,
+            timestamps: timestamps,
+            token: address(token),
+            unlockAmounts: unlockAmounts,
+            unlockGranularity: unlockGranularity
         });
 
         // Load the stream ID in a variable.
@@ -126,6 +136,9 @@ abstract contract SablierLockupLinear is
 
         // Effect: update cliff time.
         _cliffs[streamId] = cliffTime;
+
+        // Effect: set the unlock granularity.
+        _unlockGranularities[streamId] = unlockGranularity;
 
         // Effect: create the stream, mint the NFT and transfer the deposit amount.
         _create({
@@ -155,7 +168,8 @@ abstract contract SablierLockupLinear is
                 shape: shape
             }),
             cliffTime: cliffTime,
-            unlockAmounts: unlockAmounts
+            unlockAmounts: unlockAmounts,
+            unlockGranularity: unlockGranularity
         });
     }
 }
--- a/lockup/src/abstracts/SablierLockupState.sol
+++ b/lockup/src/abstracts/SablierLockupState.sol
@@ -56,8 +56,8 @@ abstract contract SablierLockupState is ISablierLockupState {
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Checks that actual model and expected model are equal.
-    modifier notEqual(Lockup.Model actualModel, Lockup.Model expectedModel) {
-        _notEqual(actualModel, expectedModel);
+    modifier modelCheck(Lockup.Model actualModel, Lockup.Model expectedModel) {
+        _modelCheck(actualModel, expectedModel);
         _;
     }
 
@@ -90,7 +90,7 @@ abstract contract SablierLockupState is ISablierLockupState {
         view
         override
         notNull(streamId)
-        notEqual(_streams[streamId].lockupModel, Lockup.Model.LOCKUP_LINEAR)
+        modelCheck(_streams[streamId].lockupModel, Lockup.Model.LOCKUP_LINEAR)
         returns (uint40 cliffTime)
     {
         cliffTime = _cliffs[streamId];
@@ -140,7 +140,7 @@ abstract contract SablierLockupState is ISablierLockupState {
         view
         override
         notNull(streamId)
-        notEqual(_streams[streamId].lockupModel, Lockup.Model.LOCKUP_DYNAMIC)
+        modelCheck(_streams[streamId].lockupModel, Lockup.Model.LOCKUP_DYNAMIC)
         returns (LockupDynamic.Segment[] memory segments)
     {
         segments = _segments[streamId];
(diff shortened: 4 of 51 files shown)


lockup/src/abstracts/SablierLockupLinear.sol
```diff
@@ -101,21 +104,28 @@ abstract contract SablierLockupLinear is
...
+        LockupLinear.UnlockAmounts memory unlockAmounts,
+        uint40 unlockGranularity
     )
         private
         returns (uint256 streamId)
     {
-        // Check: validate the user-provided parameters and cliff time.
+        // If sentinel value of zero is used for unlock granularity, change it to one second.
+        if (unlockGranularity == 0) {
+            unlockGranularity = 1;
```
andreivladbrg: good idea with this one

lockup/src/abstracts/SablierLockupState.sol
```diff
@@ -48,10 +48,19 @@ abstract contract SablierLockupState is ISablierLockupState {
...
 
+    /// @dev Unlock granularity mapped by stream IDs, used in LL streams.
+    mapping(uint256 streamId => uint40 unlockGranularity) internal _unlockGranularities;
+
     /*//////////////////////////////////////////////////////////////////////////
                                      MODIFIERS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @dev Checks that actual model and expected model are equal.
+    modifier notEqual(Lockup.Model actualModel, Lockup.Model expectedModel) {
```
andreivladbrg: i believe the `notEqual` name is quite vague 

what about: 

```solidity
modifier modelCheck(Lockup.Model actual, Lockup.Model expected) {
}
``` 

or if we want a more assertive version `notExpectedModel / notModelExpected`

lockup/src/libraries/Helpers.sol
```diff
@@ -164,11 +168,22 @@ library Helpers {
...
+                // Calculate the streamable range as the difference between end time and cliff time.
+                streamableRange = timestamps.end - cliffTime;
+            }
         }
         // Check: the cliff unlock amount is zero when the cliff time is zero.
         else if (unlockAmounts.cliff > 0) {
             revert Errors.SablierHelpers_CliffTimeZeroUnlockAmountNotZero(unlockAmounts.cliff);
         }
+        // Calculate the streamable range when cliff time is zero.
+        else {
```
andreivladbrg: what do you think about this change - i got a bit confused initially about the double "else":

```solidity
else if (cliffTime == 0) {
    // Check: the cliff unlock amount is zero when the cliff time is zero.
    if (unlockAmounts.cliff > 0) {
        revert Errors.SablierHelpers_CliffTimeZeroUnlockAmountNotZero(unlockAmounts.cliff);
    }

    // Calculate the streamable range when cliff time is zero.
    unchecked {
        streamableRange = timestamps.end - timestamps.start;
    }
}
```

lockup/src/interfaces/ISablierLockupState.sol
```diff
@@ -82,6 +82,11 @@ interface ISablierLockupState {
     /// @return unlockAmounts See the documentation in {LockupLinear} type.
     function getUnlockAmounts(uint256 streamId) external view returns (LockupLinear.UnlockAmounts memory unlockAmounts);
 
+    /// @notice Retrieves the smallest step in time between two consecutive unlocks.
```
andreivladbrg: here we say "two consecutive unlocks" but we also have: "two consecutive **token** unlocks"

i'd say to add "token" as well here and in: `ISablierLockupLinear.sol#L87` and `types/LockupLinear.sol#L10`

lockup/src/libraries/Helpers.sol
```diff
@@ -143,16 +144,19 @@ library Helpers {
                             PRIVATE READ-ONLY FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @dev Checks the user-provided cliff, end times, and unlock amounts of an LL stream.
+    /// @dev Checks the user-provided timestamps of an Lockup Linear stream.
```
andreivladbrg: why the change "Lockup Linear" instead of "LL"? 

we still have places where we say LL - so i'd say we change it everywhere or keep it as LL

lockup/src/types/LockupDynamic.sol
```diff
@@ -3,9 +3,9 @@ pragma solidity >=0.8.22;
 
 import { UD2x18 } from "@prb/math/src/UD2x18.sol";
 
-/// @notice Namespace for the structs used only in LD streams.
+/// @notice Namespace for the structs used only in Lockup Dynamic streams.
```
andreivladbrg: same here, LD --> Lockup Dynamic everywhere or keep LD ?

lockup/src/types/LockupLinear.sol
```diff
@@ -1,15 +1,18 @@
...
     /// @notice Struct encapsulating the cliff duration and the total duration used at runtime in
     /// {SablierLockupLinear.createWithDurationsLL} function.
     /// @param cliff The cliff duration in seconds.
     /// @param total The total duration in seconds.
+    /// @param unlockGranularity The smallest step in time between two consecutive unlocks. Zero is a sentinel value for
+    /// 1 second.
     struct Durations {
         uint40 cliff;
         uint40 total;
+        uint40 unlockGranularity;
```
andreivladbrg: IMO it’s confusing to add this to the `Durations` struct, as it makes me think that this has to be different from the one passed in the `createWithTimestampsLL` function, but the granularity has the same meaning for both (like the unlock amounts)


so, i'd change it to:

```solidity
function createWithDurationsLL(
    Lockup.CreateWithDurations calldata params,
    LockupLinear.UnlockAmounts calldata unlockAmounts,
    uint40 unlockGranularity,
    LockupLinear.Durations calldata durations
)

function createWithTimestampsLL(
    Lockup.CreateWithTimestamps calldata params,
    LockupLinear.UnlockAmounts calldata unlockAmounts,
    uint40 unlockGranularity,
    uint40 cliffTime
)
```

lockup/src/types/LockupTranched.sol
```diff
@@ -1,9 +1,9 @@
 // SPDX-License-Identifier: GPL-3.0-or-later
 pragma solidity >=0.8.22;
 
-/// @notice Namespace for the structs used only in LT streams.
+/// @notice Namespace for the structs used only in Lockup Tranched streams.
```
andreivladbrg: LT -> Lockup Tranched

lockup/src/libraries/LockupMath.sol
```diff
@@ -181,19 +193,28 @@ library LockupMath {
...
             } else {
-                elapsedTime = ud(blockTimestamp - cliffTime);
-                streamableRange = ud(endTime - cliffTime);
+                elapsedTimeInGranularityUnits = toScaledUD60x18((blockTimestamp - cliffTime) / unlockGranularity);
+                streamablePeriodInGranularityUnits = ud(endTime - cliffTime).div(ud(unlockGranularity));
             }
 
-            UD60x18 elapsedTimePercentage = elapsedTime.div(streamableRange);
+            // Calculate the elapsed time percentage taking into account the unlock granularity.
+            UD60x18 elapsedTimePercentage = elapsedTimeInGranularityUnits.div(streamablePeriodInGranularityUnits);
```
andreivladbrg: motivation for this comment: initially, the feature was requested by someone who wanted to use LT, but it was not possible to have so many tranches in the UI (and because of the gas limit). so the user might want to have specific timestamps and **exact** amounts unlocked

by having a double division (1st `streamablePeriodInGranularityUnits`, 2nd `elapsedTimePercentage`) and then  the multiplication, we wouldn’t be able to achieve accurate "unlocks"

for example, i wrote this test:

```solidity
function test_weekly_unlocks() public {
    _defaultParams.cliffTime = 0;
    _defaultParams.unlockAmounts.cliff = 0;
    _defaultParams.createWithTimestamps.depositAmount = 52e18;
   _defaultParams.createWithTimestamps.timestamps.end = _defaultParams.createWithTimestamps.timestamps.start + (1 weeks * 52); // almost 1 Y duration
    _defaultParams.unlockGranularity = 1 weeks;

    uint256 id = createDefaultStream();

    vm.warp(_defaultParams.createWithTimestamps.timestamps.start + 1 + (1 weeks * 51));
    assertEq(lockup.streamedAmountOf(id), 51e18);
}
``` 

and it reverts because `streamedAmountOf` returns `50999999999999999988 `, which is not accurate and does not achieve the intended functionality with _precision_

the `50999999999999999988` value is returned until the current time passes the end time, which means that the last unlock is greater than the first ones, and i believe it should be the other way around, to have this invariant: `last unlock <= first ones`


to fix this, we would need to change the order of the `mul` and `div`:

```diff
- UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
- uint128 streamedAmount = unlockAmountsSum + (elapsedTimePercentage.mul(streamableAmount)).intoUint128();
+ uint128 streamedAmount = unlockAmountsSum + elapsedTimeInGranularityUnits
+         .mul(streamableAmount)
+         .div(streamablePeriodInGranularityUnits).intoUint128();
```

which i think would actually return more mathematically accurate results in the case of `granularity = 1` as well

i don’t remember if there was a reason for not doing this in the first place. maybe it introduces issues

andreivladbrg: the big problem with this change is that we will have the streamed amount return different results (in the case of the LL model - `granularity = 1`) between versions

lockup/src/types/LockupLinear.sol
andreivladbrg: or we can keep the `LockupLinear.Durations` as it is in the PR, but add a struct for `createWithTimestampsLL` to group together `unlockGranularity`  and `cliffTime`

wdyt?

lockup/src/abstracts/SablierLockupState.sol
```diff
@@ -48,10 +48,19 @@ abstract contract SablierLockupState is ISablierLockupState {
     /// @dev Unlock amounts mapped by stream IDs, used in LL streams.
     mapping(uint256 streamId => LockupLinear.UnlockAmounts unlockAmounts) internal _unlockAmounts;
 
+    /// @dev Unlock granularity mapped by stream IDs, used in LL streams.
+    mapping(uint256 streamId => uint40 unlockGranularity) internal _unlockGranularities;
```
andreivladbrg: seeing this variable close to `_unlockAmounts` makes me think the variables are related due to the "unlock" keyword, but the name alone would be accurate if we didn’t already have these amounts

since the granularity applies only to the "streamable range", what would you say if we rename the variable to something different?
some alternatives: `streamGranularity` / `streamInterval` / `releaseInterval` / `streamingPeriod`

"period" is a common term in math / physics

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -71,18 +71,41 @@ contract StreamedAmountOf_Lockup_Linear_Integration_Concrete_Test is
...
+        _defaultParams.unlockGranularity = 3000 seconds;
+        uint256 streamId = createDefaultStream();
+
+        // Warp to a timestamp such that only one unlock granularity tick passes.
+        vm.warp({ newTimestamp: defaults.CLIFF_TIME() + 3900 seconds });
+
+        uint256 streamableAmount = defaults.DEPOSIT_AMOUNT() - defaults.CLIFF_AMOUNT();
+
+        uint256 actualStreamedAmount = lockup.streamedAmountOf(streamId);
+        uint256 expectedStreamedAmount = defaults.CLIFF_AMOUNT() + (streamableAmount * 0.4e18) / 1e18;
```
andreivladbrg, line 98: its not clear to me where this 0.4e18 came from?

or if there is a motivation behind 3000 seconds

lockup/src/libraries/LockupMath.sol
```diff
@@ -181,19 +193,28 @@ library LockupMath {
...
             if (cliffTime == 0) {
-                elapsedTime = ud(blockTimestamp - startTime);
-                streamableRange = ud(endTime - startTime);
+                elapsedTimeInGranularityUnits = toScaledUD60x18((blockTimestamp - startTime) / unlockGranularity);
+                streamablePeriodInGranularityUnits = ud(endTime - startTime).div(ud(unlockGranularity));
             } else {
-                elapsedTime = ud(blockTimestamp - cliffTime);
-                streamableRange = ud(endTime - cliffTime);
+                elapsedTimeInGranularityUnits = toScaledUD60x18((blockTimestamp - cliffTime) / unlockGranularity);
+                streamablePeriodInGranularityUnits = ud(endTime - cliffTime).div(ud(unlockGranularity));
```
andreivladbrg: disclaimer: i am not sure which is the best option to go for, but it’s worth discussing

---

using the cliff time as the starting point in the "streamable range" makes sense at first

but let’s use this scenario as an example:
- user wants a 1-week granularity from **start time**, so that it unlocks exactly at the start of each week
- at the same time have a cliff time,  `cliff time = start + 3 weeks + 1 day`  
  - basically the condition when `(cliff - start) % granularity != 0`

in the graph below you can see visually what i am referring to

notice that the period of time from the cliff to the 4000 amount unlocked is 6 days and the granularity is 7

image

in the current version, this won’t be achievable

a solution would be to have granularity configurable, and depending on what is set we would do:

```
offset = (cliffTime - startTime) % granularity
remainder = granularity - offset
streamableRange = (endTime - (cliffTime + remainder))  ...
``` 

but this would be considerably more complex

lockup/tests/integration/fuzz/lockup-linear/streamedAmountOf.t.sol
```diff
@@ -20,7 +20,7 @@ contract StreamedAmountOf_Lockup_Linear_Integration_Fuzz_Test is Lockup_Linear_I
 
         // Bound the unlock amounts.
         unlockAmounts.start = boundUint128(unlockAmounts.start, 0, depositAmount);
-        unlockAmounts.cliff = boundUint128(unlockAmounts.start, 0, depositAmount - unlockAmounts.start);
+        unlockAmounts.cliff = boundUint128(unlockAmounts.cliff, 0, depositAmount - unlockAmounts.start);
```
andreivladbrg, line 23: how this did not cause issues? 🤔

lockup/tests/invariant/README.md
```diff
@@ -60,6 +60,8 @@
 
 12. For a Linear stream,
     - If Cliff time > 0, $`\implies`$ Cliff time > Start time.
-    - End time > Cliff time
+    - End time > Cliff time,
+    - The streamed amount should not exceed the amount that would have been streamed if the stream had a granularity
+      of 1.
```
andreivladbrg, line 75: i believe a more useful one would be:

_streamed amount minus the unlock amounts is always a multiple of granularity_

and the implementation would be something like:

```solidity
uint256 streamed = lockup.streamedAmountOf(streamId);
uint256 unlockAmountsSum =
    lockup.getUnlockAmounts(streamId).start + lockup.getUnlockAmounts(streamId).cliff;
uint256 streamedNoUnlocks = streamed - unlockAmountsSum;

assertTrue(
    streamedNoUnlocks % lockup.getUnlockGranularity(streamId) == 0,
    "Invariant violation: streamed amount - unlock amounts not multiple of granularity"
);
``` 

but i believe it would fail because of the comments left under the `LockupMath.sol` file

andreivladbrg (changes requested):
Great work Shub

leaving my review below. there are some important decisions to take

lockup/src/abstracts/SablierLockupState.sol
smol-ninja: Agree, will go with `modelCheck`. Feel free to make the change.

lockup/src/libraries/Helpers.sol
smol-ninja: Sounds good. We dont need `else if`, just `else` is enough here.

lockup/src/interfaces/ISablierLockupState.sol
smol-ninja: Sounds good.

lockup/src/libraries/Helpers.sol
smol-ninja: We can go with LL. I found that at some places we were using Lockup Linear and thats why i changed it. Agree to be consistent with LL terminology here.

lockup/src/types/LockupLinear.sol
smol-ninja: I like the first approach where we have a separate `unlockGranularity` parameter to both `createWithDurationsLL` and `createWithTimestampsLL` functions.

If we grp `unlockGranularity` and `cliffTime` what would you call the struct?

lockup/src/libraries/LockupMath.sol
smol-ninja: Your suggest sounds good. Go ahead and implement it. I would like you yo implement this and run the tests to see if all tests pass without making any new change to them.

> we will have the streamed amount return different results

I think it will be very minute.

lockup/src/abstracts/SablierLockupState.sol
smol-ninja: > variables are related due to the "unlock" keyword

I kind of like `unlockGranularity` name. These two variables are indeed related to how tokens unlock in a LL stream. So how is that not relevant? Having `granularity` in it would also make the documentation easier. For example, the time step can be called as time granules etc.

Among the suggestions, I like `streamGranularity`. There could be a name with keyword `discrete` in it. However, I still prefer `unlockGranularity`. 

To remove the confusion, should we add more comments over it?

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
smol-ninja: > its not clear to me where this 0.4e18 came from?

Hand calculation. 3000 seconds is random. Feel free to do it other way if you have a better idea.

lockup/tests/integration/fuzz/lockup-linear/streamedAmountOf.t.sol
smol-ninja: Not enough runs I guess.

lockup/src/libraries/LockupMath.sol
smol-ninja: > user wants a 1-week granularity from start time

If we define that granularity only applies after cliff time, is that a problem?

lockup/tests/invariant/README.md
smol-ninja: > streamed amount minus the unlock amounts is always a multiple of granularity

I believe it cant be an exact multiple due to some rounding errors. Right?

lockup/src/abstracts/SablierLockupState.sol
andreivladbrg: @PaulRBerg requesting your opinion as well. what’s your view on the name of this new variable?

lockup/src/libraries/LockupMath.sol
andreivladbrg: > If we define that granularity only applies after cliff time, is that a problem?

technically its not a problem, its a design choice
but the scenario presented in my comment above, won't be achievable by starting from cliff time

lockup/tests/invariant/README.md
andreivladbrg: > I believe it cant be an exact multiple due to some rounding errors. Right?

will research it after the change from [this comment](https://github.com/sablier-labs/lockup/pull/1366#discussion_r2690263572), but most like you are right

lockup/src/libraries/LockupMath.sol
smol-ninja: The scenario you have assumes user wants a 1-week granularity from start time. IMO when there is a cliff time, I do not understand why user would not want to have granularity after the cliff time and not from the start time.

Let me know if you think I am misunderstanding your scenario.

lockup/src/types/LockupTranched.sol
andreivladbrg: b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/types/LockupLinear.sol
andreivladbrg: implemented the separate granularity in both functions b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/types/LockupDynamic.sol
andreivladbrg: b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/libraries/Helpers.sol
andreivladbrg: b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/interfaces/ISablierLockupState.sol
andreivladbrg: b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/libraries/Helpers.sol
andreivladbrg: b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/abstracts/SablierLockupState.sol
andreivladbrg: b710e79ae9f785d711a88c6e418e7c09b5a2fc06

lockup/src/abstracts/SablierLockupState.sol
PaulRBerg: @andreivladbrg you make a fair point. It's not just about the `_unlockAmounts` variable, but also end-user-facing shape names (e.g. "Unlock in Steps") further add to the potential confusion.

What if we call it just `_granularities`? Short and sweet.

smol-ninja: `_granularities` is a very generic time (can be used for any property). Should we call it `_ timeGranularities`?

PaulRBerg: Can it really be used for any property? I don't think it's so generic. Also, the property is always tied to a particular stream. And the "granularity of the stream" sounds quite specific to me.

I don't like `_timeGranularities`.

In any case, if we don't reach an agreement here shortly, let's merge the PR as-is, and we can return to this later.

smol-ninja: OK lets go with `_granularities`. I reasoned with ChatGPT and it says that "granularity of the stream" would generally mean granularity on the time axis.

@andreivladbrg feel free to push a commit.

lockup/scripts/solidity/Init.s.sol
```diff
@@ -56,6 +56,7 @@ contract Init is BaseScript {
                     shape: "Cliff Linear"
                 }),
                 LockupLinear.UnlockAmounts({ start: 0, cliff: 0 }),
+                1 seconds,
```
smol-ninja: Lets use named arguments here so that its clear what `1 seconds` represent.

lockup/src/interfaces/ISablierLockupLinear.sol
```diff
@@ -39,13 +41,16 @@ interface ISablierLockupLinear is ISablierLockupState {
     /// - All requirements in {createWithTimestampsLL} must be met for the calculated parameters.
     ///
     /// @param params Struct encapsulating the function parameters, which are documented in {Lockup} type.
-    /// @param durations Struct encapsulating (i) cliff period duration and (ii) total stream duration, both in seconds.
     /// @param unlockAmounts Struct encapsulating (i) the amount to unlock at the start time and (ii) the amount to
     /// unlock at the cliff time.
+    /// @param unlockGranularity The smallest step in time between two consecutive token unlocks. Zero is a sentinel value for
+    /// 1 second.
+    /// @param durations Struct encapsulating (i) cliff period duration and (ii) total stream duration, both in seconds.
```
smol-ninja, line 48: Nice catch.

lockup/tests/integration/fuzz/lockup-linear/createWithDurationsLL.t.sol
```diff
@@ -25,17 +33,19 @@ contract CreateWithDurationsLL_Integration_Fuzz_Test is Lockup_Linear_Integratio
...
 
         // Expect the relevant event to be emitted.
-        vm.expectEmit({ emitter: address(lockup) });
-        emit ISablierLockupLinear.CreateLockupLinearStream({
-            streamId: expectedStreamId,
-            commonParams: defaults.lockupCreateEvent(timestamps),
-            cliffTime: cliffTime,
-            unlockAmounts: unlockAmounts
-        });
+        // vm.expectEmit({ emitter: address(lockup) });
```
smol-ninja: I suppose you commented it for testing. Lets uncomment it now.

andreivladbrg: yep

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
andreivladbrg: let's keep it then

lockup/src/abstracts/SablierLockupLinear.sol
```diff
@@ -55,14 +56,16 @@ abstract contract SablierLockupLinear is
             timestamps: timestamps,
             token: params.token,
             transferable: params.transferable,
-            unlockAmounts: unlockAmounts
+            unlockAmounts: unlockAmounts,
+            granularity: granularity
```
andreivladbrg, line 60: ordering this alphabetically causes "stack too deep" error

lockup/scripts/solidity/Init.s.sol
andreivladbrg: 7e75fae1d89acf521cf5ec02a03e34257faca932

lockup/tests/integration/fuzz/lockup-linear/createWithDurationsLL.t.sol
andreivladbrg: 7e75fae1d89acf521cf5ec02a03e34257faca932

lockup/src/abstracts/SablierLockupLinear.sol
smol-ninja: Huh?? Order or an extra variable? In that case, lets reverse the order?

airdrops/src/SablierMerkleLL.sol
```diff
@@ -88,6 +91,7 @@ contract SablierMerkleLL is
         VESTING_START_TIME = params.vestingStartTime;
         VESTING_START_UNLOCK_PERCENTAGE = params.startUnlockPercentage;
         VESTING_TOTAL_DURATION = params.totalDuration;
+        VESTING_GRANULARITY = params.granularity;
```
smol-ninja: Should be above `VESTING_START_TIME` (alphabetical)

airdrops/src/SablierMerkleLL.sol
```diff
@@ -46,6 +46,9 @@ contract SablierMerkleLL is
     /// @inheritdoc ISablierMerkleLL
     UD60x18 public immutable override VESTING_CLIFF_UNLOCK_PERCENTAGE;
 
+    /// @inheritdoc ISablierMerkleLL
+    uint40 public immutable override VESTING_GRANULARITY;
```
smol-ninja, line 50: Shouldn't we call it `STREAM_GRANULARITY`? As per [this comment](https://github.com/sablier-labs/lockup/pull/1366#discussion_r2707549498), its defined as the granularity of the stream.

lockup/src/abstracts/SablierLockupLinear.sol
andreivladbrg: order - it has to be after the struct 

feel free to try it, maybe you can make it work

smol-ninja: OK I will have a look.

airdrops/src/SablierMerkleLL.sol
andreivladbrg: good catch

airdrops/src/SablierMerkleLL.sol
andreivladbrg: yeah, lets do it

lockup/src/libraries/LockupMath.sol
andreivladbrg: 305a0e140cb698271f706d471bcf61ff4c3cc3b6

lockup/src/abstracts/SablierLockupLinear.sol
smol-ninja: I am also seeing the same error. Now I think having `granularity` a part of `durations` was not such a bad idea (one variable less on the stack). 

Lets move forward with `granularity` as the last argument for `_createLL`.

andreivladbrg: > Lets move forward with `granularity` as the last argument for `_createLL`.


yeah, it is like this now

lockup/tests/invariant/README.md
andreivladbrg: i can't make it work - it fails

going to mark it as resolved

lockup/src/libraries/LockupMath.sol
```diff
@@ -181,24 +193,32 @@ library LockupMath {
...
 
-            UD60x18 elapsedTimePercentage = elapsedTime.div(streamableRange);
+            // Calculate the streamable amount.
             UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
 
-            // The streamed amount is the sum of the unlock amounts plus the product of elapsed time percentage and
-            // streamable amount.
-            uint128 streamedAmount = unlockAmountsSum + (elapsedTimePercentage.mul(streamableAmount)).intoUint128();
+            // The streamed amount is the sum of the unlock amounts plus the streamable amount multiplied by
+            // elapsed time in granularity units divided by streamable period in granularity units.
```
smol-ninja: The comment is very technical. I'd suggest to keep my version for simplicity.

"The streamed amount is the sum of the unlock amounts plus the product of elapsed time percentage and the streamable amount."

lockup/tests/utils/Constants.sol
```diff
@@ -7,7 +7,7 @@ abstract contract Constants {
     //////////////////////////////////////////////////////////////////////////*/
 
     uint64 public constant BATCH_SIZE = 10;
-    uint128 public constant CLIFF_AMOUNT = 2500e18 + 2534;
```
smol-ninja, line 10: Does the test fail if we keep `2534`? I wonder why you had to remove it.

smol-ninja (commented):
Comments below:

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -47,10 +47,11 @@ contract StreamedAmountOf_Lockup_Linear_Integration_Concrete_Test is
 
     function test_GivenStartAmount() external givenSTREAMINGStatus givenCliffTimeNotZero givenCliffTimeInPast {
         _defaultParams.unlockAmounts.start = 1;
+        _defaultParams.unlockAmounts.cliff -= 1;
```
smol-ninja: Why'd you have to add this?

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
```diff
@@ -47,10 +47,11 @@ contract StreamedAmountOf_Lockup_Linear_Integration_Concrete_Test is
...
         uint256 streamId = createDefaultStream();
         vm.warp({ newTimestamp: defaults.WARP_26_PERCENT() });
         uint128 actualStreamedAmount = lockup.streamedAmountOf(streamId);
-        uint128 expectedStreamedAmount = defaults.STREAMED_AMOUNT_26_PERCENT() + 1;
+        uint128 expectedStreamedAmount = defaults.STREAMED_AMOUNT_26_PERCENT();
```
smol-ninja, line 53: Why'd you have to remove `+1` from here?

lockup/tests/invariant/README.md
smol-ninja: Cool. The invariant that the streamed amount should never exceed the the streamed amount for a linear curve is sufficient. We don't need the exact multiple.

lockup/tests/utils/Constants.sol
andreivladbrg: yes, they were failing

we no longer need this remainder `0.000000000000002534e18`, which was due to the precision loss we had because of first dividing and then multiplying in the streamed amount function

lockup/src/libraries/LockupMath.sol
andreivladbrg: alright

airdrops/src/SablierMerkleLL.sol
andreivladbrg: after making the change locally and seeing how it looks, it feels better to keep it with "VESTING", as all the other variables have this prefix and all are related to the stream created in the claim function

lockup/tests/utils/Constants.sol
smol-ninja: sounds good.

airdrops/src/SablierMerkleLL.sol
smol-ninja: ok

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
andreivladbrg: we don't need it, removed it here 8d9340694e6089a7308c36b0af27a6e7ae5d5b5a

lockup/src/libraries/LockupMath.sol
andreivladbrg: f3649cb70589ac907c2ff3b110b33e641c544f13

lockup/tests/integration/concrete/lockup-linear/streamed-amount-of/streamedAmountOf.t.sol
andreivladbrg: same reason as above https://github.com/sablier-labs/lockup/pull/1366#discussion_r2708989037

lockup/src/libraries/LockupMath.sol
andreivladbrg: > I do not understand why user would not want to have granularity after the cliff time

i guess it’s just about optionality - we don't know in advance what might user want or not

as said in the initial message, it’s quite complex, and i just wanted to point this out. if it’s requested in the future, we might come back to this

airdrops/src/SablierMerkleLL.sol
smol-ninja: I think this is the only unresolved left

andreivladbrg: f1112c55b8f0fe27bc5dd87e49d89fd10b11e627

andreivladbrg (approved):
Image

PR #1367, opened 2026-01-01
refactor(utils): rename data to targetCallData in execute function

### LINKED ISSUE #1327: Rename `data` to `targetCallData` in `execute` function
`data` is a very generic term and renaming it to `targetCallData` would make it explicit about its purpose.

https://github.com/sablier-labs/lockup/blob/893eabb58200339ab3f9e29ee0bbab9b85e876b5/utils/src/SablierComptroller.sol#L177

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1327

### DIFF
--- a/utils/src/SablierComptroller.sol
+++ b/utils/src/SablierComptroller.sol
@@ -192,11 +192,19 @@ contract SablierComptroller is
     }
 
     /// @inheritdoc ISablierComptroller
-    function execute(address target, bytes calldata data) external override onlyAdmin returns (bytes memory result) {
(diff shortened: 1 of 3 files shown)


andreivladbrg:
@smol-ninja why did the coverage drop so much?

smol-ninja:
I dont know the reason but the coverage is not correctly working in the monorepo setup. I suggest to fix the coverage in the end (I think its fixable).

andreivladbrg (approved):
👍

PR #1369, opened 2026-01-02
refactor(utils): combine SetCustomFeeUSD and DisableCustomFeeUSDs events into UpdateCustomFeeUSD

### LINKED ISSUE #1330: Merge `SetCustomFeeUSD` and `DisableCustomFeeUSD` events into `UpdateCustomFeeUSD`
When a custom fee is set, the following event is emitted:

```solidity
event SetCustomFeeUSD(
    Protocol indexed protocol,
    address caller,
    address indexed user,
    uint256 previousMinFeeUSD,
    uint256 newMinFeeUSD
);
```

Then when its disabled, the following event is emitted:

```solidity
event DisableCustomFeeUSD(
    Protocol indexed protocol,
    address caller,
    address indexed user,
    uint256 previousMinFeeUSD,
    uint256 newMinFeeUSD
);
```

The indexer would have to keep track of both the events in order to fetch the latest fee applied on the user which is more complicated than if both of these events were merged into one single event.

```solidity
event UpdateCustomFeeUSD(
    Protocol indexed protocol,
    address caller,
    address indexed user,
    uint256 previousMinFeeUSD,
    uint256 newMinFeeUSD
);
```

Now the indexer can calculate the latest fee applied on the user using just one event.

Is there a reason behind having event names to match the function names? If not, any objection if we merge these two events into one?

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1330

### DIFF
--- a/utils/src/SablierComptroller.sol
+++ b/utils/src/SablierComptroller.sol
@@ -182,7 +182,7 @@ contract SablierComptroller is
         delete _protocolFees[protocol].customFeesUSD[user];
 
         // Log the update.
-        emit ISablierComptroller.DisableCustomFeeUSD({
+        emit ISablierComptroller.UpdateCustomFeeUSD({
             protocol: protocol,
             caller: msg.sender,
             user: user,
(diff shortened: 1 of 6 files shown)


andreivladbrg (approved):
shogun_lgtm

PR #1371, opened 2026-01-04
feat(utils): add a function to lower campaign using callers with fee management role

### LINKED ISSUE #1350: Allow lowering fee on existing campaigns using caller with fee management role
### Problem

Currently, on an existing campaign, the fee can be lowered only through [`execute` function](https://github.com/sablier-labs/lockup/blob/staging/utils/src/SablierComptroller.sol#L195). The problem is that `execute` cannot be called by an EOA with fee management role. This creates inefficiencies in managing custom fees for users.

### Solution

Implement a `lowerMinFeeUSD` in comptroller similar to `setCustomFeeUSD`.

### Implementation

```solidity
function lowerMinFeeUSD(address campaign, uint256 newMinFeeUSD) external;
```

smol-ninja:
### Changelog

- Closes https://github.com/sablier-labs/lockup/issues/1350
- Moves test mock contracts to a separate mock directory


utils/src/SablierComptroller.sol
```diff
@@ -221,6 +221,41 @@ contract SablierComptroller is
...
+    {
+        // Check: the campaign address is a valid Sablier Merkle.
+        (bool success, bytes memory returnData) = campaign.staticcall(abi.encodeWithSignature("IS_SABLIER_MERKLE()"));
+        if (!success || returnData.length == 0) {
+            revert Errors.SablierComptroller_MissingIsSablierMerkle(campaign);
+        }
+
+        // Check: the campaign returns true for `IS_SABLIER_MERKLE()`.
+        if (!abi.decode(returnData, (bool))) {
+            revert Errors.SablierComptroller_IsSablierMerkleReturnsFalse(campaign);
```
andreivladbrg: i believe this and the one above check are not that useful

we already have the `onlyRole(FEE_MANAGEMENT_ROLE)` in this function, and in the campaign we use the `onlyComptroller()` modifier

this execution could have been achieved anyway by calling the `execute` function which doesn't have these

andreivladbrg: also, do we have the `IS_SABLIER_MERKLE` function in all versions we have the `lowerMinFeeUSD` function?

utils/src/SablierComptroller.sol
```diff
@@ -221,6 +221,41 @@ contract SablierComptroller is
...
+        }
+
+        // Interaction: call the lowerMinFeeUSD function on the campaign.
+        (success, returnData) = campaign.call(abi.encodeWithSignature("lowerMinFeeUSD(uint256)", newMinFeeUSD));
+
+        // If the call fails, bubble up the revert reason.
+        if (!success) {
+            assembly {
```
andreivladbrg, line 248: why is the `// solhint-disable-next-line no-inline-assembly` missing, and it doesn’t throw an error?

utils/tests/integration/concrete/comptroller/lower-min-fee-usd-for-campaign/lowerMinFeeUSDForCampaign.tree
```diff
@@ -0,0 +1,17 @@
+LowerMinFeeUSDForCampaign_Comptroller_Concrete_Test
+├── when caller not admin
+│  ├── when caller without fee management role
+│  │  └── it should revert
+│  └── when caller with fee management role
+│     └── it should succeed
+└── when caller admin
+   ├── when campaign not implement Sablier Merkle
```
andreivladbrg: if we keep it, i'd be more explicit here

```suggestion
   ├── when campaign not implement "IS_SABLIER_MERKLE" boolean
```

utils/tests/mocks/MerkleMock.sol
```diff
@@ -0,0 +1,30 @@
...
+    function lowerMinFeeUSD(uint256 newMinFeeUSD) external { }
+}
+
+contract MerkleMockReverting {
+    function IS_SABLIER_MERKLE() external pure returns (bool) {
+        return true;
+    }
+
+    function lowerMinFeeUSD(uint256) external pure {
+        revert("Not gonna happen");
```
andreivladbrg, line 18: should we have different types of reverts (not only 1)?
1. `revert("String")`
2. `require(condition, "String")`
3. `revert MyCustomError()`

utils/tests/integration/concrete/comptroller/lower-min-fee-usd-for-campaign/lowerMinFeeUSDForCampaign.t.sol
```diff
@@ -0,0 +1,90 @@
...
+        // It should revert.
+        vm.expectRevert(abi.encodeWithSelector(Errors.UnauthorizedAccess.selector, users.eve, FEE_MANAGEMENT_ROLE));
+        comptroller.lowerMinFeeUSDForCampaign(address(merkleMock), AIRDROP_MIN_FEE_USD);
+    }
+
+    function test_WhenCallerWithFeeManagementRole() external whenCallerNotAdmin {
+        setMsgSender(users.accountant);
+
+        // It should not revert.
+        comptroller.lowerMinFeeUSDForCampaign(address(merkleMock), AIRDROP_MIN_FEE_USD);
```
andreivladbrg:
```suggestion
        comptroller.lowerMinFeeUSDForCampaign(address(merkleMock), AIRDROP_MIN_FEE_USD - 1);
```

andreivladbrg: or declare an internal value: `newMinFeeUSD = AIRDROP_MIN_FEE_USD - 1;` and use it

andreivladbrg (changes requested):
left my review below

utils/src/SablierComptroller.sol
smol-ninja: Yeah you are right. We don't need these. The `onlyRole` modifier is sufficient check. Feel free to remove it.

utils/src/SablierComptroller.sol
smol-ninja: It will revert after rebase because of https://github.com/sablier-labs/lockup/pull/1352.

utils/tests/integration/concrete/comptroller/lower-min-fee-usd-for-campaign/lowerMinFeeUSDForCampaign.tree
smol-ninja: Lets remove it

utils/tests/mocks/MerkleMock.sol
smol-ninja: Too verbose imo. If it works for one, wouldn't it work for others too?

utils/tests/integration/concrete/comptroller/lower-min-fee-usd-for-campaign/lowerMinFeeUSDForCampaign.t.sol
smol-ninja: sounds good. I think `AIRDROP_MIN_FEE_USD - 1)` is sufficient.

utils/src/SablierComptroller.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1371/changes/c0a57b04edb8e1a5d2bbfd47c503f44a3907bec4

utils/tests/mocks/MerkleMock.sol
andreivladbrg: fair enough

utils/src/SablierComptroller.sol
andreivladbrg: 2cd50d820461ba8239607eb00f7fba1eb62f83df

utils/tests/integration/concrete/comptroller/lower-min-fee-usd-for-campaign/lowerMinFeeUSDForCampaign.tree
andreivladbrg: 2cd50d820461ba8239607eb00f7fba1eb62f83df

utils/tests/integration/concrete/comptroller/lower-min-fee-usd-for-campaign/lowerMinFeeUSDForCampaign.t.sol
andreivladbrg: i added it because its used in all calls 2cd50d820461ba8239607eb00f7fba1eb62f83df

andreivladbrg (approved):

andreivladbrg:
@smol-ninja applied my feedback in the 2cd50d820461ba8239607eb00f7fba1eb62f83df commit - if it looks good to you, feel free to merge it

smol-ninja:
@andreivladbrg make sure to run `just setup` after this PR is merged. It will create new symlinks for `.prettierignore` similar to `.env`. I have bumped the `devkit` and now it would use `mdformat` for markdown linting.

PR #1373, opened 2026-01-07
chore: update the copyright year to 2026 in LICENSE

Radovenchyk:


smol-ninja:
Thanks for the PR @Radovenchyk. From what I am reading [here](https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html), it seems to me that there is no requirement to update the year until there are new significant changes. In that case, I would recommend to open a PR against the staging branch as thats where we have new changes.

Issue #1376, opened 2026-01-08
Add payable `transformFromPayable` function in Flow


smol-ninja:
This is not possible 👇

~As discussion in https://github.com/sablier-labs/lockup/discussions/1375, add `payable` modifier to `transferFrom` and `safeTransferFrom` functions so that they can be used in conjunction with withdraw related functions using `batch`.~

So we decided to implement the following:

As discussion in https://github.com/sablier-labs/lockup/discussions/1375, add payable `transferFromPayable` function so that it can be used in conjunction with withdraw related functions using `batch`.

### Reference

- [ERC721.sol](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/ERC721.sol)

PR #1384, opened 2026-01-19
feat(lockup, flow): add transferFromPayable

andreivladbrg:
closes https://github.com/sablier-labs/lockup/issues/1376

i implemented `transferFromPayable` instead because the compiler doesn’t allow adding `override` and the `payable` specifier to a non-payable function:

```
     Error: Compiler run failed:                                                                                       
     Error (6959): Overriding function changes state mutability from "nonpayable" to "payable".                        
        --> src/SablierFlow.sol:527:5:                                                                                 
         |                                                                                                             
     527 |     function transferFrom(address from, address to, uint256 tokenId) public payable override {              
         |     ^ (Relevant source part starts here and spans across multiple lines).                                   
     Note: Overridden function is here: 
``` 

note: i added the function to `lockup` as well because i think it’s good for consistency. lmk if you agree or if you consider the increased size a problem (it’s still below the margin)


flow/tests/integration/concrete/transfer-from-payable/transferFromPayable.tree
```diff
@@ -0,0 +1,7 @@
+TransferFromPayable_Integration_Concrete_Test
+├── when ETH value is zero
+│  ├── it should transfer the NFT
+│  └── it should emit {MetadataUpdate} and {Transfer} events
+└── when ETH value is greater than zero
```
andreivladbrg, line 2: lmk if you think we only need one branch, or if keeping both is ok

also, do you think we should use `transferFromPayable` in the `batch` tests? IMO these are sufficient

andreivladbrg:
@smol-ninja the fork tests seem to be failing. i tested locally, and this is not related to this PR, as it failed in `staging` as well. let’s fix it in another PR

smol-ninja:
@andreivladbrg the fork tests are fixed in https://github.com/sablier-labs/lockup/pull/1366. See https://github.com/sablier-labs/lockup/commit/2539c05c7e792073b0b14d2b5cb769c448561c0f

smol-ninja:
> note: i added the function to lockup as well because i think it’s good for consistency

I don't agree with this. We shouldn't add function just for the consistency. And also the size problem.

flow/tests/integration/concrete/transfer-from-payable/transferFromPayable.tree
smol-ninja: I think one branch is sufficient (non-zero ETH). If it can accept >0 ETH, it can definitely accept 0 ETH :)

> also, do you think we should use transferFromPayable in the batch tests? IMO these are sufficient

Yeah these are sufficient.

flow/src/interfaces/ISablierFlow.sol
```diff
@@ -459,6 +459,14 @@ interface ISablierFlow is
     /// @param newNFTDescriptor The address of the new NFT descriptor contract.
     function setNFTDescriptor(IFlowNFTDescriptor newNFTDescriptor) external;
 
+    /// @notice Wrapper for {IERC721.transferFrom} with the `payable` specifier so that it can be called in conjunction
```
smol-ninja, line 462:
```suggestion
    /// @notice Wrapper for {IERC721.transferFrom} with the `payable` modifier so that it can be batched
```

flow/src/interfaces/ISablierFlow.sol
```diff
@@ -459,6 +459,14 @@ interface ISablierFlow is
...
+    /// with other functions using {IBatch.batch}.
+    ///
+    /// @param from The address of the current owner of the stream NFT.
+    /// @param to The address of the new owner of the stream NFT.
+    /// @param streamId The ID of the stream NFT to transfer.
+    function transferFromPayable(address from, address to, uint256 streamId) external payable;
```
smol-ninja, line 472: We should add `Requirements` section to this:

```
/// Requirements:
/// - Refer to the requirements in https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721.sol.
```

flow/src/interfaces/ISablierFlow.sol
```diff
@@ -459,6 +459,14 @@ interface ISablierFlow is
     /// @param newNFTDescriptor The address of the new NFT descriptor contract.
     function setNFTDescriptor(IFlowNFTDescriptor newNFTDescriptor) external;
 
+    /// @notice Wrapper for {IERC721.transferFrom} with the `payable` specifier so that it can be called in conjunction
+    /// with other functions using {IBatch.batch}.
+    ///
+    /// @param from The address of the current owner of the stream NFT.
```
smol-ninja: The owner of the stream ID = the current owner. But saying `The owner of the stream ID` is more explicit.

```suggestion
    /// @param from The owner of the stream ID.
```

Also, would it be better to say "Stream ID" because its parameter of this function? Also, we refer to stream ID more often than stream NFT.

flow/src/interfaces/ISablierFlow.sol
smol-ninja, line 472: If its just a wrapper, should we use `tokenId` i.e. keep the function exactly same as https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC721/IERC721.sol?

flow/src/SablierFlow.sol
```diff
@@ -522,6 +522,11 @@ contract SablierFlow is
         uri = nftDescriptor.tokenURI({ sablierFlow: this, streamId: streamId });
     }
 
+    /// @inheritdoc ISablierFlow
+    function transferFromPayable(address from, address to, uint256 streamId) external payable {
+        ERC721.transferFrom({ from: from, to: to, tokenId: streamId });
```
smol-ninja, line 527: What do you think about using internal function `_transfer` or `_safeTransfer` here instead of public `transferFrom`?

smol-ninja (changes requested):

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: its the same in [_update](https://github.com/sablier-labs/lockup/blob/cbbc52d1698dff78e4a4be16dcc354ae462d621f/flow/src/SablierFlow.sol#L598) and we say `streamId`

andreivladbrg: it's implicit IMO - but we can add it sure

smol-ninja: OK.

smol-ninja: You are right but I was thinking that it wont show up in the etherscan natspec unless we explicitly specify it.

flow/src/SablierFlow.sol
andreivladbrg: there is no `_transferFrom` and `transferFrom`  has more checks than `_transfer`: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/239795bea728c8dca4deb6c66856dd58a6991112/contracts/token/ERC721/ERC721.sol#L312

smol-ninja: I meant [_safeTransfer](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/token/ERC721/ERC721.sol#L378). And [_transfer](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/token/ERC721/ERC721.sol#L347) has same number of checks as [transferFrom](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/token/ERC721/ERC721.sol#L137).

andreivladbrg: we shouldn't use "_safe" - it would force recipients to implement the erc721 hooks

and `_transfer`: `As opposed to {transferFrom}, this imposes no restrictions on msg.sender.`

andreivladbrg:
> I don't agree with this. We shouldn't add function just for the consistency. And also the size problem.

fair enough, we don't have `transferTokens` in lockup as well - i'll remove it

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: `payable` its not a "modifier"

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: 2f912c9be987c1626a37e11e12a1522cc08efe1f

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: ac1268c1a8fd0864d791e7d1d4c98c07f8db48a4

flow/src/SablierFlow.sol
smol-ninja: > we shouldn't use "_safe" - it would force recipients to implement the erc721 hooks

Good point.

> and _transfer: As opposed to {transferFrom}, this imposes no restrictions on msg.sender

What restriction?

flow/src/interfaces/ISablierFlow.sol
smol-ninja: Ouch. The solidity docs use the term "payable function". I have some suggestion but lets agree/disagree with https://github.com/sablier-labs/lockup/pull/1384#discussion_r2714459315 first, as I think we only need `_transfer`

flow/src/SablierFlow.sol
andreivladbrg: > What restriction?

that’s from openzeppelin’s natspec

in `transferFrom`, it’s passed as the [auth = _msgSender](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/239795bea728c8dca4deb6c66856dd58a6991112/contracts/token/ERC721/ERC721.sol#L119-L121) param in `_update`, so it’s checked

in the internal [`_transfer`](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/239795bea728c8dca4deb6c66856dd58a6991112/contracts/token/ERC721/ERC721.sol#L325) one, it’s passed `address(0)`, thus no check

smol-ninja: Oops. I missed that. LGTM.

flow/src/interfaces/ISablierFlow.sol
smol-ninja: I am going to resolve this.

flow/src/interfaces/ISablierFlow.sol
```diff
@@ -459,6 +459,18 @@ interface ISablierFlow is
     /// @param newNFTDescriptor The address of the new NFT descriptor contract.
     function setNFTDescriptor(IFlowNFTDescriptor newNFTDescriptor) external;
 
+    /// @notice Wrapper for {IERC721.transferFrom} with the `payable` specifier so that it can be called in conjunction
+    /// with other functions using {IBatch.batch}.
+    ///
+    /// Requirements:
+    /// - Refer to the requirements in:
+    /// https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/token/ERC721/IERC721.sol.
```
smol-ninja: Should we say

Refer to the requirements of `transferFrom` in https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v5.3.0/contracts/token/ERC721/IERC721.sol?

flow/src/interfaces/ISablierFlow.sol
```diff
@@ -459,6 +459,18 @@ interface ISablierFlow is
...
+    ///
+    /// @param from The owner of the stream ID.
+    /// @param to The address of the new owner of the stream NFT.
```
smol-ninja:
```suggestion
    /// @param to The address of the new owner of the stream ID.
```

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: sure

andreivladbrg: or, what do you think about pasting the exact place?

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/e4f70216d759d8e6a64144a9e1f7bbeed78e7079/contracts/token/ERC721/IERC721.sol#L75-L91

smol-ninja: For sure.

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: 9eadcbc75ef0da2b6fbb3a4047c53aaa88b80c29

flow/src/interfaces/ISablierFlow.sol
andreivladbrg: 9eadcbc75ef0da2b6fbb3a4047c53aaa88b80c29

andreivladbrg:
@smol-ninja if it looks good, could you approve it?

smol-ninja (approved):
Image

PR #1390, opened 2026-01-22
fix: don't use PRB math in LL calculation

andreivladbrg:
as a consequence of the invariant failing: https://github.com/sablier-labs/lockup/actions/runs/21181095875/job/60923599405?pr=1370

the prb math alternative (may be it can be simplified) i managed to make work is:

<details><summary>click to see it</summary>
<p>

```solidity
// Determine the reference time (cliff or start).
uint40 referenceTime = cliffTime == 0 ? startTime : cliffTime;

// Calculate elapsed granularity units (floored) and scale to UD60x18.
UD60x18 elapsedGranularityUnits = convert((blockTimestamp - referenceTime) / granularity);

UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
UD60x18 scaledGranularity = convert(granularity);
UD60x18 scaledTotalDuration = convert(endTime - referenceTime);

UD60x18 streamedPortion =
    elapsedGranularityUnits.mul(streamableAmount).mul(scaledGranularity).div(scaledTotalDuration);

// The streamed amount is the sum of the unlock amounts plus the streamed portion.
uint128 streamedAmount = unlockAmountsSum + streamedPortion.intoUint128();

```

</p>
</details> 

which has significantly more math operations. i’d now be in favor of keeping native types in ll, though there would be no consistency between models

### DIFF
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -215,7 +215,8 @@ library LockupMath {
             uint256 streamableAmount = depositedAmount - unlockAmountsSum;
 
             // Scaled numerator: floor(elapsed/g) * streamableAmount * g * 1e18.
-            uint256 streamedPortionScaled = elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) * 1e18;
+            uint256 streamedPortionScaled =
+                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) * 1e18;
 
             // Unlock amounts plus the streamed portion.
             uint128 streamedAmount =
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -1,11 +1,10 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.8.22;
 
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { PRBMathCastingUint128 as CastingUint128 } from "@prb/math/src/casting/Uint128.sol";
 import { PRBMathCastingUint40 as CastingUint40 } from "@prb/math/src/casting/Uint40.sol";
 import { SD59x18 } from "@prb/math/src/SD59x18.sol";
-
-import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 import { LockupDynamic } from "../types/LockupDynamic.sol";
 import { LockupLinear } from "../types/LockupLinear.sol";
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -141,9 +141,9 @@ library LockupMath {
     /// $$
     ///
     ///     ⌊time elapsed / granularity⌋ * granularity
-    /// x =
-    /// ───────────────────────────────────────────
+    /// x = －－－－－－－－－－－－－－－－－－－－－－－－－
     ///                streamable time
+    ///
     /// $$
     ///
     /// The floor division in the numerator creates discrete unlock steps at every granularity seconds.
@@ -199,27 +199,26 @@ library LockupMath {
             // granularity.
             uint256 elapsedTimeInGranularityUnits;
 
-            // Calculate the streamable period scaled.
-            uint256 streamableTotalDurationScaled;
+            // Calculate the streamable duration.
+            uint256 streamableTotalDuration;
 
             if (cliffTime == 0) {
                 elapsedTimeInGranularityUnits = (blockTimestamp - startTime) / granularity;
-                streamableTotalDurationScaled = uint256(endTime - startTime) * 1e18;
+                streamableTotalDuration = uint256(endTime - startTime);
             } else {
                 elapsedTimeInGranularityUnits = (blockTimestamp - cliffTime) / granularity;
-                streamableTotalDurationScaled = uint256(endTime - cliffTime) * 1e18;
+                streamableTotalDuration = uint256(endTime - cliffTime);
             }
 
             // Calculate the streamable amount.
             uint256 streamableAmount = depositedAmount - unlockAmountsSum;
 
-            // Scaled numerator: floor(elapsed/g) * streamableAmount * g * 1e18.
-            uint256 streamedPortionScaled =
-                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) * 1e18;
+            // Calculate the streamed portion: floor(elapsed/g) * streamableAmount * g / streamableTotalDuration.
+            uint256 streamedPortion =
+                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) / streamableTotalDuration;
 
-            // Unlock amounts plus the streamed portion.
-            uint128 streamedAmount =
-                unlockAmountsSum + SafeCast.toUint128(streamedPortionScaled / streamableTotalDurationScaled);
+            // The streamed amount is the sum of the unlock amounts plus the streamed portion.
+            uint128 streamedAmount = unlockAmountsSum + SafeCast.toUint128(streamedPortion);
 
             // Although the streamed amount should never exceed the deposited amount, this condition is checked
             // without asserting to avoid locking tokens in case of a bug. If this situation occurs, the withdrawn
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: BUSL-1.1
 pragma solidity >=0.8.22;
 
-import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { PRBMathCastingUint128 as CastingUint128 } from "@prb/math/src/casting/Uint128.sol";
 import { PRBMathCastingUint40 as CastingUint40 } from "@prb/math/src/casting/Uint40.sol";
 import { SD59x18 } from "@prb/math/src/SD59x18.sol";
@@ -217,8 +216,9 @@ library LockupMath {
             uint256 streamedPortion =
                 elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) / streamableTotalDuration;
 
-            // The streamed amount is the sum of the unlock amounts plus the streamed portion.
-            uint128 streamedAmount = unlockAmountsSum + SafeCast.toUint128(streamedPortion);
+            // The streamed amount is the sum of the unlock amounts plus the streamed portion. The cast to uint128 is
+            // safe because `floor(elapsed/g) * g < streamableTotalDuration`, so `streamedPortion < streamableAmount`.
+            uint128 streamedAmount = unlockAmountsSum + uint128(streamedPortion);
 
             // Although the streamed amount should never exceed the deposited amount, this condition is checked
             // without asserting to avoid locking tokens in case of a bug. If this situation occurs, the withdrawn
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -4,7 +4,8 @@ pragma solidity >=0.8.22;
 import { PRBMathCastingUint128 as CastingUint128 } from "@prb/math/src/casting/Uint128.sol";
 import { PRBMathCastingUint40 as CastingUint40 } from "@prb/math/src/casting/Uint40.sol";
 import { SD59x18 } from "@prb/math/src/SD59x18.sol";
-import { convert as toScaledUD60x18, UD60x18, ud } from "@prb/math/src/UD60x18.sol";
+
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 import { LockupDynamic } from "../types/LockupDynamic.sol";
 import { LockupLinear } from "../types/LockupLinear.sol";
@@ -139,9 +140,11 @@ library LockupMath {
     /// - $x$ is the elapsed time percentage with discrete unlocks:
     ///
     /// $$
-    ///        ⌊time elapsed / granularity⌋
-    /// x = ────────────────────────────────────
-    ///       streamable time / granularity
+    ///
+    ///     ⌊time elapsed / granularity⌋ * granularity
+    /// x =
+    /// ───────────────────────────────────────────
+    ///                streamable time
     /// $$
     ///
     /// The floor division in the numerator creates discrete unlock steps at every granularity seconds.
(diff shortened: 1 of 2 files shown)


andreivladbrg:
@smol-ninja lmk what you think

smol-ninja:
Thanks for the PR. I will review it later but I am in agreement with you.

I am in favour of removing PRB Math because the streamed amount will now be more accurate (https://github.com/sablier-labs/lockup/commit/305a0e140cb698271f706d471bcf61ff4c3cc3b6 removed residual `2534` from `CLIFF_AMOUNT` and the tests in this PR are passing implying that streamed amount is more accurate).

I also want to point out that invariant test was failing because of changes introduced by you in https://github.com/sablier-labs/lockup/commit/305a0e140cb698271f706d471bcf61ff4c3cc3b6 leading to over streaming of tokens. However, the implementation prior to https://github.com/sablier-labs/lockup/commit/305a0e140cb698271f706d471bcf61ff4c3cc3b6 was returning a value of the streamed amount which was 0.00000000000000002% (infinitesimal) less compared to theoretical value. But the implementation of this PR returns the theoretical value (to be testes).

lockup/src/libraries/LockupMath.sol
```diff
@@ -139,9 +140,11 @@ library LockupMath {
     /// - $x$ is the elapsed time percentage with discrete unlocks:
     ///
     /// $$
-    ///        ⌊time elapsed / granularity⌋
-    /// x = ────────────────────────────────────
-    ///       streamable time / granularity
+    ///
+    ///     ⌊time elapsed / granularity⌋ * granularity
+    /// x =
+    /// ───────────────────────────────────────────
```
andreivladbrg: needs to be fixed

lockup/src/libraries/LockupMath.sol
```diff
@@ -195,30 +198,29 @@ library LockupMath {
...
+                streamableTotalDurationScaled = uint256(endTime - cliffTime) * 1e18;
             }
 
             // Calculate the streamable amount.
-            UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
+            uint256 streamableAmount = depositedAmount - unlockAmountsSum;
+
+            // Scaled numerator: floor(elapsed/g) * streamableAmount * g * 1e18.
+            uint256 streamedPortionScaled =
+                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) * 1e18;
```
andreivladbrg: this math operation shouldn't overflow, since: 
(i) the max value that `elapsedTimeInGranularityUnits` can have is `type(uint40).max - 1`
(ii) the max value of `streamableAmount` is `type(uint128).max`
(iii) if `elapsedTimeInGranularityUnits` is at its max value, then `granularity` should be 1,

and all of them multiplied with `1e18` is ≈ 3.7e68 < 2^256

lockup/src/libraries/LockupMath.sol
smol-ninja: We can revert natspec change. Natspecs are for understanding of the logic and dont necessarily need to copy the exact implementation.

smol-ninja:
```
    ///        ⌊time elapsed / granularity⌋
    /// x = ────────────────────────────────────
    ///       streamable time / granularity
    ```

is self explanatory.

File: lockup/src/libraries/LockupMath.sol [hunk 3]
```diff
@@ -195,30 +198,29 @@ library LockupMath {
...
+                elapsedTimeInGranularityUnits = (blockTimestamp - cliffTime) / granularity;
+                streamableTotalDurationScaled = uint256(endTime - cliffTime) * 1e18;
             }
 
             // Calculate the streamable amount.
-            UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
+            uint256 streamableAmount = depositedAmount - unlockAmountsSum;
+
+            // Scaled numerator: floor(elapsed/g) * streamableAmount * g * 1e18.
+            uint256 streamedPortionScaled =
```
smol-ninja: This is not `streamedPortionScaled`, this is just some numerator.

The streamed portion scaled would be

```solidity
SafeCast.toUint128(streamedPortionScaled / streamableTotalDurationScaled)
```

Can we calculate `streamedPortionScaled` here and then just add it to `unlockAmountsSum` below. In that case, we wont need `1e18` here.

lockup/src/libraries/LockupMath.sol
```diff
@@ -195,30 +198,29 @@ library LockupMath {
...
+                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) * 1e18;
 
-            // The streamed amount is the sum of the unlock amounts plus the product of elapsed time percentage and the
-            // streamable amount.
-            uint128 streamedAmount = unlockAmountsSum
-                + elapsedTimeInGranularityUnits.mul(streamableAmount).div(streamablePeriodInGranularityUnits)
-                    .intoUint128();
+            // Unlock amounts plus the streamed portion.
```
smol-ninja:
```suggestion
            // The streamed amount is the sum of the unlock amounts plus the streamed amount.
```

smol-ninja (changes requested):
I am OK with the changes in this PR. Left some comments below, if you agree, make sure to update comments over code too.

lockup/src/libraries/LockupMath.sol
```diff
@@ -195,30 +198,29 @@ library LockupMath {
 
             // Calculate the time elapsed in granularity units as the floor division of time elapsed and unlock
             // granularity.
-            UD60x18 elapsedTimeInGranularityUnits;
+            uint256 elapsedTimeInGranularityUnits;
```
smol-ninja, line 199: Why not use `uint40` for this?

lockup/src/libraries/LockupMath.sol
```diff
@@ -195,30 +198,29 @@ library LockupMath {
...
             } else {
-                elapsedTimeInGranularityUnits = toScaledUD60x18((blockTimestamp - cliffTime) / granularity);
-                streamablePeriodInGranularityUnits = ud(endTime - cliffTime).div(ud(granularity));
+                elapsedTimeInGranularityUnits = (blockTimestamp - cliffTime) / granularity;
+                streamableTotalDurationScaled = uint256(endTime - cliffTime) * 1e18;
             }
 
             // Calculate the streamable amount.
-            UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
+            uint256 streamableAmount = depositedAmount - unlockAmountsSum;
```
smol-ninja, line 213: should be `uint128`

smol-ninja (changes requested):

lockup/src/libraries/LockupMath.sol
andreivladbrg: that's true - but i couldn't find a good name, and i believe having this variable is important to see the `mul` operations separately

lockup/src/libraries/LockupMath.sol
andreivladbrg: it would overflow here: 

https://github.com/sablier-labs/lockup/blob/a484730d0ae1a0934377e613d833b105f078e970/lockup/src/libraries/LockupMath.sol#L218-L219

all types need to be `uint256`

lockup/src/libraries/LockupMath.sol
andreivladbrg: nope, same reason as [above](https://github.com/sablier-labs/lockup/pull/1390#discussion_r2727085415)

lockup/src/libraries/LockupMath.sol
andreivladbrg: > Can we calculate `streamedPortionScaled` here and then just add it to `unlockAmountsSum` below. In that case, we wont need `1e18` here.

is your suggestion to do all operations in a single line?
if yes, i believe it would decrease the readability of the code

smol-ninja: I dont think it would impact the readability of the code. For someone studying the code, calculating `streamedPortionScaled` in a single line is more useful as it clearly states what the calculated amount represents.

Having numerator as a separate line does not add any value to its readability. 

```solidity
uint256 streamedPortionScaled = (elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity))
        / streamableTotalDurationScaled;
```

It would just be this. And you wont need `1e18` in the calculation of `streamableTotalDurationScaled`

lockup/src/libraries/LockupMath.sol
smol-ninja: Are you sure? Because `granularity` is casted to `uint256`, I think other can be `uint40`. The type of final value is the max type of any single value.

lockup/src/libraries/LockupMath.sol
andreivladbrg: > I dont think it would impact the readability of the code. For someone studying the code, calculating `streamedPortionScaled` in a single line is more useful as it clearly states what the calculated amount represents.
> 
> Having numerator as a separate line does not add any value to its readability.
> 
> ```solidity
> uint256 streamedPortionScaled = (elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity))
>         / streamableTotalDurationScaled;
> ```
> 
> It would just be this. And you wont need `1e18` in the calculation of `streamableTotalDurationScaled`

ooh, i thought adding the `1e18` would increase the precision, but since we do this now: `elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity)` it won't be affected

happy to make the change

lockup/src/libraries/LockupMath.sol
andreivladbrg: this test overflows:

```solidity
    function test_overflow_2() public {
        uint40 elapsedTimeInGranularityUnits = type(uint40).max - 1;
        uint128 streamableAmount = type(uint128).max / 2;
        uint256 granularity = 2;
        uint256 result = elapsedTimeInGranularityUnits * streamableAmount * granularity;
        console.log("Overflow test result:", result);
    }
```

anyway, what's the point of having them as `uint40` and `uint128` if they are anyway cast in the mul operation?

i believe it's the right choice to keep them all as `uint256` for consistency and safety - given that the code is in `unchecked` block 
if we make changes in the future it can silently overflow/underflow, so prevention

lockup/src/libraries/LockupMath.sol
andreivladbrg: 32370f0106a2379b3d65446a4e83f4f357e25500

lockup/src/libraries/LockupMath.sol
andreivladbrg: 32370f0106a2379b3d65446a4e83f4f357e25500

lockup/src/libraries/LockupMath.sol
andreivladbrg: 32370f0106a2379b3d65446a4e83f4f357e25500
lmk what you think about this new version

i'd like to have the actual logic matching the math formula

lockup/src/libraries/LockupMath.sol
smol-ninja: It can never overflow because `uint40 + uint128 + uint40 < 208` and having exact `type` guarantees that it never overflows. On the other hand, if we have `uint256` for all variables, you will have to verify the type of right hand side variables are within limits to make sure thats its just casting and wont overflow ever.

The code you provided overflows because how the multiplications are grouped together (first two first). If you run the code below, it wont overflow:

```solidity
function test_overflow_2() public {
    uint40 elapsedTimeInGranularityUnits = type(uint40).max - 1;
    uint128 streamableAmount = type(uint128).max / 2;
    uint256 granularity = 2;
    uint256 result = elapsedTimeInGranularityUnits * (streamableAmount * granularity);
}
```
First it multiplies `streamableAmount * granularity` and cast it to `uint256` because granularity is `uint256` and then we multiply `uint40 * uint256` which is safe.

lockup/src/libraries/LockupMath.sol
smol-ninja: > i'd like to have the actual logic matching the math formula

Whats the point of that? The actual logic is less intuitive than the above self explanatory comment. 

You version is technically correct, but I still think it does not explain why `granularity` is in the numerator which requires further explanation. And if we can explain the math in simple terms, a reader can easily decode the actual logic written in the code.

Also, you can use `────────────────` instead of `－－－－－－－－－－`

andreivladbrg: > The actual logic is less intuitive than the above self explanatory comment.

what do you mean, it's a math formula that must be implemented in the code
 
> Whats the point of that? And if we can explain the math in simple terms, a reader can easily decode the actual logic written in the code.

option 1 (current):
- pro: fewer words, fewer things to explain, the math function matches the code exactly
- con: just not as beautiful the dash in the fraction
   - do you really dislike it that much? IMO it's fine, anyone reading it will understand that's a fraction

option 2 (your suggestion):
- pro: more beautiful dash in the fraction
- con: more words, more things to understand when reading the natspec. not all developers have math knowledge (even though it's basic algebra)

its even harder for the AI - we would need to explain these:

```
1.
///        ⌊time elapsed / granularity⌋
/// x = ────────────────────────────────────
///       streamable time / granularity <───
2.
///        ⌊time elapsed / granularity⌋
/// x = ────────────────────────────────────
///       streamable time * (1 / granularity) <───
3.
///        ⌊time elapsed / granularity⌋ * granularity <───
/// x = ───────────────────────────────────────────────
///                streamable time 
```

i chose option 1 any day :))

lockup/src/libraries/LockupMath.sol
andreivladbrg: as discussed on slack

we ll keep the `uint256`

lockup/src/libraries/LockupMath.sol
andreivladbrg: as discussed on slack

we ll keep the uint256

lockup/src/libraries/LockupMath.sol
smol-ninja: How are there more words to understand in my suggestion?

What I am suggesting:

```solidity
///        ⌊time elapsed / granularity⌋
/// x = ────────────────────────────────────
///       streamable time / granularity
```

vs what you implemented:

```solidity
///     ⌊time elapsed / granularity⌋ * granularity
/// x = －－－－－－－－－－－－－－－－－－－－－－－－－
///                streamable time
```

But given that first may require a bit more Math to understand, lets keep the second one (your suggestion). Though both have floor function notation which requires some knowledge of Math anyways.

andreivladbrg: > How are there more words to understand in my suggestion?

the math function explained in the natspec would be different from its implementation in the code - and this would need to be explained, why they are mathematically the same

> But given that first may require a bit more Math to understand, lets keep the second one

thanks, resolving

andreivladbrg:
before merging @paulrberg, pinging you to ask if there are objections since we'll no longer use the lib for the LL model  
this current new version is simpler and leads to more accurate results

PaulRBerg (approved):
LGTM

lockup/src/libraries/LockupMath.sol
```diff
@@ -195,30 +197,28 @@ library LockupMath {
...
+            uint256 streamedPortion =
+                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) / streamableTotalDuration;
 
-            // The streamed amount is the sum of the unlock amounts plus the product of elapsed time percentage and the
-            // streamable amount.
-            uint128 streamedAmount = unlockAmountsSum
-                + elapsedTimeInGranularityUnits.mul(streamableAmount).div(streamablePeriodInGranularityUnits)
-                    .intoUint128();
+            // The streamed amount is the sum of the unlock amounts plus the streamed portion.
+            uint128 streamedAmount = unlockAmountsSum + SafeCast.toUint128(streamedPortion);
```
smol-ninja: I was wondering since `streamedPortion` can never exceed `uint128` because `streamableAmount` uses `uint128` type, why don't we just use `uint128(streamedPortion)` here instead of `SafeCast.toUint128`?

andreivladbrg: yeah, it can't overflow

will remove the SafeCast, and add an exaplantory comment, since its important

andreivladbrg: 43c09dd70d43ae55a845294a7a82a5ec55b3d0ac

if it looks good, can you approve it?

lockup/src/libraries/LockupMath.sol
```diff
@@ -195,30 +196,29 @@ library LockupMath {
...
+                streamableTotalDuration = uint256(endTime - cliffTime);
             }
 
             // Calculate the streamable amount.
-            UD60x18 streamableAmount = ud(depositedAmount - unlockAmountsSum);
+            uint256 streamableAmount = depositedAmount - unlockAmountsSum;
+
+            // Calculate the streamed portion: floor(elapsed/g) * streamableAmount * g / streamableTotalDuration.
+            uint256 streamedPortion =
+                elapsedTimeInGranularityUnits * streamableAmount * uint256(granularity) / streamableTotalDuration;
```
smol-ninja: `uint256(granularity)` is redundant, you can just use `granularity` because other two types are `uint256`.

smol-ninja (approved):
Looks good. Just one comment

lockup/src/libraries/LockupMath.sol
andreivladbrg: you’re right that since the first two types are already `uint256`, it won’t overflow

i went for extra explicitness, but i don’t have a strong opinion, since it would be automatically casted to `uint256` anyway

will change

andreivladbrg: 076b1e44c7ef50ba5179e7d88c7da91d0c593bac

smol-ninja:
Thanks feel free to merge.

PR #1392, opened 2026-01-25
chore: remove bash-scripting skill

andreivladbrg:
as discussed here: https://github.com/sablier-labs/lockup/discussions/1385


smol-ninja (approved):
Fast

Issue #1356, opened 2025-12-02
Add a new Merkle-type campaign to allow `claimAndExecute`


andreivladbrg:
We've received requests to support claiming and staking the airdrop so users do not sell the airdrop immediately.

This campaign model is neither an instant drop nor a vesting one.

The factory should pass the following inputs to the campaign constructor:
- the `target` contract  
- `bytes` data  
  - perform a low-level call to the target to verify that the function exists

```solidity
struct ConstructorParams {
    string campaignName;
    uint40 campaignStartTime;
    uint40 expiration;
    address initialAdmin;
    string ipfsCID;
    bytes32 merkleRoot;
    IERC20 token;
    // unique params
    address target; // the target contract which can be a staking contract / lending pool
    bytes data; // the data for the low level call on the target contract
}
``` 

Might be useful to include a string input that is emitted only in the `createMerkle` event (similar to `recipientCount`) so the indexer knows which protocol the campaign will interact with.

**Questions:**
- Should allow multiple target contracts and multiple data per campaign? (this would require a new param in the claim function to decide at claim time what contract to call)
- Since this campaign assumes that the target contract transfers the airdropped amount to the target contract, how do we ensure this at campaign creation time?

The campaign should implement **just** the `claimAndExecute` and no other `claim` or `claimTo` or `claimViaSig`.

```solidity
function claimAndExecute(
    uint256 index,
    uint128 amount,
    bytes32[] calldata merkleProof
)
    external
    nonReentrant
{
    // Check, Effect and Interaction: Pre-process the claim parameters on behalf of `msg.sender`.
    _preProcessClaim({ index: index, recipient: msg.sender, amount: amount, merkleProof: merkleProof });

    // Interaction: Give exact allowance, call, then revoke.
    TOKEN.approve(TARGET, amount);
    (bool success,) = TARGET.call(data);
    if (!success) revert ExecuteFailed(TARGET, data);
    TOKEN.approve(TARGET, 0);
}
``` 

Note: no `recipient` param in the claim function, since allowing it would introduce a security risk.

Related discussions: [1](https://github.com/sablier-labs/lockup/issues/1336) and [2](https://github.com/sablier-labs/lockup/discussions/1351#discussioncomment-15123010)

smol-ninja:
[...]

PR #1393, opened 2026-01-25
feat(airdrops): Add `MerkleExecute` campaign

andreivladbrg:
Closes https://github.com/sablier-labs/lockup/issues/1356

Notes: 
feel free to suggest a better name for the new campaign, don't have a strong preference, i couldn't think of somehting better 
also, lmk if you think we should rename `claimAndExecute` to just `claim`

i thought more about the suggestion made about having 50/50 of the amount directly transferred to the recipient and deposited/staked in the target contract, and i believe it is in favor of campaign creators to **enforce** recipients to fully deposit the airdropped amount (if they don’t pass the full amount in the `arguments` data, then the leftover amount is considered "forgone", like in vca)

also, if we want to implement this idea, we would need one more state variable for the percentage allowed to be directly transferred (how much is deposited in your protocol and how much is transferred directly?), which would complicate the UX for campaign creators

---

currently, `SablierMerkleExecute` includes the signature logic for `claimViaSig`, which is not needed
this logic will be moved to a separate contract in the PR that implements claiming via external attestation


airdrops/src/SablierFactoryMerkleExecute.sol
```diff
@@ -0,0 +1,129 @@
...
+                                  PRIVATE FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Validates the deployment parameters.
+    function _checkDeploymentParams(address token, address target) private view {
+        // Check: user-provided token is not the native token.
+        _forbidNativeToken(token);
+
+        // Check: target is a deployed contract.
+        if (target.code.length == 0) {
```
andreivladbrg, line 125: we could also add some bytes arguments in the create params so that we can check whether the function actually exists - we won't save this anywhere, just to check that the selector actually exists on the target
wdyt?

airdrops/src/interfaces/ISablierMerkleExecute.sol
```diff
@@ -0,0 +1,75 @@
...
+    /// Notes:
+    /// - Unlike other Merkle campaigns, this function does not have a `recipient` parameter. The recipient is always
+    /// `msg.sender` to prevent security risks where someone could claim on behalf of another user and execute
+    /// arbitrary calls.
+    /// - If `APPROVE_TARGET` is true, the function approves the exact claim amount to the {TARGET}, executes the call,
+    /// then revokes the approval.
+    /// - If the target contract transfers tokens and if there is an amount encoded in `arguments`, it must match the
+    /// airdropped `amount`. Otherwise, the remaining tokens will be left in the campaign contract.
+    ///
+    /// Requirements:
```
andreivladbrg, line 44: should we add the `nonReentrant` requirement?

airdrops/src/SablierFactoryMerkleExecute.sol
smol-ninja: yeah sure if thats doable.

airdrops/src/interfaces/ISablierMerkleExecute.sol
smol-ninja: Do we need it?

andreivladbrg: the modifier or the natspec?

currently the claim function uses the modifier

smol-ninja: Oh you mean in the natspec. Nah i dont think we need to add it.

andreivladbrg: agree

airdrops/src/SablierFactoryMerkleExecute.sol
andreivladbrg: i was wrong, this can’t be achievable as long as the selector points to a non-view/non-pure function

airdrops/src/SablierMerkleExecute.sol
```diff
@@ -0,0 +1,121 @@
...
+    function claimAndExecute(
+        uint256 index,
+        uint128 amount,
+        bytes32[] calldata merkleProof,
+        bytes calldata arguments
+    )
+        external
+        payable
+        override
+        nonReentrant
```
smol-ninja, line 90: Its entirely safe to remove `nonReentrant` because a second call would fail during `_preProcessClaim`

airdrops/tests/mocks/MockStaking.sol
smol-ninja, line 1: Why not combine all mock staking contracts into one file? Like how we do with oracle mock etc?

airdrops/src/interfaces/ISablierMerkleExecute.sol
```diff
@@ -0,0 +1,70 @@
...
+interface ISablierMerkleExecute is ISablierMerkleBase {
+    /*//////////////////////////////////////////////////////////////////////////
+                                       EVENTS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Emitted when a claim is executed on behalf of an eligible recipient.
+    /// @param index The index of the airdrop recipient in the Merkle tree.
+    /// @param recipient The address of the airdrop recipient (always `msg.sender`).
+    /// @param amount The amount of ERC-20 tokens claimed.
+    /// @param target The address of the target contract that was called.
```
smol-ninja: Lets remove params from events natspecs. We dont include params in events natspecs but we have accidentally included them in some. I have removed it locally lmk if you agree.

airdrops/src/SablierMerkleExecute.sol
```diff
@@ -0,0 +1,121 @@
...
+        // The {SafeERC20.forceApprove} function is used to handle special ERC-20 tokens (e.g. USDT) that require the
+        // current allowance to be zero before setting it to a non-zero value.
+        TOKEN.forceApprove({ spender: TARGET, value: amount });
+
+        // Prepare the call data by concatenating the selector and the arguments.
+        bytes memory callData = abi.encodePacked(SELECTOR, arguments);
+
+        // Interaction: Execute the call on the target contract.
+        (bool success, bytes memory returnData) = TARGET.call(callData);
+        if (!success) {
```
smol-ninja, line 107: There is one more risk. If the target contract does not implement the `selector` but has `fallback` function that does nothing, the `call` would succeed but nothing would happen.

airdrops/src/interfaces/ISablierMerkleExecute.sol
```diff
@@ -0,0 +1,70 @@
...
+    ///
+    /// @param index The index of `msg.sender` in the Merkle tree.
+    /// @param amount The amount of ERC-20 tokens allocated to `msg.sender`.
+    /// @param merkleProof The proof of inclusion in the Merkle tree.
+    /// @param arguments The function ABI-encoded arguments for {SELECTOR}.
+    function claimAndExecute(
+        uint256 index,
+        uint128 amount,
+        bytes32[] calldata merkleProof,
+        bytes calldata arguments
```
smol-ninja: Thoughts on renaming it to `selectorArguments`?

airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.tree
```diff
@@ -0,0 +1,16 @@
+ClaimAndExecute_MerkleExecute_Integration_Test
+└── when merkle proof valid
+   ├── when arguments not valid
+   │  └── it should revert
+   └── when arguments valid
+      ├── when reentrancy
```
smol-ninja: If we decide to remove `reentrancy`, I would suggest the following test order as I find it more explicit:

```tree
ClaimAndExecute_MerkleExecute_Integration_Test
└── when merkle proof valid
   ├── when target transfer amount overdraws
   │  └── it should revert
   └── when target transfer amount not overdraw
      ├── when target call not succeed
      │  └── it should revert
      └── when target call succeeds
         ├── when reentrancy
         │  └── it should revert
         └── when no reentrancy
            ├── it should mark the index as claimed
            ├── it should transfer the ETH to the comptroller
            ├── it should transfer the tokens to the target
            ├── it should execute the target call
            └── it should emit a {ClaimExecute} event
```

airdrops/tests/integration/concrete/factory/execute/compute-merkle-execute/computeMerkleExecute.tree
```diff
@@ -0,0 +1,8 @@
+ComputeMerkleExecute_Integration_Test
+├── when native token found
+│  └── it should revert
+└── when native token not found
+   ├── when target not contract
+   │  └── it should revert
+   └── when target is contract
```
smol-ninja:
```suggestion
   └── when target contract
```

airdrops/tests/integration/Integration.t.sol
```diff
@@ -59,10 +63,8 @@ abstract contract Integration_Test is Base_Test {
         internal
         virtual
     {
-        // Using `ISablierMerkleInstant` interface over `merkleBase` works for all Merkle contracts due to similarity in
-        // claim function signature.
-        address campaignAddr = address(merkleBase);
-        ISablierMerkleInstant(campaignAddr).claim{ value: msgValue }(index, recipient, amount, merkleProof);
+        if (claimIfMerkleExecute(msgValue, index, amount, merkleProof)) return;
```
smol-ninja: A comment as to which call this function would be helpful.

airdrops/tests/integration/Integration.t.sol
```diff
@@ -143,6 +143,52 @@ abstract contract Integration_Test is Base_Test {
         });
     }
 
+    /*//////////////////////////////////////////////////////////////////////////
+                                    MERKLE-EXECUTE
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Helper to call claimAndExecute for MerkleExecute campaigns.
+    function claimIfMerkleExecute() internal returns (bool) {
```
smol-ninja: Why not merge this inside above `claim()`?

smol-ninja (changes requested):
Nice work on the PR. Leaving some comments below. I have made some changes locally but will push after we have agreements.

FYI I am happy to make resolve the comment from my side if you agree/disagree.

airdrops/tests/mocks/MockStaking.sol
andreivladbrg: sounds good

airdrops/src/interfaces/ISablierMerkleExecute.sol
andreivladbrg: agree to remove natspec from events natspec

airdrops/src/interfaces/ISablierMerkleExecute.sol
andreivladbrg: both work, nothing against either of them, so if you'd like to make the change, feel free to do it

airdrops/src/SablierMerkleExecute.sol
andreivladbrg: don't you think the extra ~2000 gas cost for the claim function (also compared to our claim fee of 1.99) isn't better for "peace of mind"

i don't have an exact exploit POC, but think of this:
- user has two valid claim indices
- target has a hook mechanism so that it can run multiple calls
- re-entrance into the function for the second available claim, double amount allowed to be spent, run recursive transfers in the hooks from target

tho, the security risks are influenced by the staking/lending protocol. as a security prevention, IMO it's better to keep it

airdrops/src/SablierMerkleExecute.sol
andreivladbrg: that's correct

but the campaign creator chooses the `TARGET` and `SELECTOR` at deployment - and they have the grace period to claw back if they misconfigured it, right?

airdrops/tests/integration/Integration.t.sol
andreivladbrg: wdyt: `// If the campaign is a MerkleExecute campaign, call claimAndExecute instead.`

airdrops/tests/integration/Integration.t.sol
andreivladbrg: `claimIfMerkleExecute` is used both `claim` and `claimTo` - we would need to duplicate it

airdrops/src/SablierFactoryMerkleExecute.sol
```diff
@@ -0,0 +1,129 @@
...
+            revert Errors.SablierFactoryMerkleExecute_TargetNotContract(target);
+        }
```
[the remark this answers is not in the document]
smol-ninja: Its not possible to verify selector for non-view functions.

airdrops/tests/integration/concrete/factory/execute/create-merkle-execute/createMerkleExecute.t.sol
```diff
@@ -0,0 +1,124 @@
...
+        });
+
+        ISablierMerkleExecute actualExecute = createMerkleExecute(params);
+        assertLt(0, address(actualExecute).code.length, "MerkleExecute contract not created");
+        assertEq(
+            address(actualExecute), expectedMerkleExecute, "MerkleExecute contract does not match computed address"
+        );
+
+        // It should set the current factory address.
+        assertEq(actualExecute.minFeeUSD(), customFeeUSD, "min fee USD");
```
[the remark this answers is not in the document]
smol-ninja: Good catch, will be fixed.

airdrops/scripts/solidity/CreateMerkleExecute.s.sol
```diff
@@ -0,0 +1,37 @@
...
+        });
+
+        // The total amount to airdrop through the campaign.
+        uint256 aggregateAmount = 10_000e18;
+
+        // The number of eligible users for the airdrop.
+        uint256 recipientCount = 100;
+
+        // Deploy the MerkleExecute contract.
+        merkleExecute = factory.createMerkleExecute(campaignParams, aggregateAmount, recipientCount);
```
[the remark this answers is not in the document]
smol-ninja: Ignore the create script.

airdrops/src/SablierMerkleExecute.sol
smol-ninja: The claim requires `merkleProof` which cannot be generated onchain. So in your example, the re-entrant call won't be able to successfully claim the second time.

Agree with the peace of mind argument but the following line is the true peace of mind:

```solidity
TOKEN.forceApprove({ spender: TARGET, value: 0 });
```

So I do not see why we need `nonReentrant`. I can also argue that we should have `nonReentrant` in all the functions because an ERC20 token `transfer` can also make callback if maliciously implemented.

The only risk I see is that the above line does nothing i.e. an ERC20 token does not reduce allowance at all. Even in that case, the risk is isolated to only user amount.

airdrops/src/SablierMerkleExecute.sol
smol-ninja: Yes. I was just pointing out the risk.

airdrops/src/SablierMerkleExecute.sol
andreivladbrg: > The claim requires `merkleProof` which cannot be generated onchain. So in your example, the re-entrant call won't be able to successfully claim the second time.

the fact that the proof needs to be generated offchain is true, but it does not influence the scenario above, as you can pass anything to `bytes calldata arguments` -  including the `merkleProof`

> Agree with the peace of mind argument but the following line is the true peace of mind:

in the scenario presented above, it does reach that line:
1. call to `TARGET` with `abi.encodePacked(SELECTOR, arguments);`
2. let's say, for example, the target function params include: `(uint256 amount, address to, bytes calldata data)` <-- can include anything
3. calls another claim function without reaching the first `TOKEN.forceApprove({ spender: TARGET, value: 0 });`


> So I do not see why we need nonReentrant. I can also argue that we should have nonReentrant in all the functions because an ERC20 token transfer can also make callback if maliciously implemented.

i believe it is not the same thing, as in the case of classic `ERC20` transfers we don't deterministically pass some data to the transfer function
also, the fact that we renounce the allowance after the target call is important

andreivladbrg: > Agree with the peace of mind argument

lets keep it then

airdrops/src/interfaces/ISablierMerkleExecute.sol
andreivladbrg: as far i can see, some events still include the natspec, do you still have the change locally? if not i can do it, lmk

**ISablierMerkleExecute:**

https://github.com/sablier-labs/lockup/blob/3ee3478160cd27c7c5a1023676892eb71f268b8a/airdrops/src/interfaces/ISablierMerkleExecute.sol#L16-L19

**ISablierMerkleInstant:**

https://github.com/sablier-labs/lockup/blob/3ee3478160cd27c7c5a1023676892eb71f268b8a/airdrops/src/interfaces/ISablierMerkleInstant.sol#L14-L18

**ISablierMerkleLL.sol**

https://github.com/sablier-labs/lockup/blob/3ee3478160cd27c7c5a1023676892eb71f268b8a/airdrops/src/interfaces/ISablierMerkleLL.sol#L15-L37

**ISablierMerkleLT.sol**

https://github.com/sablier-labs/lockup/blob/3ee3478160cd27c7c5a1023676892eb71f268b8a/airdrops/src/interfaces/ISablierMerkleLT.sol#L14-L28

**ISablierMerkleVCA.sol**

https://github.com/sablier-labs/lockup/blob/3ee3478160cd27c7c5a1023676892eb71f268b8a/airdrops/src/interfaces/ISablierMerkleVCA.sol#L16-L41

andreivladbrg: were they added "accidentally" to this many events?

airdrops/src/interfaces/ISablierMerkleExecute.sol
andreivladbrg: i see you kept the current version `arguments`

i suppose you don't want to make the change anymore?

airdrops/src/SablierMerkleExecute.sol
smol-ninja: > calls another claim function without reaching the first

If there is a second call to `claimAndExecute`, the `msg.sender` will be the `TARGET` contract. So it will only succeed if `TARGET` is eligible for airdrop and even in that case, it will only claim its own amount. So, the re-entrant call fails. It wont even work when user has two valid claim indices.

```solidity
// Check, Effect and Interaction: Pre-process the claim parameters on behalf of `msg.sender`.
_preProcessClaim({ index: index, recipient: msg.sender, amount: amount, merkleProof: merkleProof });
```

> in the scenario presented above, it does reach that line

Are you sure there is any concern here?

airdrops/src/interfaces/ISablierMerkleExecute.sol
smol-ninja: Yes I havre removed them locally. Just waiting for agreement on https://github.com/sablier-labs/lockup/pull/1393#discussion_r2784537020 because I will have to revert it if we decide to keep reentrant.

> were they added "accidentally" to this many events?

Not accidentally but I guess I forgot that we don't use detailed natspecs in events.

airdrops/src/interfaces/ISablierMerkleExecute.sol
smol-ninja: I have made the changes. I didn't push the commits yet.

airdrops/tests/integration/Integration.t.sol
smol-ninja: Instead of duplicate what if we just call `claimIfMerkleExecute` inside `claim` and `claimTo`? I think that would be a better approach over `if (claimIfMerkleExecute(msgValue, index, amount, merkleProof)) return`. Wdyt?

andreivladbrg: i am not sure what do you mean, but if you manage to dry-fy, sure go for it

airdrops/src/interfaces/ISablierMerkleExecute.sol
andreivladbrg: > Not accidentally but I guess I forgot that we don't use detailed natspecs in events

"I guess I forgot" - well, doesn't this mean "accidentally" :))

airdrops/src/SablierMerkleExecute.sol
andreivladbrg: > If there is a second call to claimAndExecute, the msg.sender will be the TARGET contract. So it will only succeed if TARGET is eligible for airdrop and even in that case, it will only claim its own amount. So, the re-entrant call fails. It wont even work when user has two valid claim indices.

it still can

1. malicious contract = recipient for index X && the same malicious contract = recipient for index Y
2. first call to campaign for index X
3. `TARGET.foo(uint256 depAmount, address who, bytes calldata hookData)`
4. hookData calls the recipient malicious contract
5. malicious contract calls campaign for index Y
6. `_preProcessClaim` check passes

--- 

do i think this is likely to happen? def not  
is it technically possible? yes  

this would be similar to these checks:

**Lockup:**

https://github.com/sablier-labs/lockup/blob/2ea2dbe6f66d4c0ccbe486d7cccb35b6330953c4/lockup/src/libraries/LockupMath.sol#L223-L228

**Flow:**

https://github.com/sablier-labs/lockup/blob/2ea2dbe6f66d4c0ccbe486d7cccb35b6330953c4/flow/src/SablierFlow.sol#L1067-L1068

we don't have a concrete test to pass them, but we add them as a safety measure

don't you think it's a good thing to keep the reentrant, security-wise?

smol-ninja: > don't you think it's a good thing to keep the reentrant, security-wise?

I don't think so. In all above technically possible scenarios,

- Recipients can only approve up to their claim amounts.
-  Reentrancy does not lead to approving more tokens that expected. In your example, malicious contract approves Ax for X index, then withdraw it followed by approving Ay for Y index and withdrawing it. All calls then set allowance to zero at the end of the execution.

Reentrant adds value when there are no access control. In our case, the fact that only `msg.sender` can call claim for his own amount is sufficient check that no attack is possible.

To support by argument, you can check out [Uniswap V4 code](https://github.com/Uniswap/v4-core/tree/main) and they allows reentry into pool operations (`swap`, `modifyLiquidity`) during hooks, which is by design as it enables composability such as rebalance pool during a swap etc.

----------------

To close this discussion now, I will bring back reentrancy guard in my commit and push them now. We can continue to keep it.

airdrops/tests/mocks/MockStaking.sol
smol-ninja: Addressed in https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

airdrops/src/interfaces/ISablierMerkleExecute.sol
smol-ninja: Addressed in https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

airdrops/src/interfaces/ISablierMerkleExecute.sol
smol-ninja: https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.tree
smol-ninja: https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

airdrops/tests/integration/concrete/factory/execute/compute-merkle-execute/computeMerkleExecute.tree
smol-ninja: https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

airdrops/tests/integration/Integration.t.sol
smol-ninja: Included that and also refactored it in https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

```solidity
       // If the campaign type is "execute", call {claimAndExecute} function.
        if (Strings.equal(campaignType, "execute")) {
            ISablierMerkleExecute(address(merkleBase)).claimAndExecute{ value: msgValue }(
                index, amount, merkleProof, abi.encode(amount)
            );
        }
        // Otherwise, call the `claim` function using the `ISablierMerkleInstant` interface which is compatible with all
        // other Merkle contracts.
        else {
            ISablierMerkleInstant(address(merkleBase)).claim{ value: msgValue }(index, recipient, amount, merkleProof);
        }
```

airdrops/tests/integration/Integration.t.sol
smol-ninja: It was just one line (split into 3 due to line length) so I inlined the logic in https://github.com/sablier-labs/lockup/compare/f67043dbb98f988c488e3dc81144dd79be6b104e...3c04a336f7b90bbaa1bc39bda660c43308280bc8

smol-ninja (approved):

airdrops/src/interfaces/ISablierFactoryMerkleExecute.sol
```diff
@@ -0,0 +1,75 @@
...
+    /// - The create function does not validate if the `campaignParams.selector` is a function implemented by the target
+    /// contract. In that case, the `claimAndExecute` function will revert.
+    /// - If the target contract does not implement the `campaignParams.selector` but has `fallback`, the
+    /// `claimAndExecute` call may silently succeed. If fallback does not transfer claim tokens, the claim tokens will
+    /// be left in the campaign contract. These tokens can be clawbacked by the campaign creator.
+    ///
+    /// Requirements:
+    /// - `campaignParams.token` must not be the forbidden native token.
+    /// - `campaignParams.target` must be a contract.
+    /// - `campaignParams.selector` must be a function implemented by the target contract.
```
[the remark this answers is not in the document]
smol-ninja: Good catch.

https://github.com/sablier-labs/lockup/pull/1393/changes/ebe68d491a19f8a912946ec3f0f593e1b4378693

airdrops/src/types/DataTypes.sol
```diff
@@ -7,6 +7,34 @@ import { UD2x18 } from "@prb/math/src/UD2x18.sol";
...
+    /// @param campaignStartTime The start time of the campaign, as a Unix timestamp.
+    /// @param expiration The expiration of the campaign, as a Unix timestamp. A value of zero means the campaign does
+    /// not expire.
+    /// @param initialAdmin The initial admin of the campaign.
+    /// @param ipfsCID The content identifier for indexing the contract on IPFS. An empty value may break certain UI
+    /// features that depend upon the IPFS CID.
+    /// @param merkleRoot The Merkle root of the claim data.
+    /// @param selector The function selector to call on the target contract when user claim tokens.
+    /// @param target The address of the target contract (staking contract, lending pool) to which the  function
+    /// selector will be called when user claim tokens.
```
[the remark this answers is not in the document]
smol-ninja: Good catch.

https://github.com/sablier-labs/lockup/pull/1393/changes/ebe68d491a19f8a912946ec3f0f593e1b4378693

airdrops/src/SablierMerkleExecute.sol
andreivladbrg: > To support by argument, you can check out [Uniswap V4 code](https://github.com/Uniswap/v4-core/tree/main) and they allows reentry into pool operations (`swap`, `modifyLiquidity`) during hooks, which is by design as it enables composability such as rebalance pool during a swap etc

that's false. uniswap uses reentrancy logic in both [v4](https://github.com/Uniswap/v4-core/blob/d153b048868a60c2403a3ef5b2301bb247884d46/src/PoolManager.sol#L189) and [v3](https://github.com/Uniswap/v3-core/blob/d8b1c635c275d2a9450bd6a78f3fa2484fef73eb/contracts/UniswapV3Pool.sol#L607), and even in v2.

--- 

> To close this discussion now, I will bring back reentrancy guard in my commit and push them now. We can continue to keep it.

thanks, promise it won't bother you

PR #1397, opened 2026-01-26
chore(utils): update admin addresses

andreivladbrg:
@smol-ninja do you think we should a new constant `COMMON_MULTISIG = 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd` ?

### DIFF
--- a/utils/src/tests/BaseScript.sol
+++ b/utils/src/tests/BaseScript.sol
@@ -94,7 +94,7 @@ abstract contract BaseScript is Script {
     }
 
     /// @dev Returns true if the chain uses the default multisig admin.
-    function isDefaultMultisigChain(uint256 id) private pure returns (bool) {
+    function isDefaultMultisigChain(uint256 id) public pure returns (bool) {
         return id == ChainId.ARBITRUM || id == ChainId.AVALANCHE || id == ChainId.BASE || id == ChainId.BERACHAIN
             || id == ChainId.BSC || id == ChainId.ETHEREUM || id == ChainId.GNOSIS || id == ChainId.HYPEREVM
             || id == ChainId.LINEA || id == ChainId.MONAD || id == ChainId.OPTIMISM || id == ChainId.POLYGON
--- a/utils/src/tests/BaseScript.sol
+++ b/utils/src/tests/BaseScript.sol
@@ -83,20 +83,12 @@ abstract contract BaseScript is Script {
     }
 
     /// @notice Returns the admin address to be used for the comptroller.
-    /// @dev The chains listed below use multisig. In all other cases, the default admin is used.
+    /// @dev Chiliz and zkSync have specific multisig addresses. Other mainnets return the common multisig.
+    /// Testnets and unsupported chains return the default admin.
     function getAdmin() public view returns (address) {
-        if (chainId == ChainId.ARBITRUM) return 0xF34E41a6f6Ce5A45559B1D3Ee92E141a3De96376;
-        if (chainId == ChainId.AVALANCHE) return 0x4735517616373c5137dE8bcCDc887637B8ac85Ce;
-        if (chainId == ChainId.BASE) return 0x83A6fA8c04420B3F9C7A4CF1c040b63Fbbc89B66;
-        if (chainId == ChainId.BSC) return 0x6666cA940D2f4B65883b454b7Bc7EEB039f64fa3;
         if (chainId == ChainId.CHILIZ) return 0x74A234DcAdFCB395b37C8c2B3Edf7A13Be78c935;
-        if (chainId == ChainId.ETHEREUM) return 0x79Fb3e81aAc012c08501f41296CCC145a1E15844;
-        if (chainId == ChainId.GNOSIS) return 0x72ACB57fa6a8fa768bE44Db453B1CDBa8B12A399;
-        if (chainId == ChainId.LINEA) return 0x72dCfa0483d5Ef91562817C6f20E8Ce07A81319D;
-        if (chainId == ChainId.OPTIMISM) return 0x43c76FE8Aec91F63EbEfb4f5d2a4ba88ef880350;
-        if (chainId == ChainId.POLYGON) return 0x40A518C5B9c1d3D6d62Ba789501CE4D526C9d9C6;
-        if (chainId == ChainId.SCROLL) return 0x0F7Ad835235Ede685180A5c611111610813457a9;
         if (chainId == ChainId.ZKSYNC) return 0xaFeA787Ef04E280ad5Bb907363f214E4BAB9e288;
+        if (ChainId.isMainnet(chainId)) return 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd;
 
         return DEFAULT_SABLIER_ADMIN;
     }
--- a/utils/src/tests/BaseScript.sol
+++ b/utils/src/tests/BaseScript.sol
@@ -18,7 +18,7 @@ abstract contract BaseScript is Script {
     address public constant DEFAULT_SABLIER_ADMIN = 0xb1bEF51ebCA01EB12001a639bDBbFF6eEcA12B9F;
 
     /// @dev The address of the common safe multisig.
-    address public constant COMMON_SAFE_MULTISIG = 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd;
+    address public constant DEFAULT_SABLIER_MULTISIG_ADMIN = 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd;
 
     /// @dev The salt used for deterministic deployments.
     bytes32 public immutable SALT;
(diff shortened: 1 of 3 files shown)


utils/src/tests/BaseScript.sol
```diff
@@ -83,20 +83,12 @@ abstract contract BaseScript is Script {
     }
 
     /// @notice Returns the admin address to be used for the comptroller.
-    /// @dev The chains listed below use multisig. In all other cases, the default admin is used.
+    /// @dev Chiliz and zkSync have specific multisig addresses. Other mainnets return the common multisig.
+    /// Testnets and unsupported chains return the default admin.
     function getAdmin() public view returns (address) {
-        if (chainId == ChainId.ARBITRUM) return 0xF34E41a6f6Ce5A45559B1D3Ee92E141a3De96376;
```
smol-ninja, line 88: Why you removed these admin addresses?

smol-ninja:
`COMMON_MULTISIG`??? What is that?

andreivladbrg:
> COMMON_MULTISIG??? What is that?

as discussed on slack, waiting for a response

utils/src/tests/BaseScript.sol
andreivladbrg: as discussed on slack - we will use this new address for the next deployments

utils/src/tests/ChainId.sol
smol-ninja, line 140: lets replace this code with `isMainnet`

utils/src/tests/BaseScript.sol
```diff
@@ -83,20 +83,12 @@ abstract contract BaseScript is Script {
...
-        if (chainId == ChainId.BSC) return 0x6666cA940D2f4B65883b454b7Bc7EEB039f64fa3;
         if (chainId == ChainId.CHILIZ) return 0x74A234DcAdFCB395b37C8c2B3Edf7A13Be78c935;
-        if (chainId == ChainId.ETHEREUM) return 0x79Fb3e81aAc012c08501f41296CCC145a1E15844;
-        if (chainId == ChainId.GNOSIS) return 0x72ACB57fa6a8fa768bE44Db453B1CDBa8B12A399;
-        if (chainId == ChainId.LINEA) return 0x72dCfa0483d5Ef91562817C6f20E8Ce07A81319D;
-        if (chainId == ChainId.OPTIMISM) return 0x43c76FE8Aec91F63EbEfb4f5d2a4ba88ef880350;
-        if (chainId == ChainId.POLYGON) return 0x40A518C5B9c1d3D6d62Ba789501CE4D526C9d9C6;
-        if (chainId == ChainId.SCROLL) return 0x0F7Ad835235Ede685180A5c611111610813457a9;
         if (chainId == ChainId.ZKSYNC) return 0xaFeA787Ef04E280ad5Bb907363f214E4BAB9e288;
+        if (ChainId.isMainnet(chainId)) return 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd;
```
smol-ninja: There are many chains where `DEFAULT_SABLIER_ADMIN` is used. So shouldn't we hardcode chains and their safe admin address explicitly?

smol-ninja (changes requested):

utils/src/tests/ChainId.sol
andreivladbrg: good catch

utils/src/tests/BaseScript.sol
andreivladbrg: right

25c25b65be5c21912a9933d257e5ad158e7b9e2b 

lmk what you think 

note: no alphabetical order for`_isCommonSafeChain` - i like the current version as its close to the `getAdmin` function - don't have to scroll down, wdyt?

utils/src/tests/ChainId.sol
andreivladbrg: decided to remove the `isMainnet` - see https://github.com/sablier-labs/lockup/commit/25c25b65be5c21912a9933d257e5ad158e7b9e2b

utils/src/tests/BaseScript.sol
```diff
@@ -17,6 +17,9 @@ abstract contract BaseScript is Script {
     /// @dev The address of the default Sablier admin.
     address public constant DEFAULT_SABLIER_ADMIN = 0xb1bEF51ebCA01EB12001a639bDBbFF6eEcA12B9F;
 
+    /// @dev The address of the common safe multisig.
+    address public constant COMMON_SAFE_MULTISIG = 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd;
```
smol-ninja: We can move it over `DEFAULT_SABLIER_ADMIN`. And what do you think of renaming it to `DEFAULT_SABLIER_MULTISIG_ADMIN` or just `DEFAULT_MULTISIG_ADMIN`.

Its kind of default.

utils/src/tests/BaseScript.sol
```diff
@@ -82,25 +85,23 @@ abstract contract BaseScript is Script {
...
-        if (chainId == ChainId.SCROLL) return 0x0F7Ad835235Ede685180A5c611111610813457a9;
         if (chainId == ChainId.ZKSYNC) return 0xaFeA787Ef04E280ad5Bb907363f214E4BAB9e288;
-
+        if (_isCommonSafeChain(chainId)) return COMMON_SAFE_MULTISIG;
         return DEFAULT_SABLIER_ADMIN;
     }
 
+    /// @dev Returns true if the chain uses the common safe multisig.
+    function _isCommonSafeChain(uint256 id) private pure returns (bool) {
+        return id == ChainId.ARBITRUM || id == ChainId.AVALANCHE || id == ChainId.BASE || id == ChainId.BERACHAIN
```
smol-ninja, line 98: I see you added `BERACHAIN` which was not added here before. Can you please double check that this list is correct?

utils/src/tests/BaseScript.sol
```diff
@@ -82,25 +85,23 @@ abstract contract BaseScript is Script {
         return bytes32(abi.encodePacked(create2Salt));
     }
 
-    /// @notice Returns the admin address to be used for the comptroller.
-    /// @dev The chains listed below use multisig. In all other cases, the default admin is used.
+    /// @notice Returns the admin address for the comptroller.
+    /// @dev Chiliz and zkSync have unique multisig addresses. Other supported mainnets use the common safe.
```
smol-ninja: "Other supported mainnets use the common safe" not true.

Also what if we remove special mention of `Chiliz and zkSync`. The purpose of the function is to returns admin and it can become cluttered if there are more chains that use unique multisig.

utils/tests/integration/fuzz/base-script/BaseScript.t.sol
```diff
@@ -45,35 +45,25 @@ contract BaseScript_Fuzz_Test is StdAssertions {
...
             assertEq(baseScript.getAdmin(), 0xaFeA787Ef04E280ad5Bb907363f214E4BAB9e288, "zksync admin");
+        } else if (_isCommonSafeChain(chainId)) {
+            assertEq(baseScript.getAdmin(), baseScript.COMMON_SAFE_MULTISIG(), "common safe admin");
         } else {
             assertEq(baseScript.getAdmin(), baseScript.DEFAULT_SABLIER_ADMIN(), "default admin");
         }
     }
 
+    /// @dev Returns true if the chain uses the common safe multisig.
+    function _isCommonSafeChain(uint256 id) private pure returns (bool) {
```
smol-ninja: I wonder why we keep a separate copy of this? What if we make `_isCommonSafeChain` a public function and use that here instead.

smol-ninja (changes requested):

utils/tests/integration/fuzz/base-script/BaseScript.t.sol
andreivladbrg: i agree

utils/src/tests/BaseScript.sol
andreivladbrg: its not on notion, but i found it [here](https://berascan.com/address/0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd#code)

utils/src/tests/BaseScript.sol
andreivladbrg: i agree with "DEFAULT"

utils/src/tests/BaseScript.sol
andreivladbrg: right, what is your suggestion for the natspec?

smol-ninja: Lets remove `@dev` entirely? In the end the function returns the admin address whatever it is.

utils/src/tests/BaseScript.sol
smol-ninja: OK. The comptroller still returns the old admin address so I could not verify but looks good now.

utils/src/tests/BaseScript.sol
andreivladbrg: 86f6321fc4db4e9a2f63e7a850771bc0b15b2251

utils/src/tests/BaseScript.sol
andreivladbrg: 86f6321fc4db4e9a2f63e7a850771bc0b15b2251

utils/tests/integration/fuzz/base-script/BaseScript.t.sol
andreivladbrg: 86f6321fc4db4e9a2f63e7a850771bc0b15b2251

smol-ninja (approved):

utils/src/tests/BaseScript.sol
andreivladbrg: thats the case for other chains as well (e.g. arbitrum)

Issue #1358, opened 2025-12-08
Implement claim via external attestation (to enable KYC, Gitcoin Passport etc.)


smol-ninja:
<sup>Originally discussed in https://github.com/sablier-labs/command-center/discussions/227</sup>

### Feature

Implement a new claim function that accepts an EIP-712 typed signature signed by an "attestor" address.

1. Use similar logic as `claimViaSig`, the function extracts signer from the signed data.
2. It verifies that the signer matches the attestor address defined in the Comptroller contract.
3. Once verification has been passed, it executes the logic similar to `claim` function and transfers the allocation amount to the `recipient`

### Implementation

A new function in the Merkle contract:

```solidity
function claimViaAttestation(
    uint256 index,
    address recipient,
    uint128 amount,
    bytes32[] calldata merkleProof,
    bytes calldata attestation
)
    external
    payable;
```

where the `attestation` should be EIP-712 typed signature generated using the following JSON object:

```js
types: {
  EIP712Domain: [
    { name: "name", type: "string" },
    { name: "chainId", type: "uint256" },
    { name: "verifyingContract", type: "address" },
  ],
  Identity: [
    { name: "recipient", type: "address" },
  ],
},
domain: {
  name: "Sablier Airdrops Protocol",
  chainId: 1, // Chain on which the contract is deployed
  verifyingContract: "0xTheAddressOfThisContract", // The address of this contract
},
primaryType: "Identity",
message: {
  recipient: "0xTheAddressOfTheRecipient", // The address of the airdrop recipient
},
```

### Changes to the Comptroller

The "attestor" address must be queried from the Comptroller. This will allow us to change the attestor address in case it gets compromised.

The comptroller will have the following new variables and functions:

```solidity
address public attestor;

function setAttestor(address newAddress) onlyRole;
```

### QnA

**Q:** Why is attestor not set as an immutable variable inside the Merkle contract?
**A:** Due to the off-chain trust, an immutable signer can be misused to illegally attest identities if the signing server gets compromised.

### Note

If you do not like the names used in this issue, feel free to suggest new ones.

### Similar issues

- https://github.com/sablier-labs/lockup/issues/1356

andreivladbrg:
@smol-ninja shouldn't this task also include the enum of "allowed" claim functions?

**later edit:** nvm, just saw https://github.com/sablier-labs/lockup/issues/1359

andreivladbrg:
[...]

PR #1403, opened 2026-01-29
feat(airdrops): claim via attestation

andreivladbrg:
Closes https://github.com/sablier-labs/lockup/issues/1358

Q: add examples in `claimViaAttestation`, or keep the note about `claimViaSig` only?

---

note: if the attestor is changed, and a signature was created before that, the signature wouldn’t be valid anymore.
note: with the current identity hash, once a recipient has completed kyc, they can claim multiple airdrops (if eligible) with the same signature (intended)

note: due to more params in the merkle constructor, i faced issues with the "stack-too-deep" error - so, i added the new ConstructorParams struct for `MerkleBase` and `MerkleLockup`

### DIFF
--- a/airdrops/src/SablierMerkleExecute.sol
+++ b/airdrops/src/SablierMerkleExecute.sol
@@ -7,7 +7,7 @@ import { ReentrancyGuard } from "@openzeppelin/contracts/utils/ReentrancyGuard.s
 
 import { SablierMerkleBase } from "./abstracts/SablierMerkleBase.sol";
 import { ISablierMerkleExecute } from "./interfaces/ISablierMerkleExecute.sol";
-import { MerkleExecute } from "./types/DataTypes.sol";
+import { MerkleBase, MerkleExecute } from "./types/DataTypes.sol";
 
 /*
 
@@ -56,17 +56,17 @@ contract SablierMerkleExecute is
         address campaignCreator,
         address comptroller
     )
-        SablierMerkleBase(
-            campaignCreator,
-            campaignParams.campaignName,
-            campaignParams.campaignStartTime,
-            comptroller,
-            campaignParams.expiration,
-            campaignParams.initialAdmin,
-            campaignParams.ipfsCID,
-            campaignParams.merkleRoot,
-            campaignParams.token
-        )
+        SablierMerkleBase(MerkleBase.ConstructorParams({
+                campaignCreator: campaignCreator,
+                campaignName: campaignParams.campaignName,
+                campaignStartTime: campaignParams.campaignStartTime,
+                comptroller: comptroller,
+                expiration: campaignParams.expiration,
+                initialAdmin: campaignParams.initialAdmin,
+                ipfsCID: campaignParams.ipfsCID,
+                merkleRoot: campaignParams.merkleRoot,
+                token: campaignParams.token
+            }))
     {
         // Effect: set the immutable state variables.
         SELECTOR = campaignParams.selector;
--- a/airdrops/src/SablierMerkleInstant.sol
+++ b/airdrops/src/SablierMerkleInstant.sol
@@ -5,8 +5,9 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
 import { SablierMerkleBase } from "./abstracts/SablierMerkleBase.sol";
+import { SablierMerkleSignature } from "./abstracts/SablierMerkleSignature.sol";
 import { ISablierMerkleInstant } from "./interfaces/ISablierMerkleInstant.sol";
-import { MerkleInstant } from "./types/DataTypes.sol";
+import { MerkleBase, MerkleInstant } from "./types/DataTypes.sol";
 
 /*
 
@@ -29,8 +30,8 @@ import { MerkleInstant } from "./types/DataTypes.sol";
 /// @title SablierMerkleInstant
 /// @notice See the documentation in {ISablierMerkleInstant}.
 contract SablierMerkleInstant is
-    ISablierMerkleInstant, // 2 inherited components
-    SablierMerkleBase // 3 inherited components
+    ISablierMerkleInstant, // 3 inherited components
+    SablierMerkleSignature // 5 inherited components
 {
     using SafeERC20 for IERC20;
 
@@ -44,17 +45,18 @@ contract SablierMerkleInstant is
         address campaignCreator,
         address comptroller
     )
-        SablierMerkleBase(
-            campaignCreator,
-            campaignParams.campaignName,
-            campaignParams.campaignStartTime,
-            comptroller,
-            campaignParams.expiration,
-            campaignParams.initialAdmin,
-            campaignParams.ipfsCID,
-            campaignParams.merkleRoot,
-            campaignParams.token
-        )
+        SablierMerkleBase(MerkleBase.ConstructorParams({
+                campaignCreator: campaignCreator,
+                campaignName: campaignParams.campaignName,
+                campaignStartTime: campaignParams.campaignStartTime,
+                comptroller: comptroller,
+                expiration: campaignParams.expiration,
+                initialAdmin: campaignParams.initialAdmin,
+                ipfsCID: campaignParams.ipfsCID,
+                merkleRoot: campaignParams.merkleRoot,
+                token: campaignParams.token
+            }))
+        SablierMerkleSignature()
     { }
 
     /*//////////////////////////////////////////////////////////////////////////
@@ -99,6 +101,29 @@ contract SablierMerkleInstant is
     }
 
     /// @inheritdoc ISablierMerkleInstant
+    function claimViaAttestation(
+        uint256 index,
+        address to,
+        uint128 amount,
+        bytes32[] calldata merkleProof,
+        bytes calldata attestation
+    )
+        external
+        payable
+        override
+        notZeroAddress(to)
+    {
+        // Check: the attestation signature is valid and the recovered signer matches the attestor.
+        _verifyAttestationSignature(msg.sender, attestation);
+
+        // Check, Effect and Interaction: Pre-process the claim parameters on behalf of `msg.sender`.
+        _preProcessClaim({ index: index, recipient: msg.sender, amount: amount, merkleProof: merkleProof });
+
+        // Interaction: Post-process the claim parameters on behalf of `msg.sender`.
+        _postProcessClaim({ index: index, recipient: msg.sender, to: to, amount: amount, viaSig: false });
+    }
+
+    /// @inheritdoc ISablierMerkleInstant
     function claimViaSig(
         uint256 index,
         address recipient,
@@ -114,7 +139,7 @@ contract SablierMerkleInstant is
         notZeroAddress(to)
     {
         // Check: the signature is valid and the recovered signer matches the recipient.
-        _checkSignature(index, recipient, to, amount, validFrom, signature);
+        _verifyClaimSignature(index, recipient, to, amount, validFrom, signature);
 
         // Check, Effect and Interaction: Pre-process the claim parameters on behalf of the recipient.
         _preProcessClaim(index, recipient, amount, merkleProof);
--- a/airdrops/src/SablierMerkleInstant.sol
+++ b/airdrops/src/SablierMerkleInstant.sol
@@ -56,7 +56,6 @@ contract SablierMerkleInstant is
                 merkleRoot: campaignParams.merkleRoot,
                 token: campaignParams.token
             }))
-        SablierMerkleSignature()
     { }
 
     /*//////////////////////////////////////////////////////////////////////////
--- a/airdrops/src/abstracts/SablierMerkleLockup.sol
+++ b/airdrops/src/abstracts/SablierMerkleLockup.sol
@@ -5,13 +5,14 @@ import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { ISablierLockup } from "@sablier/lockup/src/interfaces/ISablierLockup.sol";
 
-import { ISablierMerkleLockup } from "../interfaces/ISablierMerkleLL.sol";
+import { ISablierMerkleLockup } from "../interfaces/ISablierMerkleLockup.sol";
+import { MerkleLockup } from "../types/DataTypes.sol";
 import { SablierMerkleBase } from "./SablierMerkleBase.sol";
 
 /// @title SablierMerkleLockup
 /// @notice See the documentation in {ISablierMerkleLockup}.
 abstract contract SablierMerkleLockup is
-    ISablierMerkleLockup, // 2 inherited components,
+    ISablierMerkleLockup, // 2 inherited components
     SablierMerkleBase // 3 inherited components
 {
     using SafeERC20 for IERC20;
@@ -40,37 +41,11 @@ abstract contract SablierMerkleLockup is
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Constructs the contract by initializing the immutable state vars, and max approving the Lockup contract.
-    constructor(
-        address campaignCreator,
-        string memory campaignName,
-        uint40 campaignStartTime,
-        bool cancelable,
-        address comptroller,
-        ISablierLockup sablierLockup,
-        uint40 expiration,
-        address initialAdmin,
-        string memory ipfsCID,
-        bytes32 merkleRoot,
-        string memory shape_,
-        IERC20 token,
-        bool transferable
-    )
-        SablierMerkleBase(
-            campaignCreator,
-            campaignName,
-            campaignStartTime,
-            comptroller,
-            expiration,
-            initialAdmin,
-            ipfsCID,
-            merkleRoot,
-            token
-        )
-    {
-        SABLIER_LOCKUP = sablierLockup;
-        streamShape = shape_;
-        STREAM_CANCELABLE = cancelable;
-        STREAM_TRANSFERABLE = transferable;
+    constructor(MerkleLockup.ConstructorParams memory lockupParams) {
+        SABLIER_LOCKUP = lockupParams.lockup;
+        STREAM_CANCELABLE = lockupParams.cancelable;
+        STREAM_TRANSFERABLE = lockupParams.transferable;
+        streamShape = lockupParams.shape;
 
         // Max approve the Lockup contract to spend funds from the Merkle Lockup campaigns.
         TOKEN.forceApprove({ spender: address(SABLIER_LOCKUP), value: type(uint256).max });
--- a/airdrops/src/abstracts/SablierMerkleSignature.sol
+++ b/airdrops/src/abstracts/SablierMerkleSignature.sol
@@ -1,0 +1,183 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity >=0.8.22;
+
+import { MessageHashUtils } from "@openzeppelin/contracts/utils/cryptography/MessageHashUtils.sol";
+import { SignatureChecker } from "@openzeppelin/contracts/utils/cryptography/SignatureChecker.sol";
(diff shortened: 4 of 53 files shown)


andreivladbrg:
@smol-ninja tentatively requested a review, there are a couple of questions to be answered first, please see the OP

smol-ninja:
Thanks for the PR. I will review the contracts tomorrow after I finish with price gated lockup today.

> backend has signed the attestation for a recipient, it means that the airdrop can be claimed by anyone

Yeah good point. Also, since attestation API would be exposed in the UI, anybody will be able to check KYC and generated signed data for any user.

> imagine a recipient just doing the KYC to check if it's eligible for it - we'd have to sign it onchain only when the recipient actually wants to claim

What do you mean by "signing onchain"? The signature will be signed offchain by the attestor private key. 

> wdyt about dropping this for MerkleVCA

Yeah I agree. 

> add a validFrom check in claimViaAttestation too?

If we do not support merkle VCA then why would we need it?

> if the attestor is changed, and a signature was created before that, the signature wouldn’t be valid anymore

Should it be added as an immutable to every campaign?

> with the current identity hash, once a recipient has completed kyc, they can claim multiple airdrops (if eligible) with the same signature

I see you used a simple struct with only `recipient` param. Why are we not using the same approach as `claimViaSig`? In the end, claim with attestation is same as claim via sign with the only difference of signer. 

Also, because EIP 712 (see struct in the issue OP) has `verifyingContract` which is merkle campaign itself, the signature cannot be used across campaigns.

andreivladbrg:
> What do you mean by "signing onchain"? The signature will be signed offchain by the attestor private key.

correct, not the best wording xD
my point is that the signing should be done _just_ before claiming (onchain) and not earlier - which is a hard constraint


> Yeah I agree.

great, removed it in 49562a5430a3ffb3064d2da5cc420a1737265a33

> Should it be added as an immutable to every campaign?

the current design i think it's better and it offers greater flexibility with

> I see you used a simple struct with only `recipient` param. Why are we not using the same approach as `claimViaSig`? In the end, claim with attestation is same as claim via sign with the only difference of signer.

my rationale was that once a user did KYC for an individual address, then it can use this KYC for all airdrops. in case there are multiple indexes with their address, otherwise there must be a signature for each individual claim. i don't see a security risk
wdyt?

> Also, because EIP 712 (see struct in the issue OP) has verifyingContract which is merkle campaign itself, the signature cannot be used across campaigns

i am not sure what you mean
the attestation uses the `_domainSeparator`, which uses `address(this)`, i.e., `verifyingContract`

airdrops/src/abstracts/SablierFactoryMerkleBase.sol
```diff
@@ -16,20 +16,37 @@ abstract contract SablierFactoryMerkleBase is
...
+    constructor(address initialAttestor, address initialComptroller) Comptrollerable(initialComptroller) {
+        attestor = initialAttestor;
+    }
 
     /*//////////////////////////////////////////////////////////////////////////
                         USER-FACING STATE-CHANGING FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @inheritdoc ISablierFactoryMerkleBase
+    function setAttestor(address newAttestor) external override onlyComptroller {
```
smol-ninja: What if we move it to the comptroller? It would be a PITA to update it across so many contracts across so many chains?

Both `attestor` and `setAttestor` can be moved to comptroller.

airdrops/src/abstracts/SablierFactoryMerkleBase.sol
```diff
@@ -16,20 +16,37 @@ abstract contract SablierFactoryMerkleBase is
...
 
     /*//////////////////////////////////////////////////////////////////////////
                                     CONSTRUCTOR
     //////////////////////////////////////////////////////////////////////////*/
 
+    /// @param initialAttestor The address of the initial attestor.
     /// @param initialComptroller The address of the initial comptroller contract.
-    constructor(address initialComptroller) Comptrollerable(initialComptroller) { }
+    constructor(address initialAttestor, address initialComptroller) Comptrollerable(initialComptroller) {
+        attestor = initialAttestor;
```
smol-ninja: Q: After the PR that allows user to choose between `default` and `claimViaAttestation`, will it be set only if user chooses `claimViaAttestation`?

airdrops/src/abstracts/SablierFactoryMerkleBase.sol
andreivladbrg: fair point, it would indeed not be so friendly to update it across so many campaigns and factories
i agree to move it in the comptroller, so wdys about this:
- remove the `attestor` from factories
- default to using the `comptroller.attestor`
  - don't remove the address in the campaigns, but still allow campaign creators to voluntarily change it
- if the address is set by the admin in a certain campaign, then don't use the comptroller one

airdrops/src/abstracts/SablierFactoryMerkleBase.sol
andreivladbrg: no, it would still require to pass it

airdrops/src/abstracts/SablierFactoryMerkleBase.sol
smol-ninja: Yes agree. Campaign will still have variable to track attestor and can be changed at the campaign level.

smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/7742b42549a3012c89aa559c7b93330893927713

airdrops/src/interfaces/ISablierMerkleInstant.sol
```diff
@@ -54,6 +54,34 @@ interface ISablierMerkleInstant is ISablierMerkleBase {
...
+    /// - The attestation signature must be valid.
+    /// - Refer to the requirements in {claim}.
+    ///
+    /// @param index The index of the `msg.sender` in the Merkle tree.
+    /// @param amount The amount of ERC-20 tokens allocated to the `msg.sender`.
+    /// @param merkleProof The proof of inclusion in the Merkle tree.
+    /// @param attestation The EIP-712 signature from the attestor.
+    function claimViaAttestation(
+        uint256 index,
+        uint128 amount,
```
smol-ninja, line 76: It should have `to` parameter same as `claimTo` and `claimViaSig`

airdrops/src/abstracts/SablierMerkleSignature.sol
```diff
@@ -0,0 +1,188 @@
...
+import { ISablierComptroller } from "@sablier/evm-utils/src/interfaces/ISablierComptroller.sol";
+
+import { ISablierMerkleSignature } from "./../interfaces/ISablierMerkleSignature.sol";
+import { Errors } from "./../libraries/Errors.sol";
+import { SignatureHash } from "./../libraries/SignatureHash.sol";
+import { MerkleBase } from "./../types/DataTypes.sol";
+import { SablierMerkleBase } from "./SablierMerkleBase.sol";
+
+/// @title SablierMerkleSignature
+/// @notice Abstract contract providing EIP-712 signature verification for Merkle campaigns.
```
smol-ninja:
```suggestion
/// @notice See the documentation in {ISablierMerkleSignature}.
```

airdrops/src/SablierMerkleExecute.sol
```diff
@@ -57,15 +57,17 @@ contract SablierMerkleExecute is
         address comptroller
     )
         SablierMerkleBase(
+            MerkleBase.ConstructorParams({
```
smol-ninja: Since `MerkleBase.ConstructorParams` is only for internal use, lets move `campaignCreator` and `comptroller` to it as well.

airdrops/src/SablierMerkleInstant.sol
```diff
@@ -44,16 +44,18 @@ contract SablierMerkleInstant is
         address campaignCreator,
         address comptroller
     )
-        SablierMerkleBase(
+        SablierMerkleSignature(
+            MerkleBase.ConstructorParams({
```
smol-ninja: We should refactor this as the following:

```solidity
SablierMerkleBase(MerkleBase.ConstructorParams({
        campaignCreator: campaignCreator,
        campaignName: campaignParams.campaignName,
        campaignStartTime: campaignParams.campaignStartTime,
        comptroller: comptroller,
        expiration: campaignParams.expiration,
        initialAdmin: campaignParams.initialAdmin,
        ipfsCID: campaignParams.ipfsCID,
        merkleRoot: campaignParams.merkleRoot,
        token: campaignParams.token
    }))
SablierMerkleSignature()
```

So we only add constructor params to the contract that requires it. It also makes it cleaner.

airdrops/src/abstracts/SablierMerkleLockup.sol
```diff
@@ -41,36 +42,17 @@ abstract contract SablierMerkleLockup is
...
-            campaignName,
-            campaignStartTime,
-            comptroller,
-            expiration,
-            initialAdmin,
-            ipfsCID,
-            merkleRoot,
-            token
-        )
+        SablierMerkleSignature(baseParams, campaignCreator, comptroller)
```
smol-ninja: We can remove it from this contract and it would be become:

```solidity
constructor(MerkleLockup.ConstructorParams memory lockupParams) {
```

airdrops/src/abstracts/SablierMerkleSignature.sol
```diff
@@ -0,0 +1,188 @@
...
+                                    CONSTRUCTOR
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Constructs the contract by initializing the immutable state variables.
+    constructor(
+        MerkleBase.ConstructorParams memory baseParams,
+        address campaignCreator,
+        address comptroller
+    )
+        SablierMerkleBase(baseParams, campaignCreator, comptroller)
```
smol-ninja: We can refactor this into

```solidity
constructor() {
```

airdrops/src/abstracts/SablierMerkleSignature.sol
```diff
@@ -0,0 +1,188 @@
...
+            SignatureChecker.isValidSignatureNow({ signer: signer, hash: digest, signature: signature });
+
+        // Check: `isSignatureValid` is true.
+        if (!isSignatureValid) {
+            revert Errors.SablierMerkleSignature_InvalidSignature();
+        }
+    }
+
+    /// @dev Returns the domain separator for the current chain.
+    function _domainSeparator() private view returns (bytes32) {
```
smol-ninja: Alphabetical order is not respected.

airdrops/tests/utils/Types.sol
```diff
@@ -12,6 +12,11 @@ struct Claim {
     uint40 validFrom;
 }
 
+/// @dev Struct to hold the Identity message parameters during ERC-712 attestation tests.
+struct Identity {
```
smol-ninja: Should be below `EIP712Domain`

airdrops/tests/integration/Integration.t.sol
```diff
@@ -91,6 +91,38 @@ abstract contract Integration_Test is Base_Test {
         ISablierMerkleInstant(address(merkleBase)).claimTo{ value: msgValue }(index, to, amount, merkleProof);
     }
 
+    /// @dev Claim using default values for {claimViaAttestation} function.
+    function claimViaAttestation() internal {
+        claimViaAttestation(users.recipient, CLAIM_AMOUNT);
+    }
+
+    /// @dev Claim using recipient and amount parameters for {claimViaAttestation} function.
+    function claimViaAttestation(address recipient, uint128 amount) internal {
```
smol-ninja: Unused helper.

airdrops/tests/integration/Integration.t.sol
```diff
@@ -122,14 +154,21 @@ abstract contract Integration_Test is Base_Test {
     )
         internal
     {
-        // Using `ISablierMerkleInstant` interface over `merkleBase` works for all Merkle contracts due to similarity in
```
smol-ninja, line 143: Why removed it?

airdrops/tests/integration/concrete/campaign/shared/claim-via-attestation/claimViaAttestation.tree
```diff
@@ -0,0 +1,18 @@
+ClaimViaAttestation_Integration_Test
+├── given attestor not set
```
smol-ninja: This is same as `given attestor zero`

airdrops/tests/integration/concrete/campaign/shared/claim-via-attestation/claimViaAttestation.t.sol
```diff
@@ -0,0 +1,86 @@
...
+        // Change caller back to the recipient.
+        setMsgSender(users.recipient);
+
+        vm.expectRevert(Errors.SablierMerkleSignature_AttestorNotSet.selector);
+        claimViaAttestation({
+            msgValue: AIRDROP_MIN_FEE_WEI,
+            index: getIndexInMerkleTree(),
+            amount: CLAIM_AMOUNT,
+            merkleProof: getMerkleProof(),
+            attestation: abi.encode(0)
```
smol-ninja: This is a test for zero attestor address so this should be `generateAttestation()`

airdrops/tests/integration/concrete/campaign/shared/claim-via-attestation/claimViaAttestation.t.sol
```diff
@@ -0,0 +1,86 @@
...
+        });
+    }
+
+    function test_RevertWhen_AttestationInvalid() external givenAttestorSet givenAttestorIsEOA {
+        // Generate an invalid attestation.
+        bytes memory invalidAttestation = vm.randomBytes(65);
```
smol-ninja: Since there is no test to verify what happens when one user uses signature generated for another user (also an invalid attestation), we should turn this into that.

smol-ninja (approved):
Leaving my comments below. I have also addressed most of these changes in my latest commits. 

Once you have reviewed those commits, this PR can be merged.

airdrops/src/interfaces/ISablierMerkleInstant.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/abstracts/SablierMerkleSignature.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/SablierMerkleExecute.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/SablierMerkleInstant.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/abstracts/SablierMerkleLockup.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/abstracts/SablierMerkleSignature.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/abstracts/SablierMerkleSignature.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/tests/utils/Types.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/tests/integration/Integration.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/tests/integration/concrete/campaign/shared/claim-via-attestation/claimViaAttestation.tree
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/tests/integration/concrete/campaign/shared/claim-via-attestation/claimViaAttestation.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/tests/integration/concrete/campaign/shared/claim-via-attestation/claimViaAttestation.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/e2ff86c0625c40fe7078a49c9aae4a70f229d915

airdrops/src/SablierMerkleInstant.sol
andreivladbrg: very good idea

airdrops/tests/integration/Integration.t.sol
andreivladbrg: no particular reason, i believe claude did it

airdrops/src/interfaces/ISablierMerkleInstant.sol
andreivladbrg: right

airdrops/tests/integration/Integration.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1403/changes/ac7089ab2bb31785658c9611453f489a548372be

utils/src/SablierComptroller.sol
```diff
@@ -60,10 +60,14 @@ contract SablierComptroller is
     /// @dev A mapping of protocol fees.
     mapping(Protocol protocol => ProtocolFees fees) private _protocolFees;
 
+    /// @inheritdoc ISablierComptroller
+    address public override attestor;
```
andreivladbrg, line 64: @smol-ninja 
is there a reason why not include it in the `initialize` function?

https://github.com/sablier-labs/lockup/blob/feat/claim-attestation/utils/src/SablierComptroller.sol#L104-L110

smol-ninja: We will probably not have an `attestor` setup in the beginning and thats why I didn't add it to `initialize` since it would be just zero address

andreivladbrg: fair enough

andreivladbrg (commented):
since we have `claimViaAttestation` in VCA as well, should we add the `validFrom` in the attestation sig?

smol-ninja:
> since we have `claimViaAttestation` in VCA as well, should we add the `validFrom` in the attestation sig?

It does not matter because the reason it was added to `claimViaSig` is that user A can use signature signed by user B and claim for user B. So having `validFrom` guarantees that such signature can only be used after a certain timestamp.

On the other hand, such cross claiming is not possible in `claimViaAttestation`. A user can only claim for himself.

So this function is not so different from claim or claimTo.

andreivladbrg:
> A user can only claim for himself.

right, good point

andreivladbrg:
thanks for the commits, merging now

Issue #1378, opened 2026-01-12
Implement Trade Escrow


smol-ninja:
As part of the [price gated Lockup contract](https://github.com/sablier-labs/lockup/issues/1280), implement a function to enable P2P swap between two users.

### Solution

User A created an order with he following parameters:

1.  The address of the counterparty
2. The address of token A that user A wants to sell
3. The amount of token A that user A wants to sell
4. Type of token A (ERC20 vs ERC721)
5. The address of the token B that user A wants to receive
6. The amount of token B that user wants to receive in exchange for the amount of token A
7. Type of token B (ERC20 vs ERC721)
8. Order expiry

When user B settles this trade, the requires amount of token B will be transferred to user A with user B receiving token A.

### References

1. [Motivation](https://github.com/sablier-labs/command-center/discussions/155#discussioncomment-15444587)

PR #1404, opened 2026-01-29
feat: add Sablier Bob and Sablier Escrow contracts

smol-ninja:
## Changelog
- Closes https://github.com/sablier-labs/lockup/issues/1378
- Adds `SablierBob` contract (see [diagram](https://github.com/sablier-labs/lockup/blob/feat/price-gated-lockup/bob/DIAGRAMS.md))
- Adds `withdrawERC20Token` function in Comptroller to withdraw ERC20

## References
- https://docs.lido.fi/guides/lido-tokens-integration-guide/

### Escrow
- [x] Integration tests
- [ ] Fuzz tests
- [ ] Fork tests
- [ ] Invariant tests
- [x] Deployment scripts
- [ ] Docs

### Bob
- [ ] Integration tests
- [ ] Fuzz tests
- [ ] Fork tests
- [ ] Invariant tests
- [x] Deployment scripts
- [ ] Docs


bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+
+    /// @inheritdoc ISablierBob
+    function createVault(
+        IERC20 token,
+        AggregatorV3Interface oracle,
+        uint40 expiry,
+        uint128 targetPrice
+    )
+        external
+        override
```
andreivladbrg, line 100: why no `nonReentrant` for this function?

bob/src/interfaces/ISablierBob.sol
```diff
@@ -0,0 +1,237 @@
...
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IBatch } from "@sablier/evm-utils/src/interfaces/IBatch.sol";
+import { IComptrollerable } from "@sablier/evm-utils/src/interfaces/IComptrollerable.sol";
+
+import { IBobVaultShare } from "./IBobVaultShare.sol";
+import { ISablierBobAdapter } from "./ISablierBobAdapter.sol";
+import { ISablierBobState } from "./ISablierBobState.sol";
+
+/// @title ISablierBob
+/// @notice Interface for the Sablier Bob protocol.
```
andreivladbrg: should we add more details about what Bob protocol is?

bob/src/interfaces/ISablierBob.sol
```diff
@@ -0,0 +1,237 @@
...
+
+    /// @notice Creates a new vault with the specified parameters.
+    ///
+    /// @dev Emits a {CreateVault} event.
+    ///
+    /// Notes:
+    /// - A new ERC-20 share token is deployed for each vault to represent user's share of deposits in the vault.
+    /// - The default adapter for the token is copied as the vault adapter. Any change in the default adapter does not
+    /// affect existing vaults.
+    /// - Vault creator is responsible for choosing a valid oracle. Using Chainlink oracles are highly recommended.
```
andreivladbrg: > Using Chainlink oracles are highly recommended.

given we use `latestRoundData()` and we expect to return its specific values: `uint80, int256 price, uint256, uint256, uint80` 
this should be a "constraint" to filter valid vaults

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+
+    /*//////////////////////////////////////////////////////////////////////////
+                         INTERNAL STATE-CHANGING FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Internal function to fetch the latest oracle price and update it in the vault storage.
+    /// @param vaultId The ID of the vault.
+    /// @param vault The vault storage reference.
+    /// @return latestPrice The latest price from the oracle.
+    function _syncPriceFromOracle(uint256 vaultId, Bob.Vault storage vault) internal returns (uint128 latestPrice) {
```
andreivladbrg: only the `vaultId` param is needed in this function

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+        // Load the vault ID from storage.
+        vaultId = nextVaultId;
+
+        // Effect: bump the next vault ID.
+        unchecked {
+            nextVaultId = vaultId + 1;
+        }
+
+        // Retrieve token symbol and token decimal.
+        string memory tokenSymbol = IERC20Metadata(address(token)).symbol();
```
andreivladbrg: we need a "safe" function for this like in the lockup nft descriptor:

https://github.com/sablier-labs/lockup/blob/2ea2dbe6f66d4c0ccbe486d7cccb35b6330953c4/lockup/src/LockupNFTDescriptor.sol#L331-L354

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+        uint8 tokenDecimals = IERC20Metadata(address(token)).decimals();
+
+        // Effect: deploy the share token for this vault.
+        IBobVaultShare shareToken = new BobVaultShare({
+            name_: string.concat("Sablier Bob ", tokenSymbol, " Vault #", vaultId.toString()),
+            symbol_: string.concat(
+                tokenSymbol,
+                "-",
+                uint256(targetPrice).toString(),
```
andreivladbrg, line 151: the target price will have 8 decimal format price, is this intended?

it won't be like: `"WETH-100-1792790393-1"`
but: `"WETH-100_00000000-1792790393-1"`

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+    function _validateOracle(AggregatorV3Interface oracle) private view returns (uint128 latestPrice) {
+        // Check: oracle is not the zero address.
+        if (address(oracle) == address(0)) {
+            revert Errors.SablierBob_InvalidOracle(address(oracle));
+        }
+
+        // Check: oracle returns 8 decimals when `decimals()` is called.
+        try oracle.decimals() returns (uint8 oracleDecimals) {
+            if (oracleDecimals != 8) {
+                revert Errors.SablierBob_InvalidOracleDecimals(address(oracle), oracleDecimals);
```
andreivladbrg: why not handle the case `decimals != 8` like we do in the comptroller:

https://github.com/sablier-labs/lockup/blob/2ea2dbe6f66d4c0ccbe486d7cccb35b6330953c4/utils/src/SablierComptroller.sol#L431-L439

instead of reverting?

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+    function redeem(uint256 vaultId)
+        external
+        payable
+        override
+        nonReentrant
+        notNull(vaultId)
+        returns (uint128 amountToTransfer, uint128 feeAmount)
+    {
+        // Check: the vault is settled.
+        if (_statusOf(vaultId) != Bob.Status.SETTLED) {
```
andreivladbrg: the `redeem` function does not sync the price

so if the price is reached and one wants to withdraw funds, they cannot, as they need to do it as a separate action

so, let's call `_syncPriceFromOracle`

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+            oracle: oracle,
+            adapter: adapter,
+            isStakedWithAdapter: true,
+            targetPrice: targetPrice,
+            lastSyncedPrice: latestPrice
+        });
+
+        // Interaction: register the vault with the adapter.
+        if (address(adapter) != address(0)) {
+            adapter.registerVault(vaultId);
```
andreivladbrg, line 180: i think it's not possible to have `address(adapter) == address(0)`, so this check is redundant

but if it's possible to have `address(adapter) == address(0)`, then this state in the vault is incorrect `isStakedWithAdapter: true`

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+        // Load the vault from storage.
+        Bob.Vault storage vault = _vaults[vaultId];
+
+        // Effect: sync the oracle price.
+        latestPrice = _syncPriceFromOracle(vaultId, vault);
+    }
+
+    /// @inheritdoc ISablierBob
+    function unstakeTokensViaAdapter(uint256 vaultId)
+        external
```
andreivladbrg, line 426: why do we need this external function? i think it's in our favor (to generate more fees) to not have a separate function for this, and only call it when first redeem occurs

bob/src/types/Bob.sol
```diff
@@ -0,0 +1,47 @@
...
+        // slot 0
+        IERC20 token;
+        uint40 expiry;
+        uint40 lastSyncedAt;
+        // slot 1
+        IBobVaultShare shareToken;
+        // slot 2
+        AggregatorV3Interface oracle;
+        // slot 3
+        ISablierBobAdapter adapter;
```
andreivladbrg, line 43: whats the benefit of storing this in each vault?

since we already save the token in the vault state and we have the: `defaultAdapter` wouldn't this be enough?

waste of gas IMO

bob/src/abstracts/SablierBobState.sol
```diff
@@ -0,0 +1,162 @@
...
+
+/// @title SablierBobState
+/// @notice See the documentation in {ISablierBobState}.
+abstract contract SablierBobState is ISablierBobState {
+    /*//////////////////////////////////////////////////////////////////////////
+                                  STATE VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierBobState
+    mapping(IERC20 token => ISablierBobAdapter adapter) public override defaultAdapter;
```
andreivladbrg: plural? `defaultAdapters`?

bob/src/BobVaultShare.sol
```diff
@@ -0,0 +1,110 @@
...
+    /// @param symbol_ The symbol of the token (e.g., "WETH-100-1792790393-1").
+    /// @param decimals_ The number of decimals.
+    /// @param sablierBob_ The address of the SablierBob contract.
+    /// @param vaultId_ The ID of the vault this share token represents.
+    constructor(
+        string memory name_,
+        string memory symbol_,
+        uint8 decimals_,
+        address sablierBob_,
+        uint256 vaultId_
```
andreivladbrg: no need for `_`

`sablierBob_` --> `sablierBob`
`vaultId_` --> `vaultId`

smol-ninja:
A `SafeOracle` library is added to `utils` package in https://github.com/sablier-labs/lockup/pull/1406. So we should use that now in Bob.

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+        if (oraclePrice <= 0) {
+            revert Errors.SablierBob_OraclePriceInvalid(vaultId, oraclePrice);
+        }
+
+        // Cast the oracle price to `uint128`.
+        latestPrice = uint256(oraclePrice).toUint128();
+
+        // Effect: update the last synced price and timestamp.
+        vault.lastSyncedPrice = latestPrice;
+        vault.lastSyncedAt = uint40(updatedAt);
```
andreivladbrg: why is the vault's last sync set to the oracle's last price update?

it should be `uint40(block.timestamp)` instead, because it's when the vault was updated, not the oracle (even tho the oracle price is used, there is no way to validate that the price is the correct one at time `t`)
 
if the oracle returns an incorrect time (in the future), then we would also have an incorrect vault time

bob/SECURITY.md
```diff
@@ -0,0 +1,25 @@
+# Security
+
+This document outlines security considerations, assumptions, and known limitations for protocols in this package.
+
+## Bug Bounty
+
+See the main repository [SECURITY.md](../SECURITY.md) for common assumptions and bug bounty details.
+
+## Protocol Assumptions - Sablier Bob
```
andreivladbrg, line 9: lets also add that if the vault is not synced at the right moment, the vault can become unredeemable if it was redeemable

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+        }
+
+        // Log the event.
+        emit CreateVault(vaultId, token, oracle, adapter, shareToken, targetPrice, expiry);
+    }
+
+    /// @inheritdoc ISablierBob
+    function enter(uint256 vaultId, uint128 amount) external override nonReentrant notNull(vaultId) {
+        // Check: the vault is not settled.
+        if (_statusOf(vaultId) == Bob.Status.SETTLED) {
```
andreivladbrg: if we first do this (check status and the sync the vault) it is possible to deposit into a settled pool

so we need to check the status before and after the sync

bob/src/interfaces/ISablierBob.sol
```diff
@@ -0,0 +1,237 @@
...
+    ///
+    /// Requirements:
+    /// - The vault must have ACTIVE status.
+    /// - The share balance of the caller must be greater than zero.
+    /// - Current timestamp must be within the grace period.
+    ///
+    /// @param vaultId The ID of the vault to exit from.
+    function exitWithinGracePeriod(uint256 vaultId) external;
+
+    /// @notice Redeem deposited tokens  from a settled vault by burning shares.
```
andreivladbrg: the user doesn't redeem only the deposited tokens - they can have shares that were directly transferred via `ERC20.transfer`

```suggestion
    /// @notice Redeem the full share balance from a settled vault by burning shares.
```

bob/src/types/Bob.sol
```diff
@@ -0,0 +1,47 @@
...
+
+import { IBobVaultShare } from "../interfaces/IBobVaultShare.sol";
+import { ISablierBobAdapter } from "../interfaces/ISablierBobAdapter.sol";
+
+/// @notice Namespace for the structs and enums used in the Sablier Bob protocol.
+library Bob {
+    /// @notice Enum representing the different statuses of a vault.
+    /// @custom:value0 ACTIVE Vault is open for deposits.
+    /// @custom:value1 SETTLED Vault is settled; either target price reached or vault has expired.
+    enum Status {
```
andreivladbrg, line 16: i feel like the status is incomplete

1. being SETTLED due to the target price being reached  
2. being SETTLED due to the vault expiry date

IMO these should be different statuses for a better API

also, the target price can be reached and the status doesn't have to be `SETTLED` if the sync function was not called

wdyt about adding more statuses for these cases?


```solidity
 enum Status {
      ACTIVE,             // Accepting deposits, price below target, not expired
      MATURED,            // Expired, awaiting final settlement
      SETTLED_BY_PRICE,   // Sync confirmed target hit (before or after expiry)
      SETTLED_BY_EXPIRY   // Expired + final sync confirmed price never reached target
  }
```

feel free to suggest better naming

bob/src/BobVaultShare.sol
```diff
@@ -0,0 +1,110 @@
...
+    function decimals() public view override(ERC20, IERC20Metadata) returns (uint8) {
+        return _DECIMALS;
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                        USER-FACING STATE-CHANGING FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc IBobVaultShare
+    function mint(address to, uint256 amount) external override onlySablierBob {
```
andreivladbrg: wdyt about checking the `vaultId` as well (beside the caller)? i believe from the share contract perspective, its safer

so:

```solidity
function mint(address to, uint256 amount, uint256 vaultId) external override onlySablierBob {
     if(vaultId != VAULT_ID) revert();
     // impl
}
```

the same for `burn`

bob/src/abstracts/SablierBobState.sol
```diff
@@ -0,0 +1,162 @@
...
+
+    /// @inheritdoc ISablierBobState
+    uint256 public override nextVaultId;
+
+    /// @dev Timestamp of first deposit for each user in each vault, used for grace period calculation.
+    mapping(uint256 vaultId => mapping(address user => uint40 depositedAt)) internal _depositedAt;
```
andreivladbrg: should we rename it to `firstDepositAt`?

since there can be more deposits

bob/src/SablierBob.sol
andreivladbrg: the first check is still relevant because it can go to settled to acitve status if the price goes lower

bob/src/types/Bob.sol
andreivladbrg: if we implement multiple status, we can have functions like isCold/isWarm in `lockup` (ofc different names)

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+            revert Errors.SablierBob_VaultSettled(vaultId);
+        }
+
+        // Check: the deposit amount is not zero.
+        if (amount == 0) {
+            revert Errors.SablierBob_DepositAmountZero(vaultId, msg.sender);
+        }
+
+        // Load the vault from storage.
+        Bob.Vault storage vault = _vaults[vaultId];
```
andreivladbrg: no need for `storage` here - vault's state is not modifed

bob/src/SablierBob.sol
```diff
@@ -0,0 +1,490 @@
...
+        uint256 vaultId,
+        address from,
+        address to,
+        uint256 amount,
+        uint256 fromBalanceBefore
+    )
+        external
+        override
+    {
+        Bob.Vault storage vault = _vaults[vaultId];
```
andreivladbrg: `storage` not needed

bob/src/BobVaultShare.sol
```diff
@@ -0,0 +1,110 @@
...
+
+    /*//////////////////////////////////////////////////////////////////////////
+                            INTERNAL OVERRIDE FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Override to notify SablierBob when shares are transferred (not minted/burned).
+    /// This allows the adapter to update wstETH attribution proportionally.
+    function _update(address from, address to, uint256 amount) internal override {
+        // Get sender's balance before the transfer.
+        uint256 fromBalanceBefore = balanceOf(from);
```
andreivladbrg: this can be moved in the `if` block below

bob/src/types/Escrow.sol
```diff
@@ -0,0 +1,49 @@
...
+    /// @param buyer The designated counterparty address specified by the seller. If its zero address, the order can be
+    /// filled by anyone.
+    /// @param sellToken The ERC-20 token being sold, deposited by the seller when the order is created.
+    /// @param buyToken The ERC-20 token the seller wants to receive.
+    /// @param sellAmount The amount of sell token that the seller is willing to exchange.
+    /// @param minBuyAmount The minimum amount of buy token that the seller is willing to accept.
+    /// @param expireAt The Unix timestamp when the order expires. Zero is sentinel for orders that never expire.
+    /// @param wasCanceled Boolean indicating if the order was canceled.
+    /// @param wasFilled Boolean indicating if the order was filled.
+    struct Order {
```
andreivladbrg, line 33: wouldn't it be useful to add cancelability feature?

bob/src/SablierLidoAdapter.sol
```diff
@@ -0,0 +1,394 @@
...
+    )
+        external
+        override
+        onlySablierBob
+    {
+        // Calculate proportional wstETH to transfer.
+        uint256 fromWstETH = _userWstETH[vaultId][from];
+
+        // Calculate the portion of wstETH to transfer.
+        uint128 wstETHToTransfer = (fromWstETH * shareAmountTransferred / userShareBalanceBeforeTransfer).toUint128();
```
andreivladbrg, line 352: `ERC20` allows zero amount transfers, so this line would revert in this case at `/ userShareBalanceBeforeTransfer`

bob/src/SablierEscrow.sol
```diff
@@ -0,0 +1,250 @@
...
+            // Set comptroller as the fee recipient.
+            address feeRecipient = address(comptroller);
+
+            // Interaction: transfer fees to the fee recipient.
+            buyToken.safeTransferFrom(msg.sender, feeRecipient, feeDeductedFromSellerAmount);
+            sellToken.safeTransfer(feeRecipient, feeDeductedFromBuyerAmount);
+        }
+
+        // Interaction: transfer buy token to the seller.
+        buyToken.safeTransferFrom(msg.sender, seller, amountToTransferToSeller);
```
andreivladbrg: this can be problematic in case the seller sets a buy token that is blacklisted for, thus the fill order will revert for the buyer

tho, not sure if there is something we can do

bob/src/interfaces/ISablierEscrow.sol
```diff
@@ -0,0 +1,120 @@
...
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { UD60x18 } from "@prb/math/src/UD60x18.sol";
+import { IBatch } from "@sablier/evm-utils/src/interfaces/IBatch.sol";
+import { IComptrollerable } from "@sablier/evm-utils/src/interfaces/IComptrollerable.sol";
+
+import { ISablierEscrowState } from "./ISablierEscrowState.sol";
+
+/// @title ISablierEscrow
+/// @notice Interface for the Sablier Escrow protocol.
```
andreivladbrg: more info here would be useful

bob/src/interfaces/ISablierEscrow.sol
```diff
@@ -0,0 +1,120 @@
...
+    /// Requirements:
+    /// - The order must exist.
+    /// - The order must be in OPEN status.
+    /// - If the order has buyer specified, the caller must be the buyer.
+    /// - `buyAmount` must be greater than or equal to the `minBuyAmount`.
+    /// - The caller must have approved this contract to transfer atleast `buyAmount` of `buyToken`.
+    ///
+    /// @param orderId The order ID to fill.
+    /// @param buyAmount The amount of buy token to exchange.
+    function fillOrder(uint256 orderId, uint128 buyAmount) external;
```
andreivladbrg: should we add some notes about the fees taken?

bob/src/SablierEscrow.sol
```diff
@@ -0,0 +1,250 @@
...
+╚════██║██╔══██║██╔══██╗██║     ██║██╔══╝  ██╔══██╗    ██╔══╝  ╚════██║██║     ██╔══██╗██║   ██║██║███╗██║
+███████║██║  ██║██████╔╝███████╗██║███████╗██║  ██║    ███████╗███████║╚██████╗██║  ██║╚██████╔╝╚███╔███╔╝
+╚══════╝╚═╝  ╚═╝╚═════╝ ╚══════╝╚═╝╚══════╝╚═╝  ╚═╝    ╚══════╝╚══════╝ ╚═════╝╚═╝  ╚═╝ ╚═════╝  ╚══╝╚══╝
+
+*/
+
+/// @title SablierEscrow
+/// @notice See the documentation in {ISablierEscrow}.
+contract SablierEscrow is
+    Batch, // 1 inherited component
```
andreivladbrg: do we really need batch here? i mean watch you gonna do, create an order and immediately cancel it?

bob/src/SablierEscrow.sol
```diff
@@ -0,0 +1,250 @@
...
+        // Check: the order is open.
+        {
+            Escrow.Status status = _statusOf(orderId);
+            if (status != Escrow.Status.OPEN) {
+                revert Errors.SablierEscrow_OrderNotOpen(orderId, status);
+            }
+        }
+
+        // Load the order from storage.
+        Escrow.Order storage order = _orders[orderId];
```
andreivladbrg:
```suggestion
        Escrow.Order memory order = _orders[orderId];
```

bob/src/SablierEscrow.sol
```diff
@@ -0,0 +1,250 @@
...
+
+    /// @inheritdoc ISablierEscrow
+    function fillOrder(uint256 orderId, uint128 buyAmount) external override notNull(orderId) {
+        // Check: the order is open.
+        {
+            Escrow.Status status = _statusOf(orderId);
+            if (status != Escrow.Status.OPEN) {
+                revert Errors.SablierEscrow_OrderNotOpen(orderId, status);
+            }
+        }
```
andreivladbrg, line 185: we can remove these

bob/src/types/Escrow.sol
```diff
@@ -0,0 +1,49 @@
...
+        // slot 0
+        address seller;
+        uint40 expireAt;
```
andreivladbrg:
```suggestion
        uint40 expireTime;
```

wdyt?

bob/src/SablierLidoAdapter.sol
```diff
@@ -0,0 +1,394 @@
...
+    /// @param sablierBob_ The address of the SablierBob contract.
+    /// @param initialSlippageTolerance The initial slippage tolerance for Curve swaps as UD60x18.
+    /// @param initialYieldFee The initial yield fee as UD60x18.
+    constructor(
+        address initialComptroller,
+        address sablierBob_,
+        address curvePool_,
+        address stETH_,
+        address wETH_,
+        address wstETH_,
```
andreivladbrg: these were previously constants, wouldn't the bob protocol *with adapter* be live only on mainnet?

these tokens & lido are deployed only to mainnet

andreivladbrg (commented):
great work Shub, i'm going to leave these review comments here

i want to mention two things
- some comments may be outdated, which we already settled on slack, and i missed removing them
- i implemented some changes already in this [PR](https://github.com/sablier-labs/lockup/pull/1412), but not all, as we still need to agree upon them

bob/src/SablierBob.sol
smol-ninja: Why do you need `nonReentrant` here? If a user uses nested calls to create multiple vaults, I don't see that as a problem.

bob/src/interfaces/ISablierBob.sol
smol-ninja: Sure we can add more details here however, they are already there in the README.

bob/src/interfaces/ISablierBob.sol
smol-ninja: I am not sure what are you suggesting here. We are advising users to use Chainlink oracles as much as possible.

bob/src/SablierBob.sol
andreivladbrg: yeah, we don't need it in create function

bob/src/SablierBob.sol
smol-ninja: Yeah good catch

bob/src/SablierBob.sol
smol-ninja: Do we? It will mess with the symbol of vault token contract. Why not just revert if the symbol does not exist? Is there any popular token that has no symbol?

bob/src/interfaces/ISablierBob.sol
andreivladbrg: my point was that instead of saying "highly recommended" to say something more assertive "use Chainlink"

bob/src/SablierBob.sol
smol-ninja: It will be like "WETH-10000000000-1792790393-1". Yes intended because if the price is $0.01 then we wont be able to capture that in the symbol name. Also imo its consistent to show the actual value here because thats what being returned by the functions too.

bob/src/SablierBob.sol
smol-ninja: Because then user has to be very careful when selecting the target price. By restricting it to follow 8 decimals format, we are creating a standard. If user requests higher decimals for price (which imo is very unlikely), we can consider adding it.

In comptroller, its necessary to handle all decimals because we can block user from claiming his token which is not the case here. Therefore, not handling all edge cases in Bob wouldn't hurt anyone.

bob/src/SablierBob.sol
smol-ninja: Agree

bob/src/SablierBob.sol
smol-ninja: > i think it's not possible to have `address(adapter) == address(0)`, so this check is redundant

Wdym? When adapter is not set, its `address(0)` hence the check.

> state in the vault is incorrect `isStakedWithAdapter: true`

Oh, good finding. It should be `false` if adapter is not set.

bob/src/SablierBob.sol
andreivladbrg: if the UI will use the name/symbol of the share contract in any place, ofc, remember the JS injection finding?

i don't see a problem adding it

bob/src/SablierBob.sol
smol-ninja: First redeem is going to be very expensive. So sometimes, if user complains, we can call this function and unstake all tokens and get our fees. hence the separate function. Thoughts?

bob/src/types/Bob.sol
smol-ninja: If we change the adapter for the token, all funds will be stuck in the old adapter hence its important from security pov to store each adapter address in the vault itself.

bob/src/SablierBob.sol
andreivladbrg: > It will be like "WETH-10000000000-1792790393-1". Yes intended because if the price is $0.01 then we wont be able to capture that in the symbol name. Also imo its consistent to show the actual value here because thats what being returned by the functions too.

what do you think about adding a "." or "," if the price is higher than $1?  i.e.  1000.00000000 --> $1000

bob/src/abstracts/SablierBobState.sol
smol-ninja: I prefer singular when a single value is returned. A mapping is a function that takes a value and returns a value. 

It should be called `defaultAdapter` if its returning a list which is not the case here.

bob/src/BobVaultShare.sol
smol-ninja: Sure

bob/src/SablierBob.sol
smol-ninja: Thats a very good point. And I agree.

bob/src/SablierBob.sol
andreivladbrg: fair enough, similar to the slack discussion about sync, resolving this

bob/SECURITY.md
smol-ninja: sounds good. I would suggest wordings to be very explicit as what is the right moment.

bob/src/SablierBob.sol
smol-ninja: Yeah sounds good.

bob/src/interfaces/ISablierBob.sol
smol-ninja: Good catch.

bob/src/types/Bob.sol
smol-ninja: What do you think about:

1. Active
2. Expired - End time reached but last synced price < target price
3. Matured - synced price > target price (regardless of end time)

There wont be a case where synced price > target price is achieved after end time is reached because we can revert in `syncPriceFromOracle` if end time is reached as it does not matter.

bob/src/BobVaultShare.sol
smol-ninja: Why check for `vaultId`? There is no delegatecall so how can tokens be minted otherwise?

bob/src/abstracts/SablierBobState.sol
smol-ninja: Sounds good

bob/src/SablierBob.sol
smol-ninja: Yup

bob/src/SablierBob.sol
smol-ninja: agree I guess

bob/src/types/Escrow.sol
smol-ninja: I guess it would be overkill. These are different from streams. Orders are rarely non-cancelable.

andreivladbrg: it's just an extra boolean check, not very complex logic

bob/src/SablierLidoAdapter.sol
smol-ninja: If user has 0 balance before transfer, it should revert.

bob/src/SablierEscrow.sol
smol-ninja: Thats OK. If one party is blacklisted for the token, the order wont be filled and then can be cancelled. I dont think we need to handle this on our end.

bob/src/interfaces/ISablierEscrow.sol
smol-ninja: Again its added to README and yeah we can add it here too

bob/src/interfaces/ISablierEscrow.sol
smol-ninja: What do you think if we return (net amount transferred, fee takes) in this function? I think that would be helpful for both integrations as well as allow us to add extra notes on fee.

bob/src/SablierEscrow.sol
smol-ninja: We can remove it

bob/src/SablierEscrow.sol
smol-ninja: I checked the gas usage, it looks like it as no impact on that because all slots are accessed during this function. Feel free to change but `storage` pointer looks better here as you wont have to use `_orders[orderId]`

bob/src/SablierEscrow.sol
smol-ninja: You will face stack too deep error with `just b`

bob/src/types/Escrow.sol
smol-ninja: I am fine with both. But I think it should be `expiryTime` right?

bob/src/SablierLidoAdapter.sol
smol-ninja: Can be deployed to other chains like sepolia or base. I realized it when deploying to Sepolia so I moved the addresses to deployment script. You can find them there. Looks good?

bob/src/interfaces/ISablierBob.sol
smol-ninja: Yeah that makes more sense.

bob/src/SablierBob.sol
smol-ninja: Lets add it because of the JS injection risk.

bob/src/SablierBob.sol
smol-ninja: Why? I feel like its too complicated. "," and "." or even decimals should be for the UIs. You will also have to handle the significant digits. For example of the target price is $4200.6718273, you will have to have both command and dot in the symbol name and include all significant digits. I believe you are thinking of simple target prices but it can be as complicated as $0.00000124 or $527000.123.

bob/src/types/Escrow.sol
smol-ninja: But whats the benefit of having non-cancelable orders? Non-cancelable streams make sense but non-cancelable orders? What can one achieve with that?

bob/src/SablierLidoAdapter.sol
andreivladbrg: don't you think it would  be better to have something more specific than `modulo zero division` revert?

bob/src/SablierEscrow.sol
andreivladbrg: fair enough

bob/src/interfaces/ISablierEscrow.sol
andreivladbrg: sounds good - will do in my PR

bob/src/SablierEscrow.sol
andreivladbrg: alright, will do in my PR

bob/src/types/Escrow.sol
andreivladbrg: > But I think it should be `expiryTime` right?

yeah

bob/src/SablierEscrow.sol
andreivladbrg: don't you think having state updates more readable like this? (btw, it's just one)

```solidity
_orders[orderId].wasFilled = true;
```

for me, it's a signal that if i see `_orders[orderId]` at the start of the line, it's a state change, and the rest, like `order.sellAmount`, are only reading from memory/storage

bob/src/SablierEscrow.sol
andreivladbrg: nope

bob/src/types/Escrow.sol
andreivladbrg: i am thinking that the escrow can be used as a "price-gated feature" as well: depositing 1 WETH and expecting to receive 5000 USDC

bob/src/SablierLidoAdapter.sol
smol-ninja: Sounds good. How about `UserBalanceZero`? Feel free to make the change in your PR.

bob/src/SablierEscrow.sol
smol-ninja: ok, lets change it to `memory`

bob/src/SablierEscrow.sol
smol-ninja: May be not after you PR. If thats the case, then we can remove it.
Screenshot 2026-02-17 at 12 38 01

bob/src/types/Escrow.sol
smol-ninja: example please. Because its hard to understand how it can be used as price gated.

bob/src/SablierEscrow.sol
andreivladbrg: > after you PR

correct

bob/src/SablierBob.sol
andreivladbrg: fair enough, lets keep the current version - will add some explanation in the constructor of the shares contract

bob/src/SablierBob.sol
andreivladbrg: > By restricting it to follow 8 decimals format, we are creating a standard.

fair enough, i agree

bob/src/BobVaultShare.sol
andreivladbrg: > Why check for `vaultId`? There is no delegatecall so how can tokens be minted otherwise?

from the `BobVaultShare` perspective, it's not guaranteed that it's called only for a certain vault - the vault that the shares contract was deployed for
the function check only who the caller is - which can be correct

bob/src/types/Bob.sol
andreivladbrg: > 1. Active
> 2. Expired - End time reached but last synced price < target price
> 3. Matured - synced price > target price (regardless of end time)

sounds good, but wdys about keeping the `SETTLED` term instead of  `MATURED`?

bob/src/BobVaultShare.sol
smol-ninja: Thats a valid point. Lets add it then.

bob/src/types/Bob.sol
smol-ninja: Because its settled in both the cases. I was looking for name that is as good as expired but to signal that synced price > target price but unfortunately I could not come up with any. I dont like matured too as matured also means the expired vault (matured in bonds mean the same thing). 

What about `TARGET_HIT`?

bob/scripts/bash/prepare-artifacts.sh
andreivladbrg, line 1: why `"$"`?

smol-ninja: Didn't verify if its true but code rabbit suggested this [here](https://github.com/sablier-labs/lockup/pull/1404#pullrequestreview-3825208079)

Screenshot 2026-02-19 at 11 50 53

andreivladbrg (approved):
merging for now

Issue #1359, opened 2025-12-08
Allow campaign creator to select from many claim functions when launching a campaign


smol-ninja:
After the next Airdrops release, each Merkle contract will have the following claim functions:

1. `claim`
2. `claimTo`
3. `claimViaSig`
4. `claimViaAttestation`: When enabled, it should disable all other claim functions.
5. `claimAndExecute`: When enabled, it should disable all other claim functions.

Since not all the above functions are complementary  to each other, the type of claim function to allow can be controlled using a new parameter in the signature of create merkle functions: `ClaimType claimType` where `ClaimType` can be declared as:

```solidity
enum ClaimType { // Feel free to suggest new variable names
    DEFAULT, // Activates claim, claimTo and claimViaSig
    ATTEST, // Activates claimViaAttestation
    Execute, // Activates claimAndExecute
}
```

The value of `ClaimType` should be stored as an immutable variable in the merkle contract.

Then in each claim function, verify the claim type. For example, `claim`, `claimTo`, and `claimViaSig` should check:

```solidity
if (CLAIM_TYPE != ClaimType.DEFAULT) revert();
```

and so on.

### Check

The size of Merkle contract does not exceed 24kB.

andreivladbrg:
> The size of Merkle contract does not exceed 24kB.

i believe the campaigns are far from having this issue, the factories would be (_if_) more problematic 

> Feel free to suggest new variable names

`ClaimType` would also work for me but here are some other alternatives:
- `ClaimAllowed`
- `ClaimMethod`
- `ClaimMode`

PaulRBerg:
I like `ClaimType`.

- `ClaimMethod` can be confused with meaning of "the claim method", i.e. "the claim function"
- `ClaimMode` is good but it can be confusing on [Mode chain](https://www.mode.network/)
- `ClaimAllowed` does not fit well

PR #1405, opened 2026-01-30
feat: claim-type

andreivladbrg:
Closes https://github.com/sablier-labs/lockup/issues/1359

Note: since we have implemented the `MerkleExecute` as a separate campaign the claim type enum include just `DEFAULT` and `ATTEST`

### DIFF
--- a/airdrops/src/SablierMerkleInstant.sol
+++ b/airdrops/src/SablierMerkleInstant.sol
@@ -49,6 +49,7 @@ contract SablierMerkleInstant is
                 campaignCreator: campaignCreator,
                 campaignName: campaignParams.campaignName,
                 campaignStartTime: campaignParams.campaignStartTime,
+                claimType: campaignParams.claimType,
                 comptroller: comptroller,
                 expiration: campaignParams.expiration,
                 initialAdmin: campaignParams.initialAdmin,
--- a/airdrops/src/SablierMerkleInstant.sol
+++ b/airdrops/src/SablierMerkleInstant.sol
@@ -72,7 +72,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
-        checkClaimType(ClaimType.DEFAULT)
+        revertIfNot(ClaimType.DEFAULT)
     {
         // Check, Effect and Interaction: Pre-process the claim parameters on behalf of the recipient.
         _preProcessClaim(index, recipient, amount, merkleProof);
@@ -91,7 +91,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
-        checkClaimType(ClaimType.DEFAULT)
+        revertIfNot(ClaimType.DEFAULT)
         notZeroAddress(to)
     {
         // Check, Effect and Interaction: Pre-process the claim parameters on behalf of `msg.sender`.
@@ -112,7 +112,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
-        checkClaimType(ClaimType.ATTEST)
+        revertIfNot(ClaimType.ATTEST)
         notZeroAddress(to)
     {
         // Check: the attestation signature is valid and the recovered signer matches the attestor.
@@ -138,7 +138,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
-        checkClaimType(ClaimType.DEFAULT)
+        revertIfNot(ClaimType.DEFAULT)
         notZeroAddress(to)
     {
         // Check: the signature is valid and the recovered signer matches the recipient.
--- a/airdrops/src/SablierMerkleInstant.sol
+++ b/airdrops/src/SablierMerkleInstant.sol
@@ -7,7 +7,7 @@ import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.s
 import { SablierMerkleBase } from "./abstracts/SablierMerkleBase.sol";
 import { SablierMerkleSignature } from "./abstracts/SablierMerkleSignature.sol";
 import { ISablierMerkleInstant } from "./interfaces/ISablierMerkleInstant.sol";
-import { MerkleBase, MerkleInstant } from "./types/DataTypes.sol";
+import { ClaimType, MerkleBase, MerkleInstant } from "./types/DataTypes.sol";
 
 /*
 
@@ -72,6 +72,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
+        checkClaimType(ClaimType.DEFAULT)
     {
         // Check, Effect and Interaction: Pre-process the claim parameters on behalf of the recipient.
         _preProcessClaim(index, recipient, amount, merkleProof);
@@ -90,6 +91,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
+        checkClaimType(ClaimType.DEFAULT)
         notZeroAddress(to)
     {
         // Check, Effect and Interaction: Pre-process the claim parameters on behalf of `msg.sender`.
@@ -110,6 +112,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
+        checkClaimType(ClaimType.ATTEST)
         notZeroAddress(to)
     {
         // Check: the attestation signature is valid and the recovered signer matches the attestor.
@@ -135,6 +138,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
+        checkClaimType(ClaimType.DEFAULT)
         notZeroAddress(to)
     {
         // Check: the signature is valid and the recovered signer matches the recipient.
--- a/airdrops/src/abstracts/SablierMerkleBase.sol
+++ b/airdrops/src/abstracts/SablierMerkleBase.sol
@@ -29,6 +29,9 @@ abstract contract SablierMerkleBase is
     uint40 public immutable override CAMPAIGN_START_TIME;
 
     /// @inheritdoc ISablierMerkleBase
+    ClaimType public immutable override CLAIM_TYPE;
+
+    /// @inheritdoc ISablierMerkleBase
     address public immutable override COMPTROLLER;
 
     /// @inheritdoc ISablierMerkleBase
@@ -45,9 +48,6 @@ abstract contract SablierMerkleBase is
 
     /// @inheritdoc ISablierMerkleBase
     string public override campaignName;
-
-    /// @inheritdoc ISablierMerkleBase
-    ClaimType public override claimType;
 
     /// @inheritdoc ISablierMerkleBase
     uint40 public override firstClaimTime;
@@ -75,10 +75,10 @@ abstract contract SablierMerkleBase is
 
     /// @dev Modifier to revert if `claimType_` value does not match the campaign's claim type.
     modifier revertIfNot(ClaimType claimType_) {
-        if (claimType != claimType_) {
+        if (CLAIM_TYPE != claimType_) {
             revert Errors.SablierMerkleBase_InvalidClaimType({
                 claimTypeCalled: claimType_,
-                claimTypeSupported: claimType
+                claimTypeSupported: CLAIM_TYPE
             });
         }
         _;
@@ -91,13 +91,13 @@ abstract contract SablierMerkleBase is
     /// @notice Constructs the contract by initializing the immutable state variables.
     constructor(MerkleBase.ConstructorParams memory baseParams) Adminable(baseParams.initialAdmin) {
         CAMPAIGN_START_TIME = baseParams.campaignStartTime;
+        CLAIM_TYPE = baseParams.claimType;
         COMPTROLLER = baseParams.comptroller;
         EXPIRATION = baseParams.expiration;
         MERKLE_ROOT = baseParams.merkleRoot;
         TOKEN = baseParams.token;
 
         campaignName = baseParams.campaignName;
-        claimType = baseParams.claimType;
         ipfsCID = baseParams.ipfsCID;
         minFeeUSD = ISablierComptroller(baseParams.comptroller)
             .getMinFeeUSDFor({ protocol: ISablierComptroller.Protocol.Airdrops, user: baseParams.campaignCreator });
--- a/airdrops/src/abstracts/SablierMerkleBase.sol
+++ b/airdrops/src/abstracts/SablierMerkleBase.sol
@@ -73,11 +73,11 @@ abstract contract SablierMerkleBase is
         _;
     }
 
-    /// @dev Modifier to revert if `claimType_` value does not match the campaign's claim type.
-    modifier revertIfNot(ClaimType claimType_) {
-        if (CLAIM_TYPE != claimType_) {
-            revert Errors.SablierMerkleBase_InvalidClaimType({
-                claimTypeCalled: claimType_,
+    /// @dev Modifier to revert if `claimType` value does not match the campaign's claim type.
+    modifier revertIfNot(ClaimType claimType) {
+        if (CLAIM_TYPE != claimType) {
+            revert Errors.SablierMerkleBase_UnsupportedClaimType({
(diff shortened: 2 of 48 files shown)


airdrops/src/types/DataTypes.sol
```diff
@@ -217,6 +232,7 @@ library MerkleVCA {
         uint128 aggregateAmount;
         string campaignName;
         uint40 campaignStartTime;
+        ClaimType claimType;
```
andreivladbrg, line 242: note: VCA inherits `SablierMerkleSignature` so it include the claim type logic

i think we need to remove it from `MerkleVCA` as there is no `claimViaAttestation` function - thus, not passing it in `factory.create` would have the `DEFAULT` value anyway

alternatively, we can keep it and check if it's `ATTEST` and revert if so

andreivladbrg: also, we can override it:

```solidity
function createMerkleVCA(MerkleVCA.ConstructorParams calldata campaignParams, uint256 recipientCount) external returns (ISablierMerkleVCA merkleVCA) {
    campaignParams.claimType = ClaimType.DEFAULT;
```

airdrops/src/types/DataTypes.sol
```diff
@@ -7,6 +7,14 @@ import { UD2x18 } from "@prb/math/src/UD2x18.sol";
 import { UD60x18 } from "@prb/math/src/UD60x18.sol";
 import { ISablierLockup } from "@sablier/lockup/src/interfaces/ISablierLockup.sol";
 
+/// @notice Enum representing the type of claim functions available for a Merkle campaign.
+/// @custom:value DEFAULT Activates `claim`, `claimTo`, and `claimViaSig` functions.
+/// @custom:value ATTEST Activates only the `claimViaAttestation` function.
+enum ClaimType {
+    DEFAULT,
+    ATTEST
```
smol-ninja: I understand that `MerkleExecute` does not implement any other claim functions, but we should still include `Execute` type to it. It would be good for API integrations.

A user will be able to query the type and understand what claim functions the contract supports.

1. If `EXECUTE`: supports only `claimAndExecute`
2. If `ATTEST`: supports only `claimViaAttestation`
3. If `DEFAULT`: supports `claim`, `claimTo` and `claimViaSig`

This would also allow us to move this to `SablierMekleBase` where it makes more sense than `MerkleSignature` contract.

airdrops/src/SablierMerkleInstant.sol
```diff
@@ -75,6 +76,7 @@ contract SablierMerkleInstant is
         external
         payable
         override
+        checkClaimType(ClaimType.DEFAULT)
```
smol-ninja: `restrictTo` or something like this? We are restricting function to be accessible when claim type matches the value.

airdrops/src/types/DataTypes.sol
smol-ninja: I think this may not be a problem after my latest commits to https://github.com/sablier-labs/lockup/pull/1403

smol-ninja (changes requested):
I am leaving a few comments below. I have resolved the conflicts already locally but need to have consensus on https://github.com/sablier-labs/lockup/pull/1403 first.

airdrops/src/SablierMerkleInstant.sol
andreivladbrg: i am open to renaming it, but to me `restrictTo` doesn't sound that good
wdyt about: `revertIfNot`

airdrops/src/types/DataTypes.sol
andreivladbrg: yes, its no longer a problem

airdrops/src/types/DataTypes.sol
andreivladbrg: idk - having it in the `SablierMerkleBase` would make me think that all campaigns must implement all claim types - which is not the case

i believe keeping it in the `MerkleSignature` contract is a simpler approach and doesn't lead to confusion

smol-ninja: > would make me think that all campaigns must implement all claim types

From where I am coming from is that there is no way to identify what claim functions does the merkle contract supports and having `EXECUTE` is a very good approach to offer that API.

The datatype already has all libraries so you can say that seeing all types in one file may also make you think that its a singleton contract. IDK. I am just trying to make analogies that may be the confusion that you are talking about may not be parmanent.

airdrops/src/SablierMerkleInstant.sol
smol-ninja: `revertIfNot` SGTM. I can make the change in my review commit.

airdrops/src/types/DataTypes.sol
andreivladbrg: > From where I am coming from is that there is no way to identify what claim functions does the merkle contract supports and having `EXECUTE` is a very good approach to offer that API.

since the campaign only implements `claimAndExecute` (and is the only campaign that implements this function) and no other function, i believe that's enough to distinguish between them

> The datatype already has all libraries so you can say that seeing all types in one file may also make you think that its a singleton contract. IDK. I am just trying to make analogies that may be the confusion that you are talking about may not be parmanent.

to me, it's confusing to include in the enum a value that is related to a unique campaign, which implements just that value and the others don't implement it

andreivladbrg: `EXECUTE ` is not included in any of the other campaigns:

```mermaid
graph TD
    classDef campaignStyle fill:#4a90d9,stroke:#2c5f8a,color:#fff
    classDef funcStyle fill:#50b86c,stroke:#357a48,color:#fff
    classDef claimTypeStyle fill:#e8a838,stroke:#b07c1e,color:#fff
    classDef executeStyle fill:#d94a6e,stroke:#8a2c45,color:#fff
    classDef groupStyle fill:#6c5ce7,stroke:#4834a8,color:#fff

    Campaigns(("All Campaigns"))

    Execute(("SablierMerkleExecute"))
    Instant(("SablierMerkleInstant"))
    LL(("SablierMerkleLL"))
    LT(("SablierMerkleLT"))
    VCA(("SablierMerkleVCA"))

    Campaigns --> Execute
    Campaigns --> Instant
    Campaigns --> LL
    Campaigns --> LT
    Campaigns --> VCA

    ClaimAndExecute(("claimAndExecute"))
    Execute --> ClaimAndExecute

    DEFAULT(("ClaimType: DEFAULT"))
    ATTEST(("ClaimType: ATTEST"))

    Instant --> DEFAULT
    LL --> DEFAULT
    LT --> DEFAULT
    VCA --> DEFAULT

    Instant --> ATTEST
    LL --> ATTEST
    LT --> ATTEST
    VCA --> ATTEST

    Claim(("claim"))
    ClaimTo(("claimTo"))
    ClaimViaSig(("claimViaSig"))

    DEFAULT --> Claim
    DEFAULT --> ClaimTo
    DEFAULT --> ClaimViaSig

    ClaimViaAttestation(("claimViaAttestation"))
    ATTEST --> ClaimViaAttestation

    class Campaigns groupStyle
    class Execute,Instant,LL,LT,VCA campaignStyle
    class DEFAULT,ATTEST claimTypeStyle
    class Claim,ClaimTo,ClaimViaSig,ClaimViaAttestation funcStyle
    class ClaimAndExecute executeStyle
```

smol-ninja: > since the campaign only implements claimAndExecute (and is the only campaign that implements this function) and no other function, i believe that's enough to distinguish between them

Yeah thats write but my question was from the perspective of the public API. For ex, I can query claim type function to a merkle campaign and based on what it returns I can know which function to call on it. The differences at the code level cannot be seen through an API. I think thats a better UX for integrators and even for our own UI.

> to me, it's confusing to include in the enum a value that is related to a unique campaign, which implements just that value and the others don't implement it

I don't see that as a problem. I am purely coming from the DX pov (my above point) even if `EXECUTE` is only used in one campaign.

To resolve this discussion, I would like to know what you think of my API argument

Also, if we allow `MerkleExecute` to have claim type, we can move `claimType` to Merkle Base which will also be a good design practice.

airdrops/src/SablierMerkleInstant.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1405/changes/760fa7898c6366bd800419bb6d4ae664ac98705d

I also inlined the code into modifier because we dont have size issue here plus it offers better readability.

airdrops/src/abstracts/SablierMerkleSignature.sol
```diff
@@ -32,19 +33,40 @@ abstract contract SablierMerkleSignature is
...
+    /// @inheritdoc ISablierMerkleSignature
+    ClaimType public override claimType;
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                     MODIFIERS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Modifier to revert if `claimType_` value does not match the campaign's claim type.
+    modifier revertIfNot(ClaimType claimType_) {
+        if (claimType != claimType_) {
```
andreivladbrg: why did you remove the function and moved the logic into the modifier?

andreivladbrg: @smol-ninja

airdrops/src/types/DataTypes.sol
andreivladbrg: alright, feel free to move it to base contract and include the `EXECUTE` value in the claim type, but order them like this:

```solidity
enum ClaimType {
    DEFAULT,
    ATTEST,
    EXECUTE
}
```

airdrops/src/abstracts/SablierMerkleSignature.sol
smol-ninja: I can bring it back. I dropped a https://github.com/sablier-labs/lockup/pull/1405#discussion_r2788084904. Let me know what you think.

airdrops/src/types/DataTypes.sol
smol-ninja: Alright thank you. I will order them like this.

smol-ninja: https://github.com/sablier-labs/lockup/pull/1405/changes/b928ca0a79d6a29ec7128c52dad662d5cb8a2b84

airdrops/src/abstracts/SablierMerkleSignature.sol
andreivladbrg: oh missed it

airdrops/src/SablierMerkleInstant.sol
andreivladbrg: alright, sounds good then, we can keep the logic in the modifier

andreivladbrg:
@smol-ninja do you think it would be useful to deploy campaigns with `ClaimType.ATTEST` only if the attestor is set in the comptroller? (not `address(0)`)

smol-ninja:
Attestor can also be set afterwards, imo, we should deploy it regardless of its value in comptroller.

airdrops/src/abstracts/SablierMerkleBase.sol
```diff
@@ -46,6 +46,9 @@ abstract contract SablierMerkleBase is
     /// @inheritdoc ISablierMerkleBase
     string public override campaignName;
 
+    /// @inheritdoc ISablierMerkleBase
+    ClaimType public override claimType;
+
```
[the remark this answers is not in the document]
andreivladbrg: @smol-ninja its a good point, should we mark it as `immutable` ?

i believe yes, since we don't have a setter to switch between them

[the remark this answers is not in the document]
smol-ninja: Yeah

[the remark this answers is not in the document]
andreivladbrg: 93e433c37bfd4e1065f9aa122029916b968921e3

andreivladbrg:
@smol-ninja pushed a two commits 93e433c37bfd4e1065f9aa122029916b968921e3 and 5e8c0712ed47c16de0b10312484fdaf826f166b3, if they look good, feel free to merge it

airdrops/src/abstracts/SablierMerkleBase.sol
```diff
@@ -64,8 +67,20 @@ abstract contract SablierMerkleBase is
...
     modifier notZeroAddress(address to) {
-        _revertIfToZeroAddress(to);
+        if (to == address(0)) {
+            revert Errors.SablierMerkleBase_ToZeroAddress();
+        }
+        _;
+    }
 
+    /// @dev Modifier to revert if `claimType_` value does not match the campaign's claim type.
+    modifier revertIfNot(ClaimType claimType_) {
```
smol-ninja: We can use `claimType` type now that the storage variable has been renamed to `CLAIM_TYPE`

airdrops/src/libraries/Errors.sol
```diff
@@ -81,6 +83,9 @@ library Errors {
     /// @notice Thrown when trying to claim the same index more than once.
     error SablierMerkleBase_IndexClaimed(uint256 index);
 
+    /// @notice Thrown when trying to call a claim function not supported in the campaign.
+    error SablierMerkleBase_InvalidClaimType(ClaimType claimTypeCalled, ClaimType claimTypeSupported);
```
smol-ninja: Its a bit confusing. So, we can rename it to:

```solidity
error SablierMerkleBase_UnsupportedClaimType(ClaimType claimTypeRequired, ClaimType claimTypeSupported);
```

airdrops/tests/integration/concrete/campaign/instant/claim/claim.t.sol
```diff
@@ -11,9 +13,20 @@ contract Claim_MerkleInstant_Integration_Test is Claim_Integration_Test, MerkleI
         MerkleInstant_Integration_Shared_Test.setUp();
     }
 
+    function test_RevertGiven_ClaimTypeATTEST() external {
```
smol-ninja, line 16: I think these should be moved to `shared/claim` tests.

smol-ninja (approved):
Pushing a commit to address some of the comments below.

airdrops/src/abstracts/SablierMerkleBase.sol
andreivladbrg: good catch

airdrops/src/libraries/Errors.sol
andreivladbrg: both work to me - so if you want, make the change

airdrops/src/abstracts/SablierMerkleBase.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1405/changes/8ab44a56d2879873f72f99055964cb181d0b8e57

airdrops/src/libraries/Errors.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1405/changes/8ab44a56d2879873f72f99055964cb181d0b8e57

lmk if you disagree with the change

airdrops/tests/integration/concrete/campaign/instant/claim/claim.t.sol
smol-ninja: have done something locally, will push afterwards. We can merge the PR without that.

andreivladbrg: if we dry fy, sure

airdrops/src/libraries/Errors.sol
andreivladbrg: looks good

PR #1406, opened 2026-02-01
feat: add price gated model

### LINKED ISSUE #1280: Implement price gated timelock stream
As discussed in https://github.com/sablier-labs/command-center/discussions/175, implement price-gated timelock stream. 

1. For PoC design refer to [this thread](https://github.com/sablier-labs/command-center/discussions/175#discussioncomment-13961923).
2. Add a [new invariant](https://github.com/sablier-labs/command-center/discussions/175#discussioncomment-13986809).
3. Some good [Qs](https://github.com/sablier-labs/command-center/discussions/175#discussioncomment-14019783) and [As](https://github.com/sablier-labs/command-center/discussions/175#discussioncomment-14021342).

The proof of concept in the linked discussion is purely for demonstration purpose. Therefore, you are at the liberty of improving the technical implementation, variable and function names, data types as well as the storage layouts.

Do check the contract size after the implementation and ask questions in case you hit a wall.

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1280 with 200 optimizer runs.

I ended up building a simple version in which a sender can create a stream for a recipient. The tokens are deposited only at the time of creation.

### Access Control

 | Condition                       | Cancel          | Withdraw       |                               
  |---------------------------------|-----------------|----------------|                               
  | Price >= target (anytime)       | ❌ Reverts      | ✅ Can withdraw upto 100% |                               
  | Price < target, not expired     | ✅ Returns 100% | ❌ Reverts     |                               
  | Expired (regardless of price)   | ❌ Reverts      | ✅ Can withdraw upto 100% |

If the price goes above the target price and  then goes down later, sender will now be able to cancel the stream. However, if end time is reached, recipient will be able to withdraw regardless of the price.

### Changelog

- [x] Fork tests with mainnet chainlink oracle for ETH/USD
- [x] concrete test for streamed amount is enough to validate withdraw works correctly
- [x] Statuses may be fucked up. for ex. it would show streaming if `target price < oracle price` and `blockTime < endTime`


lockup/src/interfaces/ISablierLockupPriceGated.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// the sum of `block.timestamp` and `duration`. The stream is funded by `msg.sender` and is wrapped in an
+    /// ERC-721 NFT.
+    ///
+    /// @dev Emits a {Transfer}, {CreateLockupPriceGatedStream} and {MetadataUpdate} event.
+    ///
+    /// Notes:
+    /// - The recipient can withdraw the full deposited amount when either:
+    ///   1. The oracle price reaches or exceeds the target price, OR
+    ///   2. Current time is greater than the stream's end time.
+    /// - The sender can cancel the stream when price < target AND end time is in the future.
```
andreivladbrg:
```suggestion
    /// - The sender can cancel the stream when price is less than target price AND end time is in the future.
```

lockup/src/abstracts/SablierLockupPriceGated.sol
```diff
@@ -0,0 +1,93 @@
...
+        payable
+        override
+        noDelegateCall
+        returns (uint256 streamId)
+    {
+        // Check: validate that the oracle implements the {AggregatorV3Interface} interface.
+        uint128 latestPrice = LockupHelpers.validateOracle(oracle);
+
+        // Check: the target price is greater than the latest price.
+        if (targetPrice <= latestPrice) revert Errors.SablierLockup_TargetPriceTooLow(targetPrice, latestPrice);
```
andreivladbrg: why not move this in the `LockupHelpers` ?

lockup/src/types/Lockup.sol
```diff
@@ -89,7 +89,8 @@ library Lockup {
     enum Model {
         LOCKUP_LINEAR,
         LOCKUP_DYNAMIC,
-        LOCKUP_TRANCHED
+        LOCKUP_TRANCHED,
+        LOCKUP_PRICE_GATED
```
andreivladbrg: not alphabetically ordered

lockup/src/libraries/LockupHelpers.sol
```diff
@@ -140,6 +166,34 @@ library LockupHelpers {
         _checkTranches(tranches, depositAmount, timestamps);
     }
 
+    /// @dev Validates the oracle address used in the LPG stream.
+    function validateOracle(AggregatorV3Interface oracle) public view returns (uint128 latestPrice) {
```
andreivladbrg: what if we declare this function in the `Comptroller`? we already need it there, and we need it for the Bob protocol as well.

alternatively, we could add a library in `utils` for this

andreivladbrg: also, since the function also returns the price, what if rename it to `safeOraclePrice`?

lockup/src/interfaces/ISablierLockupPriceGated.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// - The function does not check if the provided oracle reports the price for the deposited token.
```
andreivladbrg, line 43: wdym by "if the provided oracle reports the price for the deposited token"?

`validateOracle` its called in the create function, thus, the oracles is checked if it returns a price

lockup/src/interfaces/ISablierLockupPriceGated.sol
```diff
@@ -0,0 +1,71 @@
...
+    ///
+    /// Requirements:
+    /// - Must not be delegate called.
+    /// - `params.depositAmount` must be greater than zero.
+    /// - `params.sender` must not be the zero address.
+    /// - `params.recipient` must not be the zero address.
+    /// - `duration` must be greater than zero.
+    /// - `oracle` must implement the {AggregatorV3Interface} interface.
```
andreivladbrg:
```suggestion
    /// - `oracle` must implement Chainlink's {AggregatorV3Interface} interface.
```

lockup/src/interfaces/ISablierLockupPriceGated.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// - `oracle` must return 8 decimals when the `decimals()` function is called.
+    /// - `oracle` must return a positive price when the `latestRoundData()` function is called.
+    /// - `targetPrice` must be greater than the current oracle price.
+    /// - `msg.sender` must have allowed this contract to spend at least `params.depositAmount` tokens.
+    /// - `params.token` must not be the native token.
+    /// - `params.shape.length` must not be greater than 32 characters.
+    ///
+    /// @param params Struct encapsulating the function parameters, which are documented in {Lockup} type.
+    /// @param oracle The address of a price feed oracle used for retrieving the latest price.
+    /// @param targetPrice The price that must be reached to unlock the tokens, denoted in 8 decimals where 1e8 = $1.
```
andreivladbrg:
```suggestion
    /// @param targetPrice The price that must be reached to unlock the tokens, denominated in Chainlink's 8-decimal, where 1e8 = $1.
```

.github/workflows/ci-module.yml
```diff
@@ -133,5 +133,5 @@ jobs:
       foundry-version: "v1.5.0"
       match-path: "tests/{integration,unit}/**/*.sol"
       name: "Coverage"
-      via-ir-minimum: true
+      via-ir-minimum: false
```
andreivladbrg, line 136: why this?

lockup/src/types/LockupPriceGated.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity >=0.8.22;
+
+import { AggregatorV3Interface } from "@chainlink/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol";
+
+/// @notice Namespace for the structs used only in LPG streams.
+library LockupPriceGated {
+    /// @notice Struct encapsulating the unlock parameters for price-gated streams.
+    /// @param oracle The price feed oracle address used to retrieve the latest price.
+    /// @param targetPrice The price that must be reached to unlock the tokens, denoted in 8 decimals where 1e8 = $1.
```
andreivladbrg:
```suggestion
    /// @param targetPrice The price that must be reached to unlock the tokens, denominated in Chainlink's 8-decimal, where 1e8 = $1.
```

lockup/src/types/LockupPriceGated.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity >=0.8.22;
+
+import { AggregatorV3Interface } from "@chainlink/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol";
+
+/// @notice Namespace for the structs used only in LPG streams.
+library LockupPriceGated {
+    /// @notice Struct encapsulating the unlock parameters for price-gated streams.
```
andreivladbrg:
```suggestion
    /// @notice Struct encapsulating the unlock parameters for LPG streams.
```

lockup/src/libraries/LockupMath.sol
```diff
@@ -231,6 +232,46 @@ library LockupMath {
         }
     }
 
+    /// @notice Calculates the streamed amount of LPG streams.
+    /// @dev The LPG streaming model uses all-or-nothing unlock based on price threshold:
+    ///
+    /// $$
+    ///        ⎧ deposited - refunded, block timestamp >= end time OR latest price >= target price
```
andreivladbrg: > deposited - refunded

if the stream is canceled (i.e. refunded > 0), it means that the refunded = deposited, right?

if yes, why not simply check `wasCanceled == true` and return 0 instead of having refunded param here

andreivladbrg: and in the `_cancel` function change it to:

```solidity
uint128 streamedAmount;
if (_streams[streamId].lockupModel == Lockup.Model.LOCKUP_PRICE_GATED) {
    streamedAmount = 0;
} else {
    streamedAmount = _streamedAmountOf(streamId);
}
```

lockup/src/libraries/LockupMath.sol
```diff
@@ -231,6 +232,46 @@ library LockupMath {
...
+        AggregatorV3Interface oracle,
+        uint128 targetPrice
+    )
+        external
+        view
+        returns (uint128)
+    {
+        // If the stream has expired, return the full amount.
+        if (block.timestamp >= endTime) {
+            return deposited - refunded;
```
andreivladbrg: if the target price's end time has passed, shouldn't the streamed amount be 0?

lockup/src/libraries/LockupMath.sol
```diff
@@ -231,6 +232,46 @@ library LockupMath {
...
+    /// $$
+    ///
+    /// Assumptions:
+    /// 1. The oracle is assumed to be returning the correct price.
+    function calculateStreamedAmountLPG(
+        uint128 deposited,
+        uint128 refunded,
+        uint40 endTime,
+        AggregatorV3Interface oracle,
+        uint128 targetPrice
```
andreivladbrg: instead of these 2 params, why not change it to one struct param? `LockupPriceGated.UnlockParams`

lockup/src/SablierLockup.sol
```diff
@@ -480,6 +482,16 @@ contract SablierLockup is
                 tranches: _tranches[streamId]
             });
         }
+        // Calculate the streamed amount for the LPG model.
+        else if (stream.lockupModel == Lockup.Model.LOCKUP_PRICE_GATED) {
```
andreivladbrg: the `else if` are not alphabetically ordered

lockup/src/libraries/LockupMath.sol
```diff
@@ -231,6 +232,46 @@ library LockupMath {
...
+        view
+        returns (uint128)
+    {
+        // If the stream has expired, return the full amount.
+        if (block.timestamp >= endTime) {
+            return deposited - refunded;
+        }
+
+        // Get the current oracle price.
+        (, int256 price,,,) = oracle.latestRoundData();
```
andreivladbrg: lets use the try catch here like in the comptroller's convert function

lockup/src/interfaces/ISablierLockupPriceGated.sol
```diff
@@ -0,0 +1,71 @@
...
+    /// @param duration The total duration of the stream in seconds.
+    /// @return streamId The ID of the newly created stream.
+    function createWithDurationsLPG(
+        Lockup.CreateWithDurations calldata params,
+        AggregatorV3Interface oracle,
+        uint128 targetPrice,
```
andreivladbrg: why not `LockupPriceGated.UnlockParams` as param instead?

lockup/src/libraries/LockupMath.sol
```diff
@@ -231,6 +232,46 @@ library LockupMath {
...
+
+        // If the oracle price is at or above the target price, return the full amount.
+        if (price > 0 && uint128(uint256(price)) >= targetPrice) {
```
andreivladbrg: why no safe cast here?

andreivladbrg (commented):
great work Shub

leaving these comments just for the `src` code in order to be quicker with the review (have not looked at the tests)

also, i want to add a note that the status transitions invariant should mention that it does not apply in the case of the LPG model

lockup/src/libraries/LockupMath.sol
andreivladbrg: we should add an invariant for: if `refunded > 0` then `refunded = deposited`

for this model

lockup/src/abstracts/SablierLockupPriceGated.sol
smol-ninja: as far as i remember, I moved it initially but that increased the contract size due to how it uses evm stack when delegating call to a library. Inlining is less contract heavy.

lockup/src/libraries/LockupHelpers.sol
smol-ninja: > what if we declare this function in the Comptroller

I would say instead of comptroller it should be part of `utils`. Comptroller should be used to set admin related values only. This is more like a utility function.

> since the function also returns the price, what if rename it to safeOraclePrice

Since its only used during create and not every time, `safeOraclePrice` isn't being honest. Any other suggestion or you still think `safeOraclePrice` is a better name over `validateOracle`?

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: > if the provided oracle reports the price for the deposited token

That is it returns the price for the deposited token. For ex, what if a user creates stream with USDC but uses oracle for UNI token. Returning the price is not sufficient and hence the note.

.github/workflows/ci-module.yml
smol-ninja: Coverage tests return `stack too deep` error otherwise. Because `via-ir-minimum: true` also build tests with optimized profile which returns the error. Why did we need it to be `true` in the first place? Its just code coverage right?

lockup/src/libraries/LockupMath.sol
smol-ninja: > if yes, why not simply check wasCanceled == true and return 0 instead of having refunded param here

Because its a math library and I thought writing the function using math values only would be nicer way to write the function.

> we should add an invariant for: if refunded > 0 then refunded = deposited

Agree

lockup/src/libraries/LockupMath.sol
smol-ninja: No, it should be either `deposited` if not canceled or 0 if canceled and hence `deposited - refunded`.

lockup/src/libraries/LockupMath.sol
smol-ninja: Can try and see if the contract size remains same.

lockup/src/SablierLockup.sol
smol-ninja: will do

lockup/src/libraries/LockupMath.sol
smol-ninja: OK

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: Will do

lockup/src/libraries/LockupMath.sol
smol-ninja: Good point.

lockup/src/libraries/LockupHelpers.sol
andreivladbrg: > Comptroller should be used to set admin related values only. This is more like a utility function.

the comptroller is used in the `withdraw` function to calculate the minimum fee, tho. either way, the fact that it is moved to the `utils` folder is the important thing

> Since its only used during create and not every time, `safeOraclePrice`

as said above, this would be used in `withdraw` as well - also bob
i personally prefer `safeOraclePrice`, but i don't have a strong opinion

lockup/src/interfaces/ISablierLockupPriceGated.sol
andreivladbrg: oh i see, right

let's add an assumption in the `Security.md` file that there is trust between the sender and the recipient that the oracle is set for the deposited token

.github/workflows/ci-module.yml
andreivladbrg: yeah, we don't need it

lockup/src/libraries/LockupMath.sol
andreivladbrg: > Because its a math library and I thought writing the function using math values only would be nicer way to write the function.
> 
> >

i see, tho, i think its more efficient to just check a boolean, right? (instead of having a subtraction)

lockup/src/libraries/LockupMath.sol
andreivladbrg: right, i left this comment before our slack conversation

lockup/src/libraries/LockupHelpers.sol
smol-ninja: I will change the name to `safeOraclePrice`

lockup/src/libraries/LockupMath.sol
smol-ninja: Yeah. Let me see what I can do here.

lockup/tests/integration/concrete/lockup-price-gated/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -0,0 +1,11 @@
+StreamedAmountOf_Lockup_PriceGated_Integration_Concrete_Test
+├── given depleted stream
+│  └── it should return the deposited
+└── given not depleted stream
+   ├── when latest price below target
+   │  ├── when expired stream
+   │  │  └── it should return zero
```
andreivladbrg: the branch says return zero, but the test asserts deposited

Image

lockup/tests/integration/concrete/lockup-price-gated/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -0,0 +1,11 @@
...
+   │  └── when not expired stream
+   │     └── it should return the deposited amount
```
andreivladbrg: same here, it says "return the deposited amount" but asserts zero

Image

lockup/src/abstracts/SablierLockupPriceGated.sol
andreivladbrg: interesting, i will test this

lockup/tests/fork/tokens/WETH.t.sol
```diff
@@ -0,0 +1,15 @@
...
+import { AggregatorV3Interface } from "@chainlink/contracts/src/v0.8/shared/interfaces/AggregatorV3Interface.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { Lockup_PriceGated_Fork_Test } from "../LockupPriceGated.t.sol";
+
+/// @dev For testing price-gated streams.
+IERC20 constant WETH = IERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+
+/// @dev The Chainlink ETH/USD price feed on Ethereum mainnet.
+AggregatorV3Interface constant CHAINLINK_ETH_USD = AggregatorV3Interface(0x5f4eC3Df9cbd43714FE2740f5E3616155c5b8419);
```
andreivladbrg, line 13: should we use utils' `BaseScript` for the chainlink oracle?

lockup/tests/mocks/OracleMock.sol
andreivladbrg, line 1: some mocks are duplicated from `utils/src/mocks/ChainlinkMocks.sol` - why not use them & edit them if needed?

lockup/tests/integration/Integration.t.sol
```diff
@@ -97,15 +98,44 @@ abstract contract Integration_Test is Base_Test {
     //////////////////////////////////////////////////////////////////////////*/
 
     function initializeDefaultStreams() internal {
-        ids.defaultStream = createDefaultStream();
-        ids.notAllowedToHookStream = createDefaultStreamWithRecipient(address(recipientInterfaceIDIncorrect));
-        ids.notCancelableStream = createDefaultStreamNonCancelable();
-        ids.notTransferableStream = createDefaultStreamNonTransferable();
+        if (lockupModel == Lockup.Model.LOCKUP_PRICE_GATED) {
```
andreivladbrg: instead of duplicating the creation of the default streams, why not use an else if branch in this function:

```solidity
function createDefaultStream(Lockup.CreateWithTimestamps memory params) internal returns (uint256 streamId) {
    if (lockupModel == Lockup.Model.LOCKUP_DYNAMIC) {
        streamId = lockup.createWithTimestampsLD(params, _defaultParams.segments);
    } else if (lockupModel == Lockup.Model.LOCKUP_LINEAR) {
        streamId = lockup.createWithTimestampsLL(
            params, _defaultParams.unlockAmounts, _defaultParams.granularity, _defaultParams.cliffTime
        );
    } else if (lockupModel == Lockup.Model.LOCKUP_TRANCHED) {
        streamId = lockup.createWithTimestampsLT(params, _defaultParams.tranches);
    }
}
```

and pass the timestamp params to the duration struct

lockup/tests/integration/fuzz/lockup/cancel.t.sol
```diff
@@ -116,7 +116,7 @@ abstract contract Cancel_Integration_Fuzz_Test is Integration_Test {
 
         // Assert that the stream's status is "CANCELED".
```
andreivladbrg: the comment was not updated accordingly to the change:

```solidity
Lockup.Status expectedStatus = recipientAmount == 0 ? Lockup.Status.DEPLETED : Lockup.Status.CANCELED;
```

lockup/tests/integration/fuzz/lockup-price-gated/streamedAmountOf.t.sol
```diff
@@ -0,0 +1,76 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22 <0.9.0;
+
+import { Lockup_PriceGated_Integration_Fuzz_Test } from "./LockupPriceGated.t.sol";
+
+contract StreamedAmountOf_Lockup_PriceGated_Integration_Fuzz_Test is Lockup_PriceGated_Integration_Fuzz_Test {
+    /// @dev Given enough fuzz runs, all of the following scenarios will be fuzzed:
+    ///
+    /// - Multiple time jumps before end time
+    /// - Oracle price below target
```
andreivladbrg, line 10: why not include the case - canceled or not?

lockup/tests/integration/concrete/lockup/cancel/cancel.t.sol
```diff
@@ -180,6 +181,10 @@ abstract contract Cancel_Integration_Concrete_Test is Integration_Test {
         // It should make Sablier run the recipient hook.
         uint128 senderAmount = lockup.refundableAmountOf(ids.recipientReentrantStream);
         uint128 recipientAmount = lockup.withdrawableAmountOf(ids.recipientReentrantStream);
+
+        // Return if the recipient amount is 0, because otherwise withdraw would revert.
+        if (recipientAmount == 0) return;
```
andreivladbrg: woulnd't it be clearer to return in case of the LPG model?

lockup/tests/integration/concrete/lockup-price-gated/streamed-amount-of/streamedAmountOf.tree
```diff
@@ -0,0 +1,11 @@
+StreamedAmountOf_Lockup_PriceGated_Integration_Concrete_Test
+├── given depleted stream
+│  └── it should return the deposited
+└── given not depleted stream
+   ├── when latest price below target
+   │  ├── when expired stream
```
andreivladbrg: wouldn't it be better to say, end time in future or not?

"expired" is an unused term across the codebase

lockup/tests/integration/concrete/lockup-price-gated/create-with-durations-lpg/createWithDurationsLPG.tree
```diff
@@ -0,0 +1,31 @@
...
+                  └── when target price exceeds current price
+                     ├── when duration zero
+                     │  └── it should revert
+                     └── when duration not zero
+                        ├── it should create the stream
+                        ├── it should bump the next stream ID
+                        ├── it should store the unlock params
+                        ├── it should mint the NFT
+                        ├── it should perform the ERC-20 transfer
+                        └── it should emit {CreateLockupPriceGatedStream} and {MetadataUpdate} events
```
andreivladbrg, line 31: missing the transfer event

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/types/Lockup.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/libraries/LockupHelpers.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/types/LockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/types/LockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/libraries/LockupMath.sol
smol-ninja: Implemented `wasCanceled`

https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/tests/fork/LockupPriceGated.t.sol
```diff
@@ -0,0 +1,442 @@
...
+        uint256 streamId;
+        Lockup.Timestamps timestamps;
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                  STATE VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev The Chainlink price feed oracle used for this test.
+    AggregatorV3Interface internal immutable ORACLE;
```
andreivladbrg, line 60: here we say `ORACLE` and in the `tokens/WETH.t.sol` we say `CHAINLINK_ETH_USD` - small inconsistency

lockup/src/libraries/LockupMath.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/tests/integration/fuzz/lockup-price-gated/LockupPriceGated.t.sol
```diff
@@ -0,0 +1,43 @@
...
+        vm.warp({ newTimestamp: defaults.START_TIME() });
+        initializeDefaultStreams();
+    }
+}
+
+/*//////////////////////////////////////////////////////////////////////////
+                                SHARED TESTS
+//////////////////////////////////////////////////////////////////////////*/
+
+contract Cancel_Lockup_PriceGated_Integration_Fuzz_Test is
```
andreivladbrg, line 27: why only cancel and not withdraw in case of LPG?

lockup/src/SablierLockup.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/libraries/LockupMath.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/interfaces/ISablierLockupPriceGated.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

lockup/src/libraries/LockupMath.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/ce9b6553cd44db41d8dce962d582c952160b85c5

smol-ninja:
@andreivladbrg I have implemented all your suggestion and contract size is also within the limit.

lockup/tests/fork/LockupPriceGated.t.sol
```diff
@@ -0,0 +1,442 @@
...
+            oracle: ORACLE,
+            targetPrice: params.targetPrice,
+            duration: params.duration
+        });
+
+        // Run post-create assertions.
+        _assertPostCreate(params);
+
+        /*//////////////////////////////////////////////////////////////////////////
+                                          CANCEL
```
andreivladbrg, line 281: we should create a new stream to also test `withdraw`

only `cancel` is tested here

andreivladbrg (commented):
reviewed the tests as well

we should add some tests for the status - it can go from `Settled` to `Streaming`
which was not the case until now

utils/src/libraries/SafeOracle.sol
```diff
@@ -0,0 +1,41 @@
...
+
+        // Check: oracle returns a positive price when `latestRoundData()` is called.
+        try oracle.latestRoundData() returns (uint80, int256 price, uint256, uint256, uint80) {
+            if (price <= 0) {
+                revert Errors.SafeOracle_NegativePrice(address(oracle));
+            }
+            latestPrice = uint256(price).toUint128();
+        } catch {
+            revert Errors.SafeOracle_MissesInterface(address(oracle));
+        }
```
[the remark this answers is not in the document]
smol-ninja: We do not want to check for the oracle staleness

lockup/tests/fork/tokens/WETH.t.sol
smol-ninja: In fork tests, we do hardcode addresses so I guess its fine here.

lockup/tests/mocks/OracleMock.sol
smol-ninja: I have already deleted them.

lockup/tests/integration/Integration.t.sol
smol-ninja: Not sure what you meant. Feel free to push a commit.

lockup/tests/integration/fuzz/lockup-price-gated/streamedAmountOf.t.sol
smol-ninja: Actually I am not sure if it would add any value. All these tests are for multiple values of time and price. `wasCanceled` is just a single state so imo integration test should be enough

lockup/tests/integration/concrete/lockup-price-gated/streamed-amount-of/streamedAmountOf.tree
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/3b68aaf6e819f297de1383f98a4bbf2fde5e7b4f

lockup/tests/integration/concrete/lockup-price-gated/streamed-amount-of/streamedAmountOf.tree
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/3b68aaf6e819f297de1383f98a4bbf2fde5e7b4f

lockup/tests/integration/fuzz/lockup/cancel.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/3b68aaf6e819f297de1383f98a4bbf2fde5e7b4f

lockup/tests/integration/concrete/lockup/cancel/cancel.t.sol
smol-ninja: I don't think it would be any clearer. Its rather clear this way because you cannot withdraw if `recipientAmount` is zero.

lockup/tests/integration/concrete/lockup-price-gated/create-with-durations-lpg/createWithDurationsLPG.tree
smol-ninja: Its missing in all create events hence the consistency. There is a `expectCallToTransferFrom` to test the token transfer.

lockup/tests/integration/concrete/lockup-price-gated/streamed-amount-of/streamedAmountOf.tree
smol-ninja: Agree and changed in https://github.com/sablier-labs/lockup/pull/1406/changes/e7ea86c89089b75fff5b59781c312a682a888d83

lockup/tests/fork/LockupPriceGated.t.sol
smol-ninja: Thats because from this contract point of view, there is a `forkToken` and `oracle`. And from `WETH_Lockup_PriceGated_Fork_Test` point of view, there is `WETH` and `CHAINLINK_ETH_USD` hence the difference.

lockup/tests/integration/fuzz/lockup-price-gated/LockupPriceGated.t.sol
smol-ninja: Its because imo `contract StreamedAmountOf_Lockup_PriceGated_Integration_Fuzz_Test` is covers withdraw too. As long as this function is well tested, its enough. Rest of the logic is common across all models.

lockup/tests/fork/LockupPriceGated.t.sol
smol-ninja: What do you want to test here? There is test for withdraw in fork

```solidity
                                          WITHDRAW
```

lockup/tests/integration/concrete/lockup/cancel/cancel.t.sol
andreivladbrg: > because you cannot withdraw if recipientAmount is zero.

thats correct, but why is it zero? it is because the LPG model is used

lockup/tests/integration/concrete/lockup-price-gated/create-with-durations-lpg/createWithDurationsLPG.tree
andreivladbrg: right

lockup/tests/fork/LockupPriceGated.t.sol
andreivladbrg: oh, mb there are 2 functions, i missed it

andreivladbrg: wdys about removing the create assertions from one of them? make the file shorter?

lockup/tests/integration/fuzz/lockup-price-gated/LockupPriceGated.t.sol
andreivladbrg: > StreamedAmountOf_Lockup_PriceGated_Integration_Fuzz_Test is covers

partially true - the difference is that withdraw is a state changing function, not only view

smol-ninja:
The CI was failing so I pushed a [new commit](https://github.com/sablier-labs/lockup/pull/1406/changes/c64263055a9ae751681e1926971bb903e85c6c37) which should fix it. I have also moved `SafeOracle` to `libraries` directory because its a library.

lockup/tests/integration/concrete/lockup/cancel/cancel.t.sol
smol-ninja: Thats true. OK I will change it.

lockup/tests/fork/LockupPriceGated.t.sol
smol-ninja: sgtm

lockup/tests/fork/tokens/WETH.t.sol
andreivladbrg: fair

lockup/tests/integration/fuzz/lockup-price-gated/LockupPriceGated.t.sol
smol-ninja: yes but between all models, the only logic that changes is streamed amount. So, whatever value this function returns withdraw uses that. 
If we add withdraw to it, it will just be a duplicate version of streamed amount of + shared/withdraw test. Unfortunately, we cannot re-use shared withdraw test here.

lockup/tests/integration/concrete/lockup/cancel/cancel.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/6ac1d87584f0dec22e358a93f2bd2f406e6d78db

lockup/tests/fork/LockupPriceGated.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1406/changes/6ac1d87584f0dec22e358a93f2bd2f406e6d78db

smol-ninja:
Added LPG create handler in https://github.com/sablier-labs/lockup/pull/1406/changes/966966456dd13a6bb4333573ddb98eb496c15fd4

@andreivladbrg if everything looks good now, can you approve?

lockup/tests/integration/Integration.t.sol
andreivladbrg: 566acf0f892a5f1be0cf8d3f55cb207d26ec0da2 lmk if it looks good

utils/src/libraries/SafeOracle.sol
```diff
@@ -0,0 +1,43 @@
...
+
+import { Errors } from "./Errors.sol";
+
+/// @title SafeOracle
+/// @notice Library with helper functions for validating oracle addresses.
+library SafeOracle {
+    using SafeCast for uint256;
+
+    /// @dev Validates the oracle address and returns the latest price.
+    function safeOraclePrice(AggregatorV3Interface oracle) public view returns (uint128 latestPrice) {
```
andreivladbrg: why is it not `internal`? we'd need to deploy it as well otherwise

lockup/tests/integration/Integration.t.sol
smol-ninja: The function takes `Lockup.CreateWithTimestamps` as the input argument:

```solidity
function createDefaultStream(Lockup.CreateWithTimestamps memory params) internal returns (uint256 streamId) {
```

but do this 

```solidity
Lockup.CreateWithDurations memory params_ = Lockup.CreateWithDurations
```

So basically if I am calling a custom `createDefaultStream(params)` the params get overridden. I don't find this as a good practice. Instead of this, we should have a bespoke create function for LPG.

utils/src/libraries/SafeOracle.sol
smol-ninja: Marking it internal would increase the contract size of `Lockup` because of inlining.

andreivladbrg: yes, but does it exceeds the size?

then, why not simply implement it in the comptroller instead of having one more deployable library

you know its messy

andreivladbrg: working on a refactor, will push if it works

lockup/tests/integration/Integration.t.sol
andreivladbrg: why does it matter what params it takes ?

the purpose of it is to create a default stream (what the name says)

andreivladbrg: these are tests - KISS - what is less go for it

utils/src/libraries/SafeOracle.sol
smol-ninja: > why not simply implement it in the comptroller

Sounds good.

lockup/tests/integration/Integration.t.sol
smol-ninja: > the purpose of it is to create a default stream (what the name says)

Why not add it to `function createDefaultStream()` instead of `createDefaultStream(Lockup.CreateWithTimestamps memory params)`? IMO it makes it complex because you would assume a helper function takes argument but gets overridden by internal logic.

> why does it matter what params it takes ?

It matters because the point of having a param in a function is be able to use that param in the function which is not the case for LPG logic.

utils/src/libraries/SafeOracle.sol
andreivladbrg: 7ed5cb0ac96c9bb3de37ca37e794ddc648f14dfb lmk what you think

smol-ninja: 1. How does the contract size look like?
2. Also the `optimizer_runs` is 200 right now. If you have to lower it further to incorporate the change, I wouldn't go for the change as I think we should keep 200 as a buffer.

andreivladbrg: It was below the limit

smol-ninja: Actually even better than what it was before. So nice work. The utils test is failing, make sure to fix that before merging the PR.

lockup/tests/integration/Integration.t.sol
andreivladbrg: > Why not add it to `function createDefaultStream()` instead of

because the no-params function always creates the same stream

> IMO it makes it complex because you would assume a helper function takes argument but gets overridden by internal logic

how does it override the internal logic? we want to create a stream with a set of params (no matter the struct name or the function called) and we are doing this 

> t matters because the point of having a param in a function is be able to use that param in the function which is not the case for LPG logic.

the point of having the param is to be able to pass different values than the all the default ones - no matter whether it's `createWithTimestamps` or `createWithDurations`

can we keep the current version?

lockup/tests/fork/LockupPriceGated.t.sol
andreivladbrg: alright

lockup/tests/integration/fuzz/lockup-price-gated/streamedAmountOf.t.sol
andreivladbrg: alright, lets keep the current version

lockup/tests/integration/Integration.t.sol
smol-ninja: OK lets keep the current version

lockup/tests/integration/fuzz/lockup-price-gated/LockupPriceGated.t.sol
andreivladbrg: leaving this like this for now, to merge the PR

lockup/src/abstracts/SablierLockupPriceGated.sol
andreivladbrg: managed to increase the runs back to 500👍

andreivladbrg (approved):
merging 🚀

PR #1413, opened 2026-02-16
refactor: safe oracle split

andreivladbrg:
this PR isolates the changes made to `lockup` and `utils` in my bob review

- rename the previous `SafeOracle.safeOraclePrice` to `SafeOracle.validateOracle`
- add `SafeOracle.safeOraclePrice`, which never reverts, and use it in `LockupMath`

note: lockup's size is still below the margin

image

### DIFF
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -5,6 +5,7 @@ import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { PRBMathCastingUint128 as CastingUint128 } from "@prb/math/src/casting/Uint128.sol";
 import { PRBMathCastingUint40 as CastingUint40 } from "@prb/math/src/casting/Uint40.sol";
 import { SD59x18 } from "@prb/math/src/SD59x18.sol";
+import { SafeOracle } from "@sablier/evm-utils/src/libraries/SafeOracle.sol";
 
 import { LockupDynamic } from "../types/LockupDynamic.sol";
 import { LockupLinear } from "../types/LockupLinear.sol";
@@ -261,19 +262,15 @@ library LockupMath {
             return deposited;
         }
 
-        // Get the current oracle price using try-catch. Because users may not always use Chainlink oracles, we do not
-        // check for the staleness of the price.
-        try unlockParams.oracle.latestRoundData() returns (uint80, int256 price, uint256, uint256, uint80) {
-            // If the oracle price is at or above the target price, return the deposited amount.
-            if (price > 0 && uint256(price).toUint128() >= unlockParams.targetPrice) {
-                return deposited;
-            }
-            // Otherwise, return 0.
-            return 0;
-        } catch {
-            // If the oracle call fails, return 0.
-            return 0;
+        // Get the latest price from the oracle with safety checks.
+        uint128 latestPrice = SafeOracle.safeOraclePrice(unlockParams.oracle);
+
+        // If the oracle price is at or above the target price, return the deposited amount.
+        if (latestPrice > 0 && latestPrice >= unlockParams.targetPrice) {
+            return deposited;
         }
+
+        return 0;
     }
 
     /// @notice Calculates the streamed amount of LT streams.
--- a/lockup/src/libraries/LockupMath.sol
+++ b/lockup/src/libraries/LockupMath.sol
@@ -263,7 +263,7 @@ library LockupMath {
         }
 
         // Get the latest price from the oracle with safety checks.
-        uint128 latestPrice = SafeOracle.safeOraclePrice(unlockParams.oracle);
+        (uint128 latestPrice,) = SafeOracle.safeOraclePrice(unlockParams.oracle);
 
         // If the oracle price is at or above the target price, return the deposited amount.
         if (latestPrice > 0 && latestPrice >= unlockParams.targetPrice) {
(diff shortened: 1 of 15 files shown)


utils/src/libraries/SafeOracle.sol
```diff
@@ -7,12 +7,60 @@ import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
...
+        // Due to reorgs and latency issues, the oracle can have an `updatedAt` timestamp that is in the future. In
+        // this case, we ignore the price and return 0.
+        if (block.timestamp < updatedAt) {
+            return 0;
+        }
+
+        // If the oracle hasn't been updated in the last 24 hours, we ignore the price and return 0. This is a safety
+        // check to avoid using outdated prices.
+        unchecked {
+            if (block.timestamp - updatedAt > 24 hours) {
```
smol-ninja: Is bob using `safeOraclePrice` or only `validateOracle`? If bob is also using `safeOraclePrice`, can we remove this 24 hours restriction? The outdated price is useful in case of fee collection but not in case of Bob.

utils/src/libraries/SafeOracle.sol
```diff
@@ -29,12 +77,12 @@ library SafeOracle {
         }
 
         // Check: oracle returns a positive price when `latestRoundData()` is called.
-        try oracle.latestRoundData() returns (uint80, int256 price, uint256, uint256, uint80) {
+        try oracle.latestRoundData() returns (uint80, int256 _price, uint256, uint256, uint80) {
             // Because users may not always use Chainlink oracles, we do not check for the staleness of the price.
-            if (price <= 0) {
+            if (_price <= 0) {
                 revert Errors.SafeOracle_NegativePrice(address(oracle));
```
smol-ninja: Should we rename it to `SafeOracle_NotPositivePrice`?

lockup/src/libraries/LockupMath.sol
```diff
@@ -261,19 +262,15 @@ library LockupMath {
...
-            // Otherwise, return 0.
-            return 0;
-        } catch {
-            // If the oracle call fails, return 0.
-            return 0;
+        // Get the latest price from the oracle with safety checks.
+        uint128 latestPrice = SafeOracle.safeOraclePrice(unlockParams.oracle);
+
+        // If the oracle price is at or above the target price, return the deposited amount.
+        if (latestPrice > 0 && latestPrice >= unlockParams.targetPrice) {
```
smol-ninja, line 269: IMO we should not allow users to withdraw in case of outdated price. How about we return latest price in `safeOraclePrice` and skip the outdated check for this case?

utils/src/SablierComptroller.sol
```diff
@@ -429,39 +430,14 @@ contract SablierComptroller is
...
-        if (block.timestamp < updatedAt) {
+        // If the price is 0, skip the calculations.
+        if (price == 0) {
             return 0;
         }
 
-        // If the oracle hasn't been updated in the last 24 hours, we ignore the price and return 0. This is a safety
-        // check to avoid using outdated prices.
-        unchecked {
-            if (block.timestamp - updatedAt > 24 hours) {
```
smol-ninja, line 460: So we keep this logic here and remove this check from the SafeOracle library

smol-ninja (changes requested):

utils/src/libraries/SafeOracle.sol
andreivladbrg: > Is bob using `safeOraclePrice` or only `validateOracle`? If bob is also using `safeOraclePrice`, can we remove this 24 hours restriction? The outdated price is useful in case of fee collection but not in case of Bob.

both are used, this is why i added them

i believe having the 24h check is a best practice

utils/src/libraries/SafeOracle.sol
andreivladbrg: sure, both work for me

utils/src/libraries/SafeOracle.sol
smol-ninja: We added 24 hrs check because in case of fee, an outdated price can hurt users as they may end up paying high sablier fees.

In case of vaults, the outdated price checks add no value, rather it can hurt users if price is outdated but above target price. Thats why I dont agree that we should add it here. Imagine oracle returning price over target price but last updated is > 24 hrs. Why shouldn't user be able to withdraw the tokens?

utils/src/libraries/SafeOracle.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1413/changes/544b46e7955ffa4d0b6670e457a1454bfc08de04

smol-ninja (approved):

utils/src/libraries/SafeOracle.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1413/changes/46cc08eda5069aac42fa89988a815ff5bc3bf958

utils/src/SablierComptroller.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1413/changes/46cc08eda5069aac42fa89988a815ff5bc3bf958

lockup/src/libraries/LockupMath.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1413/changes/46cc08eda5069aac42fa89988a815ff5bc3bf958

smol-ninja:
I have removed both timestamp check (outdated price and timestamp in future) from safe oracle library and moved them to the contracts where they are necessary.

Can you please review my commit and if it looks good, lets merge this PR.

andreivladbrg:
looks good, thanks

PR #1415, opened 2026-02-17
Refactor claim tests, add function to query token admin address

smol-ninja:
### Changelog
- DRY'fies claim related tests.
- Fix failing fork tests as a result of using token admin as one of the fuzzed user.

### DIFF
--- a/airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.t.sol
+++ b/airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.t.sol
@@ -6,15 +6,23 @@ import { ISablierMerkleExecute } from "src/interfaces/ISablierMerkleExecute.sol"
 import { MerkleExecute } from "src/types/MerkleExecute.sol";
 
 import { MockStakingReentrant, MockStakingRevert } from "./../../../../../mocks/MockStaking.sol";
-import { MerkleExecute_Integration_Shared_Test } from "./../MerkleExecute.t.sol";
+import { Claim_Integration_Test } from "./../../shared/claim/claim.t.sol";
+import { Integration_Test, MerkleExecute_Integration_Shared_Test } from "./../MerkleExecute.t.sol";
 
-contract ClaimAndExecute_MerkleExecute_Integration_Test is MerkleExecute_Integration_Shared_Test {
-    function setUp() public virtual override {
+contract ClaimAndExecute_MerkleExecute_Integration_Test is
+    Claim_Integration_Test,
+    MerkleExecute_Integration_Shared_Test
+{
+    function setUp() public virtual override(MerkleExecute_Integration_Shared_Test, Integration_Test) {
         MerkleExecute_Integration_Shared_Test.setUp();
 
         // Make the recipient the caller.
         setMsgSender(users.recipient);
     }
+
+    /// @dev This test is not applicable to {MerkleExecute} campaigns since they have {ClaimType.EXECUTE} claim type by
+    /// default.
+    function test_RevertGiven_NotDefaultClaimType() external pure override { }
 
     function test_RevertWhen_TargetTransferAmountOverdraws() external whenMerkleProofValid {
         uint128 overdrawAmount = CLAIM_AMOUNT + 1;
--- a/airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.t.sol
+++ b/airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.t.sol
@@ -6,32 +6,17 @@ import { ISablierMerkleExecute } from "src/interfaces/ISablierMerkleExecute.sol"
 import { MerkleExecute } from "src/types/MerkleExecute.sol";
 
 import { MockStakingReentrant, MockStakingRevert } from "./../../../../../mocks/MockStaking.sol";
-import { Integration_Test } from "./../../../../Integration.t.sol";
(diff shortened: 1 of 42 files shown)


airdrops/src/types/DataTypes.sol
```diff
@@ -7,12 +7,23 @@ import { UD2x18 } from "@prb/math/src/UD2x18.sol";
...
 
+/// @notice Enum representing the type of claim functions supported by a Merkle campaign.
+/// @custom:value0 DEFAULT Activates `claim`, `claimTo`, and `claimViaSig` functions.
+/// @custom:value1 ATTEST Activates `claimViaAttestation` function only.
+/// @custom:value2 EXECUTE Activates `claimAndExecute` function only.
+enum ClaimType {
+    DEFAULT,
+    ATTEST,
+    EXECUTE
+}
```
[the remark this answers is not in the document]
smol-ninja: Thats by design.

airdrops/tests/integration/concrete/campaign/ll/claim/claim.tree
```diff
@@ -1,24 +1,21 @@
...
+   ├── when vesting end time not exceed claim time
+   │  ├── it should transfer the tokens to the recipient
+   │  └── it should emit a {ClaimLLWithTransfer} event
+   └── when vesting end time exceeds claim time
+      ├── when total percentage greater than 100
+      │  └── it should revert
+      └── when total percentage not greater than 100
+         ├── when vesting start time zero
+         │  ├── it should create a stream with block timestamp as vesting start time
+         │  └── it should emit a {ClaimLLWithVesting} event
```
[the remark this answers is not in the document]
smol-ninja: This is acceptable as in case of vesting airdrops, the tokens are sent to Lockup contract to create the stream. Detailed tests can be found in the Lockup contract's tests.

airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.t.sol
```diff
@@ -6,32 +6,17 @@ import { ISablierMerkleExecute } from "src/interfaces/ISablierMerkleExecute.sol"
 import { MerkleExecute } from "src/types/MerkleExecute.sol";
 
 import { MockStakingReentrant, MockStakingRevert } from "./../../../../../mocks/MockStaking.sol";
-import { Integration_Test } from "./../../../../Integration.t.sol";
-import { Claim_Integration_Test } from "./../../shared/claim/claim.t.sol";
 import { MerkleExecute_Integration_Shared_Test } from "./../MerkleExecute.t.sol";
 
-contract ClaimAndExecute_MerkleExecute_Integration_Test is
-    MerkleExecute_Integration_Shared_Test,
-    Claim_Integration_Test
```
andreivladbrg, line 15: why no more shared claim tests for MerkleExecute

andreivladbrg (commented):
generally looks good, one question:

airdrops/tests/integration/concrete/campaign/execute/claim-and-execute/claimAndExecute.t.sol
smol-ninja: `Claim_Integration_Test` has `test_RevertGiven_NotDefaultClaimType` test which is invalid for `claimAndExecute` function. So, the test becomes very unstructured if I try to deal with that.

Its also the same reason that we do not have `Claim_Integration_Test` inherited by `ClaimViaSig_XYZ_Integration_Test`, `ClaimViaAttestation_XYZ_Integration_Test`, and `ClaimTo_XYZ_Integration_Test` as they have their own shared test testing specific logic and skipping the common logic from `_preProcessClaim` which is thoroughly tested in `Claim_Integration_Test`.

So, similar to other tests, `ClaimAndExecute_XYZ_Integration_Test` can test the logic implemented after `_preProcessClaim`.

andreivladbrg: > `Claim_Integration_Test` has `test_RevertGiven_NotDefaultClaimType` test which is invalid for `claimAndExecute` function. So, the test becomes very unstructured if I try to deal with that.

check the merkle type, and if it's `execute`, return - easy, but not great that it would still display the test as passing in the `ClaimAndExecute_MerkleExecute_Integration_Test` contract

> Its also the same reason that we do not have Claim_Integration_Test inherited by ClaimViaSig_XYZ_Integration_Test, ClaimViaAttestation_XYZ_Integration_Test, and ClaimTo_XYZ_Integration_Test as they have their own shared test testing specific logic and skipping the common logic from _preProcessClaim which is thoroughly tested in Claim_Integration_Test.

the difference here is that there is no `claim` where the tests are actually run.
in the case of the other campaigns, the `Claim_Integration_Test`s are run on `claim`, and it's not needed for `claimTo` / `claimViaSig`, etc.

since here there's only one function, `claimAndExecute`, i'd say it's useful to add it and check the campaign type in the claim type test.

wdyt?

smol-ninja: Good point. Addressed in https://github.com/sablier-labs/lockup/pull/1415/changes/2ae56569689e596f95733d89b857b02bdd934a65

smol-ninja:
@andreivladbrg if the above commit looks good, shall we merge it?

andreivladbrg (approved):
> @andreivladbrg if the above commit looks good, shall we merge it?

looks good, thanks

PR #1418, opened 2026-02-19
script: polish bob and comptroller upgrade scripts

smol-ninja:
### Bob

- Removes `deploy-lido-adapter-mainnet` recipe as adapter deployment is now merged with bob deployment script
- `DeployBob` deploys both Bob and Adapter on Ethereum and Sepolia
- Renames `LidoAdapterConstants` to `LidoAdapterUtils`
- Adds `bob` to justfile's `PACKAGES` variable

### Comptroller

- Add instructions for upgrading comptroller proxy


utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
```diff
@@ -16,6 +27,9 @@ contract UpgradeComptrollerProxy is BaseScript {
         // See https://docs.openzeppelin.com/upgrades-plugins/faq#how-can-i-disable-checks
         opts.unsafeAllow = "constructor";
 
+        // Set the reference contract for the storage layout comparison.
+        opts.referenceContract = "SablierComptrollerV1.sol:SablierComptroller";
```
andreivladbrg: shouldn't the version be `V1.0`?

justfile
```diff
@@ -18,7 +18,7 @@ export FOUNDRY_DISABLE_NIGHTLY_WARNING := "true"
 export FOUNDRY_FUZZ_SEED := `echo $(($EPOCHSECONDS / 604800))`
 
 # All monorepo packages
-PACKAGES := "airdrops flow lockup utils"
+PACKAGES := "airdrops bob flow lockup utils"
```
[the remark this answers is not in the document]
andreivladbrg: it imports them from `import "./node_modules/@sablier/devkit/just/evm.just"`

its not a problem

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
```diff
@@ -4,8 +4,19 @@ pragma solidity >=0.8.22;
...
 
-/// @notice Deploys a new Sablier Comptroller.
+/// @notice Upgrades the Sablier Comptroller.
 /// @dev The deployed Sablier Comptroller is set as the implementation of the existing proxy.
+///
+/// The following upgrade script runs a storage collision check between the new implementation contract and the previous
+/// version. The function requires access to the previous version of the contract. Therefore, to perform the upgrade,
+/// follow the steps below:
+/// 1. Flatten the previous version of the Comptroller contract by using the following command on
+/// https://github.com/sablier-labs/evm-utils/blob/main/src/SablierComptroller.sol:
```
andreivladbrg, line 14: this link won't be good in case we upgrade in future versions

for example, we now upgrade to `SablierComptrollerV1.1`, and in the future if we upgrade to `v1.2`, we will need to check the storage from `v1.1` and not `v1.0`

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
smol-ninja: Yeah, thats why V1. What do you suggest?

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
smol-ninja: We can update the link then. What do you suggest we should do here?

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
andreivladbrg: > What do you suggest?

`SablierComptrollerV1.0.sol:SablierComptroller`

beacause `v1` applies to current version as well: `v1.1`

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
andreivladbrg: >What do you suggest we should do here?

link to both `evm-utils` repo, and this (future) monorepo

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
smol-ninja: I see. Should we call it `SablierComptrollerV10` instead? Its quite common naming convention when there is a period in the version number. 

https://github.com/sablier-labs/lockup/pull/1418/changes/f498e7eb9de859ab0ec0910079954ba7c09764b6

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
smol-ninja: I would suggest lets update it during the next comptroller release. The instructions are for the upcoming release and after that we can update it. 

To add more context, I can update it as 

```solidity
/// 1. Flatten the previous version of the Comptroller contract. Current version is v1.0 which can be found at
/// https://github.com/sablier-labs/evm-utils/blob/main/src/SablierComptroller.sol:
```

And then later on, we can just remove this and point it to the monorepo.

https://github.com/sablier-labs/lockup/pull/1418/changes/f498e7eb9de859ab0ec0910079954ba7c09764b6

andreivladbrg: alright

utils/scripts/solidity/UpgradeComptrollerProxy.s.sol
andreivladbrg: looks good

andreivladbrg (approved):

PR #1419, opened 2026-02-20
test(bob-escrow): fork and fuzz

andreivladbrg:


bob/tests/escrow/integration/fuzz/cancelOrder.t.sol
```diff
@@ -0,0 +1,76 @@
...
+        minBuyAmount = boundUint128(minBuyAmount, 1, MAX_UINT128);
+        expiryTime = boundUint40(expiryTime, getBlockTimestamp() + 1, getBlockTimestamp() + 5 * 365 days);
+        timeJump = boundUint40(timeJump, 0, uint40(10 * 365 days));
+
+        // Deal sell tokens to seller.
+        deal({ token: address(sellToken), to: users.seller, give: sellAmount });
+
+        // Create the order as seller.
+        setMsgSender(users.seller);
+        uint256 orderId = escrow.createOrder({
```
smol-ninja: I think this is an overkill. We should just test cancelling order based on timestamp warp.

```solidity
function testFuzz_CancelOrder(uint64 warpTime)
```

where we warp time to a random value between order creation to until 1 month after expiry. And the fuzz test should allow cancelling it.

bob/tests/escrow/integration/fuzz/createOrder.t.sol
```diff
@@ -0,0 +1,66 @@
...
+    function testFuzz_CreateOrder(uint128 sellAmount, uint128 minBuyAmount, uint40 expiryTime) external {
+        sellAmount = boundUint128(sellAmount, 1, MAX_UINT128);
+        minBuyAmount = boundUint128(minBuyAmount, 1, MAX_UINT128);
+
+        // If non-zero, bound to [now+1, now+5years]; otherwise keep 0 (never expires).
+        if (expiryTime != 0) {
+            expiryTime = boundUint40(expiryTime, getBlockTimestamp() + 1, getBlockTimestamp() + 5 * 365 days);
+        }
+
+        // Deal sell tokens to seller so any uint128 amount works.
```
smol-ninja: What does "so any uint128 amount works" mean?

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
```diff
@@ -0,0 +1,182 @@
...
+
+import { ISablierEscrow } from "src/interfaces/ISablierEscrow.sol";
+import { Errors } from "src/libraries/Errors.sol";
+import { Escrow } from "src/types/Escrow.sol";
+
+import { Integration_Test } from "./../Integration.t.sol";
+
+contract FillOrder_Integration_Fuzz_Test is Integration_Test {
+    /// @dev Designated buyer, default trade fee (1%).
+    function testFuzz_FillOrder(
```
smol-ninja: It should also just have fuzzed timestamp and a fuzzed amount. The test should test that

- Any buyer can fill order until expiry with fuzzed amount >= min amount. So this test would have a fuzzed buyer, fuzzed amount and a fuzzed timestamp.
- If order has designated buyer, no buyer can fill the order. So that would have a fuzzed buyer address, a fuzzed timestamp and a fuzzed amount.

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
```diff
@@ -0,0 +1,182 @@
...
+            buyAmount: buyAmount,
+            expiryTime: expiryTime,
+            timeJump: timeJump,
+            buyer: users.buyer,
+            fee: DEFAULT_TRADE_FEE
+        });
+    }
+
+    /// @dev Designated buyer, zero trade fee — exercises the `if (currentTradeFee > 0)` skip branch.
+    function testFuzz_FillOrder_GivenTradeFeeZero(
```
smol-ninja: Why do we need a test with trade fee as zero? Isn't a test with default trade fee but different buy amount enough?

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
```diff
@@ -0,0 +1,182 @@
...
+            buyAmount: buyAmount,
+            expiryTime: expiryTime,
+            timeJump: timeJump,
+            buyer: users.buyer,
+            fee: ZERO
+        });
+    }
+
+    /// @dev Open order (buyer = address(0)), default trade fee — exercises the open-order access path.
+    function testFuzz_FillOrder_GivenOpenOrder(
```
smol-ninja: Open order is confusing here since there is a status called `OPEN`. We use the term no designated buyer for this

smol-ninja (changes requested):

bob/tests/escrow/integration/fuzz/cancelOrder.t.sol
andreivladbrg: your point is that we should use the default order?

bob/tests/escrow/integration/fuzz/createOrder.t.sol
andreivladbrg: am not sure - claude added it xD

my guess is overflow 

we can remove it

bob/tests/escrow/integration/fuzz/cancelOrder.t.sol
smol-ninja: Yes and fuzz timestamp only. Because for an open order, cancel should work always. Creating order with different value has no effect on it.

andreivladbrg: the reason i've created custom orders is that, unlike in flow where we have the [fixtures](https://github.com/sablier-labs/lockup/blob/staging/flow/tests/integration/fuzz/Fuzz.t.sol#L19) (defaults + custom ones), i thought having only new orders would be useful

wdyt?

smol-ninja: `cancel` only has one check for `OPEN` order: "revert if `msg.sender != order.seller`". It does not depend on other parameters of the order. 

Therefore, given our style for fuzz tests in other protocols, it only makes sense to have two tests (1) `testFuzz_ReverWhen_CallerNotSeller(address caller)` (2) `testFuzz_CancelOrder(uint64 warpTime)` and then just use default order.

[An example from Lockup](https://github.com/sablier-labs/lockup/blob/staging/lockup/tests/integration/fuzz/lockup/cancel.t.sol).

andreivladbrg: > `cancel` only has one check for `OPEN` order: "revert if `msg.sender != order.seller`". It does not depend on other parameters of the order.

ik ik. 

i don't have a strong preference - i used this approach because we have in solana as well - i tested the waters here xD

would u like to make the change?

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
andreivladbrg: i added because of this check:

https://github.com/sablier-labs/lockup/blob/3051b8594b55f05652f8cdd757d8f2f873f40f42/bob/src/SablierEscrow.sol#L207

i think its useful to keep as there less operations done in comparison with `fee > 0`

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
andreivladbrg: i am not sure i understand your point

that’s what it should test - `fillOrder` when the status is `OPEN`

smol-ninja: `OPEN` status means order can be filled.  In the comment, open implies `buyer = address(0)`. Both have different meanings.

bob/tests/escrow/integration/fuzz/cancelOrder.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1419/changes/160dc6aa6ee1a2ea32a8c6a219e158f844493a1f

I also added status to some modifiers to make them explicit about it.

bob/tests/escrow/integration/fuzz/createOrder.t.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1419/changes/160dc6aa6ee1a2ea32a8c6a219e158f844493a1f

bob/tests/escrow/fork/Fork.t.sol
```diff
@@ -0,0 +1,23 @@
...
+import { Base_Test } from "./../Base.t.sol";
+
+/// @notice Base logic needed by the fork tests.
+abstract contract Fork_Test is Base_Test {
+    IERC20 internal constant USDC = IERC20(0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48); // sell token
+    IERC20 internal constant WETH = IERC20(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2); // buy token
+
+    function setUp() public virtual override {
+        // Fork Ethereum Mainnet at the latest block number.
+        vm.createSelectFork({ urlOrAlias: "ethereum" });
```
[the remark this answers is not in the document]
smol-ninja: Its OK to fork it at latest block. This helps us to test protocol against the latest fork.

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
smol-ninja: OK sounds good. But we should refactor the test to simplify them a bit similar to cancel order. See [this comment](https://github.com/sablier-labs/lockup/pull/1419#discussion_r2864926192).

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
andreivladbrg: right, iget it now

97b7ef0d21168bfc4ed6b6b524bd0786c846621b

bob/tests/escrow/integration/fuzz/fillOrder.t.sol
andreivladbrg: f3e9a42ab1f5dd411ffd5fd90136a734da5c7bd0

andreivladbrg:
@smol-ninja does it look good now?

smol-ninja (approved):
LGTM.

PR #1421, opened 2026-02-24
feat: add initial bob min fee in usd

andreivladbrg:
refactor: change the min fee usd in deployment scripts to $0.99 chore: remove unused custome error


utils/src/libraries/Errors.sol
```diff
@@ -33,9 +33,6 @@ library Errors {
     /// @notice Thrown when the provided campaign address returns false for `IS_SABLIER_MERKLE()`.
     error SablierComptroller_IsSablierMerkleReturnsFalse(address campaign);
```
smol-ninja: This is also unused. Can you verify and remove it?

smol-ninja (changes requested):

utils/src/libraries/Errors.sol
andreivladbrg: 8108094079a4b993007d3ea6e0c61bf15f1829a9

smol-ninja (approved):

PR #1422, opened 2026-02-25
perf: remove duplicated check

andreivladbrg:


andreivladbrg:
@smol-ninja ignore the failing test - is not related to this PR , and its going to be fixed in the upcoming test PR

andreivladbrg:
nvm, closing as this change is it not needed xD 
we need the double check

smol-ninja:
> we need the double check

Yeah. I got so confused as I (as in I) added it myself and not using some AI which means less likely it was a mistake 😆

PR #1423, opened 2026-02-27
Audit fixes - cantina

andreivladbrg:
Addresses:
- [finding 7](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/7)
- [finding 11](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/11)
- [finding 18](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/18)
- [finding 19](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/19)
- [finding 21](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/21)

### DIFF
--- a/lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
+++ b/lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
@@ -28,12 +28,7 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         });
     }
 
-    function test_RevertWhen_AmountLessThanDeposit()
-        external
-        givenEndTimeInFuture
-        whenLatestPriceNotBelowTarget
-        whenAmountLessThanDeposit
-    {
+    function test_RevertWhen_AmountLessThanDeposit() external givenEndTimeInFuture whenLatestPriceNotBelowTarget {
         // Update oracle to return target price so the stream unlocks.
         oracle.setPrice(defaults.LPG_TARGET_PRICE());
 
@@ -55,12 +50,7 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         });
     }
 
-    function test_WhenAmountEqualsDeposit()
-        external
-        givenEndTimeInFuture
-        whenLatestPriceNotBelowTarget
-        whenAmountEqualsDeposit
-    {
+    function test_WhenAmountEqualsDeposit() external givenEndTimeInFuture whenLatestPriceNotBelowTarget {
         // Update oracle to return target price.
         oracle.setPrice(defaults.LPG_TARGET_PRICE());
 
@@ -90,7 +80,8 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         assertEq(actualStatus, expectedStatus);
     }
 
-    function test_GivenEndTimeNotInFuture() external givenEndTimeNotInFuture {
+    function test_GivenEndTimeNotInFuture() external {
+        vm.warp({ newTimestamp: defaults.END_TIME() });
         uint128 withdrawAmount = defaults.DEPOSIT_AMOUNT();
 
         // It should emit {WithdrawFromLockupStream} and {MetadataUpdate} events.
--- a/lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
+++ b/lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
@@ -15,7 +15,10 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         uint128 withdrawableAmount = 0;
         vm.expectRevert(
             abi.encodeWithSelector(
-                Errors.SablierLockup_Overdraw.selector, ids.defaultStream, withdrawAmount, withdrawableAmount
+                Errors.SablierLockup_WithdrawAmountNotEqualWithdrawableAmount.selector,
+                ids.defaultStream,
+                withdrawAmount,
+                withdrawableAmount
             )
         );
         lockup.withdraw{ value: LOCKUP_MIN_FEE_WEI }({
@@ -25,7 +28,39 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         });
     }
 
-    function test_WhenLatestPriceNotBelowTarget() external givenEndTimeInFuture {
+    function test_RevertWhen_AmountLessThanDeposit()
+        external
+        givenEndTimeInFuture
+        whenLatestPriceNotBelowTarget
+        whenAmountLessThanDeposit
+    {
+        // Update oracle to return target price so the stream unlocks.
+        oracle.setPrice(defaults.LPG_TARGET_PRICE());
+
+        uint128 withdrawableAmount = defaults.DEPOSIT_AMOUNT();
+        uint128 withdrawAmount = withdrawableAmount - 1;
+
+        vm.expectRevert(
+            abi.encodeWithSelector(
+                Errors.SablierLockup_WithdrawAmountNotEqualWithdrawableAmount.selector,
+                ids.defaultStream,
+                withdrawAmount,
+                withdrawableAmount
+            )
+        );
+        lockup.withdraw{ value: LOCKUP_MIN_FEE_WEI }({
+            streamId: ids.defaultStream,
+            to: users.recipient,
+            amount: withdrawAmount
+        });
+    }
+
+    function test_WhenAmountEqualsDeposit()
+        external
+        givenEndTimeInFuture
+        whenLatestPriceNotBelowTarget
+        whenAmountEqualsDeposit
+    {
         // Update oracle to return target price.
         oracle.setPrice(defaults.LPG_TARGET_PRICE());
 
@@ -55,37 +90,7 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         assertEq(actualStatus, expectedStatus);
     }
 
-    function test_WhenLatestPriceBelowTarget() external givenEndTimeNotInFuture {
-        uint128 withdrawAmount = defaults.DEPOSIT_AMOUNT();
-
-        // It should emit {WithdrawFromLockupStream} and {MetadataUpdate} events.
-        vm.expectEmit({ emitter: address(lockup) });
-        emit ISablierLockup.WithdrawFromLockupStream({
-            streamId: ids.defaultStream,
-            to: users.recipient,
-            token: dai,
-            amount: withdrawAmount
-        });
-        vm.expectEmit({ emitter: address(lockup) });
-        emit IERC4906.MetadataUpdate({ _tokenId: ids.defaultStream });
-
-        // Make the withdrawal.
-        lockup.withdraw{ value: LOCKUP_MIN_FEE_WEI }({
-            streamId: ids.defaultStream,
-            to: users.recipient,
-            amount: withdrawAmount
-        });
-
-        // It should mark the stream as depleted.
-        Lockup.Status actualStatus = lockup.statusOf(ids.defaultStream);
-        Lockup.Status expectedStatus = Lockup.Status.DEPLETED;
-        assertEq(actualStatus, expectedStatus);
-    }
-
-    function test_WhenLatestPriceNotBelowTarget_GivenEndTimeNotInFuture() external givenEndTimeNotInFuture {
-        // Update oracle to return target price.
-        oracle.setPrice(defaults.LPG_TARGET_PRICE());
-
+    function test_GivenEndTimeNotInFuture() external givenEndTimeNotInFuture {
         uint128 withdrawAmount = defaults.DEPOSIT_AMOUNT();
 
         // It should emit {WithdrawFromLockupStream} and {MetadataUpdate} events.
--- a/lockup/tests/invariant/handlers/LockupHandler.sol
+++ b/lockup/tests/invariant/handlers/LockupHandler.sol
@@ -154,7 +154,7 @@ contract LockupHandler is BaseHandler {
         vm.assume(withdrawableAmount != 0);
 
         // If the model is LPG, the withdraw amount must equal the withdrawable amount.
-        if (uint8(lockup.getLockupModel(currentStreamId)) == uint8(Lockup.Model.LOCKUP_PRICE_GATED)) {
+        if (lockup.getLockupModel(currentStreamId) == Lockup.Model.LOCKUP_PRICE_GATED) {
             withdrawAmount = withdrawableAmount;
         }
         // Otherwise, bound the withdraw amount so that it is not zero.
--- a/lockup/tests/invariant/handlers/LockupHandler.sol
+++ b/lockup/tests/invariant/handlers/LockupHandler.sol
@@ -153,8 +153,14 @@ contract LockupHandler is BaseHandler {
         uint128 withdrawableAmount = lockup.withdrawableAmountOf(currentStreamId);
         vm.assume(withdrawableAmount != 0);
 
-        // Bound the withdraw amount so that it is not zero.
-        withdrawAmount = boundUint128(withdrawAmount, 1, withdrawableAmount);
+        // If the model is LPG, the withdraw amount must equal the withdrawable amount.
+        if (uint8(lockup.getLockupModel(currentStreamId)) == uint8(Lockup.Model.LOCKUP_PRICE_GATED)) {
+            withdrawAmount = withdrawableAmount;
+        }
+        // Otherwise, bound the withdraw amount so that it is not zero.
+        else {
+            withdrawAmount = boundUint128(withdrawAmount, 1, withdrawableAmount);
+        }
 
         // There is an edge case when the sender is the same as the recipient. In this scenario, the withdrawal
         // address must be set to the recipient.
--- a/utils/src/SablierComptroller.sol
+++ b/utils/src/SablierComptroller.sol
@@ -460,7 +460,7 @@ contract SablierComptroller is
             return 0;
         }
 
-        // Get the latest price and feed updated timestamp from the oracle.
+        // Get the latest price (normalized to 8 decimals) and feed updated timestamp from the oracle.
         (uint256 price, uint256 updatedAt) = SafeOracle.safeOraclePrice(AggregatorV3Interface(oracle));
 
         // Skip the calculations if any of the following conditions are met:
@@ -470,25 +470,8 @@ contract SablierComptroller is
             return 0;
         }
 
-        uint8 oracleDecimals;
-
-        // Interactions: query the oracle decimals.
-        try AggregatorV3Interface(oracle).decimals() returns (uint8 _decimals) {
-            oracleDecimals = _decimals;
-        } catch {
-            // If the oracle call fails, return 0.
-            return 0;
-        }
-
-        // If the oracle decimals are 8, calculate the minimum fee in wei.
-        if (oracleDecimals == 8) {
-            return minFeeUSD * 1e18 / uint256(price);
-        }
-        // Otherwise, adjust the calculation to account for the oracle decimals.
-        else {
-            // The USD fee is assumed to be much less than the maximum value of `uint256` so it is safe to multiply.
-            return minFeeUSD * 10 ** (10 + oracleDecimals) / uint256(price);
-        }
+        // Calculate the minimum fee in wei.
+        return minFeeUSD * 1e18 / price;
     }
 
     /// @dev See the documentation for the user-facing functions that call this private function.
--- a/utils/src/libraries/SafeOracle.sol
+++ b/utils/src/libraries/SafeOracle.sol
@@ -11,30 +11,45 @@ import { Errors } from "./Errors.sol";
 library SafeOracle {
     using SafeCast for uint256;
 
-    /// @dev Fetches the latest price from the oracle without reverting. Returns 0 if any of the following conditions
-    /// are met:
+    /*//////////////////////////////////////////////////////////////////////////
+                            INTERNAL READ-ONLY FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Fetches the latest price from the oracle, normalized to 8 decimals. Returns 0 if any of the following
+    /// conditions are met:
     /// - Oracle address is zero.
     /// - Call to `latestRoundData()` fails.
+    /// - Call to `decimals()` fails.
+    /// - The `decimals()` call returns 0 or more than 36 decimals.
     /// - The price is not positive.
     /// - The price exceeds `uint128` max.
     /// - The `updatedAt` timestamp is in the future.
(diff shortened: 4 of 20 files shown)


utils/src/libraries/SafeOracle.sol
```diff
@@ -49,24 +55,52 @@ library SafeOracle {
...
+        uint8 oracleDecimals;
+        try oracle.decimals() returns (uint8 _decimals) {
+            // Check: decimals is not zero.
+            if (_decimals == 0) {
+                revert Errors.SafeOracle_DecimalsZero(address(oracle));
+            }
+
+            // Check: decimals is not too high to avoid overflow in normalization.
+            if (_decimals > 36) {
+                revert Errors.SafeOracle_DecimalsTooHigh(address(oracle), _decimals);
```
andreivladbrg, line 109: added this check (no higher than 36) because the normalization would overflow otherwise here

```solidity
return price / uint128(10 ** (oracleDecimals - 8));
```

and since the oracle is an input param i believe its useful, also this limit is reasonable IMO

utils/src/SablierComptroller.sol
andreivladbrg, line 537: should we call `SafeOracle.validateOracle()` here as well for consistency?

smol-ninja: Yes we should.

andreivladbrg: actually, we want to have the possibility of setting the oracle to `address(0)` - thus we can’t use it here, as it would revert

smol-ninja: Oh right.

utils/src/libraries/SafeOracle.sol
```diff
@@ -33,53 +42,103 @@ library SafeOracle {
...
             }
 
-            // Due to reorgs and latency issues, the oracle can have an `updatedAt` timestamp in the future.
-            if (block.timestamp < _updatedAt) {
+            normalizedPrice = uint256(_price);
+
+            // If the initial price is greater than max `uint128`, return 0 for price.
+            if (normalizedPrice > type(uint128).max) {
                 return (0, _updatedAt);
             }
```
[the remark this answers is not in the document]
andreivladbrg: this is intended

we need the double check of the `uint128` max because we don’t want to revert at all in the `safeOraclePrice` function  

and if the price is close to `uint256` max, it can overflow in the `_normalizePrice` function  

in practice, the price can be higher than `uint128` and still not revert (`type(int256).max / 10 ** 7`), but we use this margin conservatively

andreivladbrg:
@smol-ninja requested a review. each commit addresses a unique finding. if changes need to be made, we should checkout to the specific commit and make them there and force push

that’s only if they’re not _only_ test-related - if they are, we can push directly

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
```diff
@@ -15,7 +15,10 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         uint128 withdrawableAmount = 0;
         vm.expectRevert(
             abi.encodeWithSelector(
-                Errors.SablierLockup_Overdraw.selector, ids.defaultStream, withdrawAmount, withdrawableAmount
+                Errors.SablierLockup_WithdrawAmountNotEqualWithdrawableAmount.selector,
```
smol-ninja, line 18: In this case, it should revert with `SablierLockup_Overdraw` and not with `SablierLockup_WithdrawAmountNotEqualWithdrawableAmount`. withdraw amount is indeed equal to withdrawable amount as both of them are zero.

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
```diff
@@ -25,37 +28,42 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         });
     }
 
-    function test_WhenLatestPriceNotBelowTarget() external givenEndTimeInFuture {
-        // Update oracle to return target price.
+    function test_RevertWhen_AmountLessThanDeposit()
+        external
+        givenEndTimeInFuture
+        whenLatestPriceNotBelowTarget
+        whenAmountLessThanDeposit
```
smol-ninja: AI slop xD

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
```diff
@@ -25,37 +28,42 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
...
-        Lockup.Status expectedStatus = Lockup.Status.DEPLETED;
-        assertEq(actualStatus, expectedStatus);
     }
 
-    function test_WhenLatestPriceBelowTarget() external givenEndTimeNotInFuture {
+    function test_WhenAmountEqualsDeposit()
+        external
+        givenEndTimeInFuture
+        whenLatestPriceNotBelowTarget
+        whenAmountEqualsDeposit
```
smol-ninja: AI slop

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
```diff
@@ -82,10 +90,7 @@ contract Withdraw_Lockup_PriceGated_Integration_Concrete_Test is Lockup_PriceGat
         assertEq(actualStatus, expectedStatus);
     }
 
-    function test_WhenLatestPriceNotBelowTarget_GivenEndTimeNotInFuture() external givenEndTimeNotInFuture {
-        // Update oracle to return target price.
-        oracle.setPrice(defaults.LPG_TARGET_PRICE());
-
+    function test_GivenEndTimeNotInFuture() external givenEndTimeNotInFuture {
```
smol-ninja: AI slop extra modifier

lockup/tests/invariant/handlers/LockupHandler.sol
```diff
@@ -153,8 +153,14 @@ contract LockupHandler is BaseHandler {
         uint128 withdrawableAmount = lockup.withdrawableAmountOf(currentStreamId);
         vm.assume(withdrawableAmount != 0);
 
-        // Bound the withdraw amount so that it is not zero.
-        withdrawAmount = boundUint128(withdrawAmount, 1, withdrawableAmount);
+        // If the model is LPG, the withdraw amount must equal the withdrawable amount.
+        if (uint8(lockup.getLockupModel(currentStreamId)) == uint8(Lockup.Model.LOCKUP_PRICE_GATED)) {
```
smol-ninja: Do we need to cast `uint8` here? I wonder if 

```solidity
if (_streams[streamId].lockupModel == Lockup.Model.LOCKUP_PRICE_GATED)
```

works why shouldn't it work here?

lockup/tests/invariant/handlers/LockupHandler.sol
```diff
@@ -153,8 +153,14 @@ contract LockupHandler is BaseHandler {
...
+            withdrawAmount = withdrawableAmount;
```
smol-ninja, line 158: We should use `withdrawAmount = deposit amount` for explicitness.

utils/src/libraries/SafeOracle.sol
```diff
@@ -11,19 +11,28 @@ import { Errors } from "./Errors.sol";
...
     function safeOraclePrice(AggregatorV3Interface oracle) internal view returns (uint128 price, uint256 updatedAt) {
         // If the oracle is not set, return 0 for both price and updated timestamp.
         if (address(oracle) == address(0)) {
             return (0, 0);
         }
 
+        uint256 normalizedPrice;
+
         // Interactions: query the oracle price and the time at which it was updated.
         try AggregatorV3Interface(oracle).latestRoundData() returns (
```
smol-ninja:
```suggestion
        try oracle.latestRoundData() returns (
```

`oracle` is already of `AggregatorV3Interface` type.

utils/src/libraries/SafeOracle.sol
```diff
@@ -33,53 +42,103 @@ library SafeOracle {
...
+
+        // Normalize the price to 8 decimals and safe cast to uint128.
+        price = _normalizePrice(normalizedPrice, oracleDecimals).toUint128();
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                            PRIVATE READ-ONLY FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Normalizes a price to 8 decimals given the oracle's native decimal count.
```
smol-ninja:
```suggestion
    /// @dev Normalizes the price from oracle decimals to 8 decimals.
```

Not sure `given the oracle's native decimal count.` mean here?

utils/src/libraries/SafeOracle.sol
```diff
@@ -33,53 +42,103 @@ library SafeOracle {
...
+    function _normalizePrice(uint256 price, uint8 oracleDecimals) private pure returns (uint256) {
+        if (oracleDecimals > 8) {
+            return price / (10 ** (oracleDecimals - 8));
+        } else if (oracleDecimals < 8) {
```
smol-ninja: Since we return in each of the `if` branch, we can just use `if` instead of `else if`

utils/src/libraries/SafeOracle.sol
```diff
@@ -33,53 +42,103 @@ library SafeOracle {
...
 
-            price = uint128(uint256(_price));
             updatedAt = _updatedAt;
         } catch {
             // If the oracle call fails, return 0 for both price and updated timestamp.
             return (0, 0);
         }
+
+        // Normalize the price to 8 decimals.
+        try oracle.decimals() returns (uint8 oracleDecimals) {
```
smol-ninja: We should move it over `latestRoundData` call since its less complicated and checking for decimals would lead us to revert if decimals throw error.

utils/src/libraries/SafeOracle.sol
```diff
@@ -33,53 +42,103 @@ library SafeOracle {
...
                 revert Errors.SafeOracle_NotPositivePrice(address(oracle));
             }
-            price = uint256(_price).toUint128();
+            normalizedPrice = uint256(_price);
         } catch {
             revert Errors.SafeOracle_MissesInterface(address(oracle));
         }
+
+        // Normalize the price to 8 decimals and safe cast to uint128.
+        price = _normalizePrice(normalizedPrice, oracleDecimals).toUint128();
```
smol-ninja: If `normalizedPrice` exceeds `uint128` it would revert. So why dont we move 

```solidity
// If the normalized price is greater than max `uint128`, return 0 for price.
if (normalizedPrice > type(uint128).max) {
    return (0, updatedAt);
}
```

inside `function _normalizePrice`?

utils/src/libraries/SafeOracle.sol
```diff
@@ -33,53 +42,103 @@ library SafeOracle {
...
 
         // Check: oracle returns a positive price when `latestRoundData()` is called.
+        uint256 normalizedPrice;
         try oracle.latestRoundData() returns (uint80, int256 _price, uint256, uint256, uint80) {
             // Because users may not always use Chainlink oracles, we do not check for the staleness of the price.
             if (_price <= 0) {
                 revert Errors.SafeOracle_NotPositivePrice(address(oracle));
             }
-            price = uint256(_price).toUint128();
+            normalizedPrice = uint256(_price);
```
smol-ninja: It can be zero too if price is such that price / division is zero. So we should move this over `if (_price <= 0) {`. See how I did in my gist.

smol-ninja (changes requested):
Nice work on the PR. I am leaving some comments below. The only change I would have made is to `SafeOracle` but since I am not sure if you would agree, I am leaving the code below. Please have a look and if you like the small change, feel free to add it to the PR.

https://gist.github.com/smol-ninja/ec49b85320930e10b435a409a3172a17

utils/src/libraries/SafeOracle.sol
andreivladbrg: your point is

```diff
 function _normalizePrice(uint256 price, uint8 oracleDecimals) private pure returns (uint256) {
     if (oracleDecimals > 8) {
         return price / (10 ** (oracleDecimals - 8));
-    } else if (oracleDecimals < 8) {
+    }  
+
+   if (oracleDecimals < 8) {
         return price * (10 ** (8 - oracleDecimals));
     }
+
     return price;
 }
```

?
sure, same thing for me

utils/src/libraries/SafeOracle.sol
andreivladbrg: >lead us to revert if decimals throw error.

i don't get this 

if we move this above the `latestRoundData` call - we would need to declare one more local variable in the function (decimals) because we are doing this in the try catch block:

https://github.com/sablier-labs/lockup/blob/adb8c09b81e9a591d08f64e804d389faac9519c5/utils/src/libraries/SafeOracle.sol#L69

lockup/tests/invariant/handlers/LockupHandler.sol
andreivladbrg: we don't - we can remove the casting

lockup/tests/invariant/handlers/LockupHandler.sol
andreivladbrg: `withdrawableAmount`  equals `deposited` - its pretty clear that for LPG, it either 0 or deposited IMO

i followed the check in the `src`

utils/src/libraries/SafeOracle.sol
andreivladbrg: good catch

utils/src/libraries/SafeOracle.sol
andreivladbrg: right, the version from the gist doesn't take into account passing to `_normalizePrice` a negative value

we should check if the price is negative before calling `_normalizePrice` and in `safeOraclePrice`we can simply keep the current version as if it becomes 0 after normalization there is no problem, it would return zero

utils/src/libraries/SafeOracle.sol
andreivladbrg: my goal was for `safeOraclePrice` to never revert, and it’s currently not possible because of [this](https://github.com/sablier-labs/lockup/pull/1423#discussion_r2877780662)

and to revert in `validateOracle` if it overflows `uint128`, for explicitness

the version from the gist is also functional, but it reverts with `SafeOracle_NotPositivePrice` if it overflows `uint128` which is not ideal IMO
and if the price returned by `latestRoundData` is uint256 max it still can overflow in [here](https://gist.github.com/smol-ninja/ec49b85320930e10b435a409a3172a17#file-safeoracle-sol-L136) before the check

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: whats the issue with `SablierLockup_WithdrawAmountNotEqualWithdrawableAmount` ?

this is also correct, right?

"Overdraw" is no longer checked in case of LPG, one less check

smol-ninja: What do you think about renaming `SablierLockup_WithdrawAmountNotEqualWithdrawableAmount` to `SablierLockup_WithdrawAmountNotEqualDepositAmount`? The error is wrong because `withdraw amount == withdrawable amount == 0`.

Not a very strong opinion, just checking your view on this.

lockup/tests/invariant/handlers/LockupHandler.sol
smol-ninja: Its not clear I suppose. It says "If the model is LPG, the withdraw amount must equal the withdrawable amount" which is true when withdrawable amount is 0. What do you think?

utils/src/libraries/SafeOracle.sol
smol-ninja: Yeah. sort of.

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: > The error is wrong because withdraw amount == withdrawable amount == 0.

then, it would revert with `SablierLockup_WithdrawAmountZero`, right?

> SablierLockup_WithdrawAmountNotEqualDepositAmount

not against it, but seeing the code checking the withdrawable amount made me think its clearer

both  work, so whatever you think is better

utils/src/libraries/SafeOracle.sol
smol-ninja: > i don't get this

I meant if decimals are invalid, we will revert before `latestRoundData`. The benefit is that, normalization requires price and decimals both. And having checked decimals early, we can move `_normalizePrice` inside `latestRoundData`.

utils/src/libraries/SafeOracle.sol
smol-ninja: > but it reverts with SafeOracle_NotPositivePrice if it overflows uint128 which is not ideal IMO

Thats a good point. My goal was to optimize `SafeOracle` a bit I found reading the code a bit difficult. But I agree to your point here.

utils/src/libraries/SafeOracle.sol
smol-ninja: OK.

I was thinking what do you think if we dry'ify the library such that instead of reverting with custom errors for all cases, we just revert with some common `_Invalid` error. The benefit would be we will be able to dry'ify `try oracle.decimals()` too. We can create a private function for `try oracle.decimals()` logic that returns `0` if decimals are out of ranges (0, 36) and then inside validate and oracle price function, we can revert with `Decimals_Invalid`. 

There is a lot of duplicate logic between the two functions. The only difference is that one reverts while other returns 0.

utils/src/libraries/SafeOracle.sol
andreivladbrg: right, i agree to first query the decimals

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
smol-ninja: > then, it would revert with `SablierLockup_WithdrawAmountZero`, right?

Yes that would be preferrable.

> not against it, but seeing the code checking the withdrawable amount made me think its clearer

Sounds good then. Lets keep it this way.

utils/src/libraries/SafeOracle.sol
andreivladbrg: fair point about the duplication and i agree thats true, but i kept the current version because we would not be able to include the decimals in the custom error if we returned 0 in the private function. and in the case where the interface does not implement the function, we would revert with `SafeOracle_InvalidDecimals` instead of `SafeOracle_MissesInterface`, which is misleading because it makes it seem like the function is implemented but returns an incorrect value

wdyt ?

smol-ninja: thats a good point. Reminds me of why we didn't dry'ify merkle contracts.

utils/src/libraries/SafeOracle.sol
andreivladbrg: forced pushed the commit 260ffc4b301c12831b8bbaeebc2f4cf1304f4625

if it looks feel free to resolve this

utils/src/libraries/SafeOracle.sol
andreivladbrg: forced pushed the commit 260ffc4b301c12831b8bbaeebc2f4cf1304f4625

if it looks feel free to resolve this

utils/src/libraries/SafeOracle.sol
andreivladbrg: forced pushed the commit 260ffc4b301c12831b8bbaeebc2f4cf1304f4625

if it looks feel free to resolve this

utils/src/libraries/SafeOracle.sol
andreivladbrg: 260ffc4b301c12831b8bbaeebc2f4cf1304f4625

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: 7238de8b94bbca71333b521d075cc6db35f944f9

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: 7238de8b94bbca71333b521d075cc6db35f944f9

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: 7238de8b94bbca71333b521d075cc6db35f944f9

lockup/tests/invariant/handlers/LockupHandler.sol
andreivladbrg: 7238de8b94bbca71333b521d075cc6db35f944f9

andreivladbrg:
@smol-ninja addressed all comments - if it looks good, could you mark them as resolved and approve?

also, i merged your changes into this commit: https://github.com/sablier-labs/lockup/pull/1423/commits/4cd8705e476e2a48c3d39561d123c1835762bb7b

they come from [your PR](https://github.com/sablier-labs/lockup/pull/1429/changes/936dd69f266496f6c27b7a66d522ce1a80efff2d), and i added you as co-author

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
smol-ninja: This is not addressed in the commit

andreivladbrg: 0e272a3949eac1f659f39cf62303b9dac953a427

lockup/tests/invariant/handlers/LockupHandler.sol
andreivladbrg: 0e272a3949eac1f659f39cf62303b9dac953a427

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: 0e272a3949eac1f659f39cf62303b9dac953a427

lockup/tests/integration/concrete/lockup-price-gated/withdraw/withdraw.t.sol
andreivladbrg: 0e272a3949eac1f659f39cf62303b9dac953a427

smol-ninja (changes requested):
Thanks for the commits. LGTM. Lets merge it after you have responded to my comment on the Slack.

PR #1424, opened 2026-02-27
refactor: add `SafeTokenSymbol` library in utils

smol-ninja:
### Changelog

- Move duplicated `safeTokenSymbol` and `isAllowedCharacter` from `SablierBob` and `LockupNFTDescriptor` into a shared `SafeTokenSymbol` library in utils.
- Relocate local mock contracts from `utils/src/mocks` to `utils/tests/mocks`.


smol-ninja:
Good news that coverage has gone up to 99.30% (+1.15%).

utils/src/libraries/SafeTokenSymbol.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity >=0.8.22;
+
+import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+
+/// @title SafeTokenSymbol
+/// @notice Library with helper functions for safely retrieving and validating ERC-20 token symbols.
```
andreivladbrg: i believe "validating" is not a good word here since we do not revert  anywhere


```suggestion
/// @notice Library with helper functions for safely retrieving an ERC-20 token symbols, preventing JS injection.
```

or without "preventing JS injection."

utils/tests/Base.t.sol
```diff
@@ -41,8 +43,10 @@ abstract contract Base_Test is BaseTest, Modifiers, StdAssertions, Utils {
     ComptrollerableMock internal comptrollerableMock;
     MerkleMock internal merkleMock;
     NoDelegateCallMock internal noDelegateCallMock;
+    Noop internal noop;
```
andreivladbrg: we can move it to `utils/src/tests/BaseTest.sol` and use it both in `lockup` and `here`

andreivladbrg (commented):
why did you change the imports `test/BaseTest.sol` to `[many=../../]/BaseTest.sol`
it easier to do the first approach as you don't have to count the number of `../`

utils/tests/Base.t.sol
andreivladbrg: f852d7311a469ae751b49d4950e4a7b9bd06de98

andreivladbrg: forgot about label ded2a2e60b5fa4ca5009081d79e7a5e0053acde0

utils/src/libraries/SafeTokenSymbol.sol
smol-ninja: Sounds good

utils/tests/Base.t.sol
smol-ninja: Yeah good catch

smol-ninja:
> why did you change the imports test/BaseTest.sol to [many=../../]/BaseTest.sol
it easier to do the first approach as you don't have to count the number of ../

At some places we are using relative imports for tests and at other places we are using direct import so to make it consistent, I changed it to using relative imports but feel free to revert it back but then let's do it for all the imports.

Both are easy because cursor automatically identifies the correct path so it was not a huge work to change that

andreivladbrg:
> At some places we are using relative imports for tests and at other places we are using direct import so to make it consistent, I changed it to using relative imports but feel free to revert it back but then let's do it for all the imports.

yeah, it was not consistent, but the readability was increased - nvm small thing

> Both are easy because cursor automatically identifies the correct path so it was not a huge work to change that

do you still use an IDE?

utils/src/libraries/SafeTokenSymbol.sol
andreivladbrg: afff2da79b868449707caf1fc9b191f0aec22b6a

andreivladbrg (approved):
👍

smol-ninja:
> do you still use an IDE?

Cursor all the time to review code and write on my own

PR #1428, opened 2026-03-03
Addresses audit findings 5, 6, 12, 14, 15 from Cantina

smol-ninja:
### Changelog

- Addresses [finding 5](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/5)
- Addresses [finding 6](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/6)
- Addresses [finding 12](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/12)
- Addresses [finding 14](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/14)
- Addresses [finding 15](https://cantina.xyz/code/fed466ff-5256-4c6a-8021-6da2c113beb2/findings/15)

I will create another PR for the remaining findings as there have been many changes in this PR already.

### DIFF
--- a/airdrops/src/SablierFactoryMerkleLL.sol
+++ b/airdrops/src/SablierFactoryMerkleLL.sol
@@ -74,7 +74,7 @@ contract SablierFactoryMerkleLL is ISablierFactoryMerkleLL, SablierFactoryMerkle
 
     /// @inheritdoc ISablierFactoryMerkleLL
     function createMerkleLL(
-        MerkleLL.ConstructorParams calldata campaignParams,
+        MerkleLL.ConstructorParams memory campaignParams,
         uint256 aggregateAmount,
         uint256 recipientCount
     )
@@ -84,6 +84,9 @@ contract SablierFactoryMerkleLL is ISablierFactoryMerkleLL, SablierFactoryMerkle
     {
         // Check: user-provided token is not the native token.
         _forbidNativeToken(address(campaignParams.token));
+
+        // Set the granularity to 1 second if it is provided as a zero.
+        campaignParams.granularity = campaignParams.granularity == 0 ? 1 seconds : campaignParams.granularity;
 
         // Hash the parameters to generate a salt.
         bytes32 salt = keccak256(abi.encodePacked(msg.sender, comptroller, abi.encode(campaignParams)));
--- a/airdrops/src/SablierMerkleVCA.sol
+++ b/airdrops/src/SablierMerkleVCA.sol
@@ -62,6 +62,9 @@ contract SablierMerkleVCA is
     /// @inheritdoc ISablierMerkleVCA
     uint128 public override totalForgoneAmount;
 
+    /// @inheritdoc ISablierMerkleVCA
+    uint128 public override totalRedistributionAmountPaid;
+
     /// @dev Tracks the full amount allocated to the recipients who claimed before the vesting end time.
     uint128 private _fullAmountAllocatedToEarlyClaimers;
 
@@ -171,6 +174,7 @@ contract SablierMerkleVCA is
         uint256 index,
         address to,
         uint128 fullAmount,
+        uint40 expireAt,
         bytes32[] calldata merkleProof,
         bytes calldata attestation
     )
@@ -181,7 +185,7 @@ contract SablierMerkleVCA is
         notZeroAddress(to)
     {
         // Check: the attestation signature is valid and the recovered signer matches the attestor.
-        _verifyAttestationSignature(msg.sender, attestation);
+        _verifyAttestationSignature(msg.sender, expireAt, attestation);
 
         // Check, Effect and Interaction: Pre-process the claim parameters on behalf of `msg.sender`.
         _preProcessClaim({ index: index, recipient: msg.sender, amount: fullAmount, merkleProof: merkleProof });
@@ -218,6 +222,11 @@ contract SablierMerkleVCA is
 
     /// @inheritdoc ISablierMerkleVCA
     function enableRedistribution() external override onlyAdmin {
+        // Check: `VESTING_END_TIME` is in the future.
+        if (VESTING_END_TIME <= block.timestamp) {
+            revert Errors.SablierMerkleVCA_VestingEndTimeNotInFuture(VESTING_END_TIME, block.timestamp);
+        }
+
         // Check: the redistribution is not already enabled.
         if (isRedistributionEnabled) {
             revert Errors.SablierMerkleVCA_RedistributionAlreadyEnabled();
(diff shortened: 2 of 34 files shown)


airdrops/src/SablierFactoryMerkleLL.sol
```diff
@@ -85,6 +85,9 @@ contract SablierFactoryMerkleLL is ISablierFactoryMerkleLL, SablierFactoryMerkle
         // Check: user-provided token is not the native token.
         _forbidNativeToken(address(campaignParams.token));
 
+        // Set the granularity to 1 second if it is provided as a zero.
+        campaignParams.granularity = campaignParams.granularity == 0 ? 1 seconds : campaignParams.granularity;
```
andreivladbrg, line 89: lets keep the struct as `calldata` and do this in the `SablierMerkleLL.constructor()`:

```solidity
VESTING_GRANULARITY = campaignParams.granularity == 0 ? 1 seconds: campaignParams.granularity; 
```

the benefit is that you can’t deploy campaigns (outside the factory) with `granularity == 0`

but the downside is that the granularity in the emitted event won’t match the one in storage- tho, this was already the case before adding this (factory emits 0, then the stream is created with 1)

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
```diff
@@ -99,17 +99,60 @@ contract ClaimTo_MerkleVCA_Integration_Test is
...
+        // `fullAmountAllocatedToRemainingClaimers` becomes VCA_CLAIM_AMOUNT. This sets the uncapped reward to
+        // (VCA_FULL_AMOUNT * (VCA_FULL_AMOUNT - VCA_CLAIM_AMOUNT)) / VCA_CLAIM_AMOUNT, which far exceeds the
+        // redistribution balance (VCA_CLAIM_AMOUNT).
+        MerkleVCA.ConstructorParams memory params = merkleVCAConstructorParams();
+        params.aggregateAmount = VCA_FULL_AMOUNT + VCA_CLAIM_AMOUNT;
+        merkleVCA = createMerkleVCA(params);
+        merkleBase = merkleVCA;
+
+        // Enable redistribution on the new campaign.
         merkleVCA.enableRedistribution();
```
andreivladbrg: we can create it directly enabled

airdrops/src/SablierMerkleVCA.sol
```diff
@@ -324,8 +333,19 @@ contract SablierMerkleVCA is
...
+                // If reward amount exceeds the remaining redistribution balance, cap it to the remaining balance. In a
+                // very unlikely scenario, the reward amount can exceed the remaining redistribution balance if
+                // `AGGREGATE_AMOUNT` is set lower than the actual total allocations in the Merkle tree.
+                if (rewardAmount > remainingRedistributionBalance) {
+                    rewardAmount = remainingRedistributionBalance;
+                }
+
                 // Update the transfer amount if there are rewards to distribute.
                 if (rewardAmount > 0) {
+                    totalRedistributionAmountPaid += rewardAmount;
```
andreivladbrg, line 348: would it be useful to include the `totalRedistributionAmountPaid` amount in `RedistributionReward` event ?

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
```diff
@@ -99,17 +99,60 @@ contract ClaimTo_MerkleVCA_Integration_Test is
...
-        setMsgSender(users.campaignCreator);
+        // Deploy a new campaign with undervalued aggregate amount. We choose the value such that after one early claim
+        // (which sets `_fullAmountAllocatedToEarlyClaimers` = VCA_FULL_AMOUNT),
+        // `fullAmountAllocatedToRemainingClaimers` becomes VCA_CLAIM_AMOUNT. This sets the uncapped reward to
+        // (VCA_FULL_AMOUNT * (VCA_FULL_AMOUNT - VCA_CLAIM_AMOUNT)) / VCA_CLAIM_AMOUNT, which far exceeds the
+        // redistribution balance (VCA_CLAIM_AMOUNT).
+        MerkleVCA.ConstructorParams memory params = merkleVCAConstructorParams();
+        params.aggregateAmount = VCA_FULL_AMOUNT + VCA_CLAIM_AMOUNT;
+        merkleVCA = createMerkleVCA(params);
+        merkleBase = merkleVCA;
```
andreivladbrg: we already doing this in `_test_ClaimTo`

andreivladbrg (commented):
great work, left few comments below

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
andreivladbrg: 9c2fc29d697c7626bd5b1d2f0bb3f5a68a8cf66a

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
andreivladbrg: 9c2fc29d697c7626bd5b1d2f0bb3f5a68a8cf66a

airdrops/src/SablierFactoryMerkleLL.sol
smol-ninja: The reason I moved it into create function is that it creates inconsistency between what is stored and what is emitted in the event.

> the benefit is that you can’t deploy campaigns (outside the factory) with granularity == 0

I wouldn't worry about this because it is a deployed outside the create flow. I am worried about the above point where event emits `0` and value stored is `1`. Any indexer relying on these events will store incorrect value for  `granularity`.

So, I would prioritise the consistency between event and value stored over the other.

> this was already the case before adding this

Yes but previously the value emitted is also the value stored in the campaign contract.

airdrops/tests/integration/concrete/campaign/vca/claim-to/claimTo.t.sol
smol-ninja: Oh right.

airdrops/src/SablierMerkleVCA.sol
smol-ninja: I dont think so. We are already emitting `rewardAmount` and indexer can some `rewardAmount` to get `totalRedistributionAmountPaid` without making any RPC calls.

airdrops/src/SablierFactoryMerkleLL.sol
andreivladbrg: wdys about having them in both?

smol-ninja: Why do you want both?

In my opinion having it in create function is sufficient because that's the actual flow. The direct deployment is one in a million case and even in that case the stream created will handle the granularity correctly.

We can have both, but I think it's redundant.

If you have a very strong opinion, okay let's have both.

airdrops/src/SablierMerkleVCA.sol
andreivladbrg: fair enough

airdrops/src/SablierFactoryMerkleLL.sol
andreivladbrg: > If you have a very strong opinion, okay let's have both.

i don’t - let’s keep it

> Why do you want both?

my motivation was that you could have two campaigns that, in practice, have the same config but different values in state - but that’s not a strong point

smol-ninja: Sounds good.

smol-ninja:
@andreivladbrg can you plz approve the PR?

andreivladbrg (approved):

PR #1435, opened 2026-03-05
ai: no redundant modifier in btt skill

andreivladbrg:
@smol-ninja As we also discussed on Slack, we both faced this issue multiple times.

### DIFF
--- a/.claude/skills/btt-testing/SKILL.md
+++ b/.claude/skills/btt-testing/SKILL.md
@@ -190,7 +190,24 @@ For success cases, enumerate all side effects:
 
 Put events in parenthesis: {EventName}.
 
-### 5. Use proper indentation
+### 5. Don't Add a Modifier with the Same Name as the Function
+
+A test function should NEVER have a modifier that matches its own name. The function name already encodes the condition —
+adding a same-name modifier is redundant.
+
+```solidity
+// WRONG: modifier matches the function name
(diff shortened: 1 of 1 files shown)


.claude/skills/btt-testing/SKILL.md
```diff
@@ -190,7 +190,27 @@ For success cases, enumerate all side effects:
 
 Put events in parenthesis: {EventName}.
 
-### 5. Use proper indentation
+### 5. Don't Add a Redundant Modifier for the Last Branch
+
+The last (deepest) test function in a BTT path should NOT have a modifier matching its own name. The function name
```
smol-ninja: We should add it as a general rule and not just for the last branch. I have seen it adds the modifier with the same number for other tests too.

smol-ninja (changes requested):
Lets also update it in sablier skills repo after this is merged.

.claude/skills/btt-testing/SKILL.md
andreivladbrg: good idea

> Lets also update it in sablier skills repo after this is merged.

ok, will do

andreivladbrg: forced pushed ca61f230599485015a0058cddfa9ddabb5ba9bc4

smol-ninja (approved):

Issue #1417, opened 2026-02-19
Bob and Escrow Checklist


smol-ninja:
Copied from https://github.com/sablier-labs/lockup/pull/1404

### Escrow
- [x] Integration tests
- [x] Fuzz tests
- [x] Fork tests
- [x] Invariant tests
- [x] Deployment scripts
- [x] Docs

### Bob
- [x] Integration tests
- [x] Fuzz tests
- [x] Fork tests
- [x] Invariant tests
- [x] Deployment scripts
- [x] Docs

PfanP:
I am very interested in handling this task, if you are open to assigning it to me

andreivladbrg:
@PfanP thanks for offering, currently @smol-ninja and i are actually working on this issue

smol-ninja:
All done.

PR #1436, opened 2026-03-05
feat(adapter): integrate Lido native withdrawal as fallback unstaking path for stETH token

smol-ninja:
### Changelog

- Addresses [Finding 25](https://github.com/Cyfrin/audit-2026-02-sablier-bob-2/issues/25)
- Adds Lido native withdrawal queue as a fallback that can only be triggered by Comptroller regardless of vault status
- Adds minimal fork tests
- Add integration tests

### Resources

- https://vscode.blockscan.com/ethereum/0xe42c659dc09109566720ea8b2de186c2be7d94d9
- https://docs.lido.fi/contracts/withdrawal-queue-erc721
- [Fluid Contract](https://github.com/Instadapp/fluid-contracts-public/blob/a9949b48ba1247d4f478cd0acb40896b5c8bf3f8/contracts/protocols/steth/main.sol#L277-L283)


bob/src/interfaces/external/ILidoWithdrawalQueue.sol
```diff
@@ -0,0 +1,56 @@
...
+        uint256[] calldata _requestIds,
+        uint256 _firstIndex,
+        uint256 _lastIndex
+    )
+        external
+        view
+        returns (uint256[] memory hintIds);
+
+    /// @notice length of the checkpoint array. Last possible value for the hint.
+    function getLastCheckpointIndex() external view returns (uint256);
```
andreivladbrg, line 39: the functions are not divided by READ-ONLY / STATE-CHANGING headers

smol-ninja:
I have added integration tests (reviewed too). I have added fork tests (reviewed briefly but it covers an end to end interaction with both Lido and Curve on mainnet).

We can refactor fork tests later as a part of https://github.com/sablier-labs/lockup/issues/1417.

smol-ninja:
Question to @sablier-labs/evm, should we allow lido native withdrawal only if vault has settled or expired? Currently, comptroller can request lido withdrawal anytime. 

If we go with the new approach, I will add `requestLidoWithdrawal` to `Bob` as well which will make call to `requestLidoWithdrawal` on adapter contract.

andreivladbrg:
> Question to @sablier-labs/evm, should we allow lido native withdrawal only if vault has settled or expired? Currently, comptroller can request lido withdrawal anytime.

i don't think that's a good idea

the motivation of adding this lido withdrawal queue is to prevent cases like when curve gets outdated and (i) the expiry date is far in the future (ii) the target price is high, and in the meantime the pool gets out of liquidity

what's the worst that can happen - 2 weeks of waiting to receive the funds, right?

smol-ninja:
I just realized my question was stupid. Because `unstakeFullAmount` can only be called if a vault is settled or expired, be able to call `requestLidoWithdrawal` anytime is good for users. We can call `requestLidoWithdrawal` in advance so that once vault is expired/settled, user can immediately claim their funds.

bob/src/SablierLidoAdapter.sol
```diff
@@ -227,6 +243,69 @@ contract SablierLidoAdapter is
...
+            amounts = new uint256[](totalRequests);
+
+            // Assign amounts for each request except the last one.
+            uint256 lastIndex = totalRequests - 1;
+            for (uint256 i; i < lastIndex; ++i) {
+                amounts[i] = maxAmountPerRequest;
+            }
+
+            // Assign the remaining amount to the last request.
+            amounts[lastIndex] = stETHAmount - maxAmountPerRequest * lastIndex;
```
andreivladbrg: the last request's remainder could be below lido's `MIN_STETH_WITHDRAWAL_AMOUNT` (100 wei), causing `requestWithdrawals` to revert

please have a look at how [fluid integrated lido](https://github.com/Instadapp/fluid-contracts-public/blob/a9949b48ba1247d4f478cd0acb40896b5c8bf3f8/contracts/protocols/steth/main.sol#L277-L283)

andreivladbrg (commented):
@smol-ninja had a quick look, more review comments upcoming

but first i have a design question: shouldn't we revert in the `enter` function if the comptroller requested a lido withdrawal? because in the current `requestLidoWithdrawal` function, there are no state updates, can't this lead to messed up values in `redeem` & `processTokenRedemption`?

smol-ninja:
> but first i have a design question: shouldn't we revert in the enter function if the comptroller requested a lido withdrawal

Oh thats a very good point.

We can revert but wouldn't it be better to allow `requestLidoWithdrawal` only if vault has settled or expired (my original question was not stupid I guess xD)? In that case, `enter` would revert automatically and there would be no need to add a new storage variables. If we want to revert in `enter` and allow `requestLidoWithdrawal` to be called anytime then we will have to introduce a new storage variable.

bob/src/interfaces/external/ILidoWithdrawalQueue.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1436/changes/5d92d4b17982afc50b5d977ec71d0c084823e2dc

bob/src/SablierLidoAdapter.sol
smol-ninja: good catch

andreivladbrg:
> vault has settled or expired

yeah, that would fix it as well - tho, if someone redeems and we didn't call request withdrawal on lido, it can either use curve to swap. are we ok with that? or revert due to not enough liquidity

can anything else happen?

> my original question was not stupid I guess xD)

😅

bob/src/SablierLidoAdapter.sol
smol-ninja: Fixed it in https://github.com/sablier-labs/lockup/pull/1436/changes/285b8c9ad5d8b90d029b0835e3f6c1829b866c6c. It was a really good catch.

smol-ninja:
> if someone redeems and we didn't call request withdrawal on lido, it can either use curve to swap. are we ok with that?

First of all, this is a very unlikely scenario that we would use Lido native queue. And even if we do, if someone redeems, there are two scenarios:

1. It goes through Curve if Curve has liquidity (in that case we will anyways not use Lido native queue)
2. If reverts due to lack of liquidity on Curve (since it unstakes the entire amount) and in that case, we will trigger Lido and notify users via UI that they need to wait for few weeks and follow updates on TG/Discord/X. 

So, its not a problem at all.

bob/src/SablierLidoAdapter.sol
```diff
@@ -227,6 +245,95 @@ contract SablierLidoAdapter is
         _vaultYieldFee[vaultId] = feeOnYield;
     }
 
+    /// @inheritdoc ISablierLidoAdapter
+    function requestLidoWithdrawal(uint256 vaultId) external override onlyComptroller {
+        // Check: the vault status is not ACTIVE.
+        if (ISablierBobState(SABLIER_BOB).statusOf(vaultId) == Bob.Status.ACTIVE) {
```
smol-ninja, line 258: I intentionally didn't add calling `syncPriceFromOracle` function here if status is `ACTIVE` because its just an admin function. When and if we ever need it, we can just batch both `syncPriceFromOracle` and `requestLidoWithdrawal` in a single SAFE tx.

bob/src/interfaces/ISablierBob.sol
```diff
@@ -161,7 +161,8 @@ interface ISablierBob is IBatch, IComptrollerable, ISablierBobState {
     /// the adapter.
     ///
     /// Requirements:
-    /// - The vault must have SETTLED status.
+    /// - Either block timestamp must be greater than or equal to the vault expiry or the latest price from the oracle
```
smol-ninja, line 166: I rewrote it like this because the function makes call to `syncPriceFromOracle` when status is active. So even though the current status is `ACTIVE`, the function can work if sync changes the status to `SETTLED`.

bob/src/SablierLidoAdapter.sol
andreivladbrg: ok, sounds good

bob/src/SablierLidoAdapter.sol
```diff
@@ -227,6 +245,95 @@ contract SablierLidoAdapter is
...
+            revert Errors.SablierLidoAdapter_VaultActive(vaultId);
```
andreivladbrg, line 259: besides this check, wdyt about checking `isStakedInAdapter == true`?

smol-ninja: the function would revert with `SablierLidoAdapter_NoWstETHToWithdraw` if the vault is not staked in adapter. So I dont think we should check `isStakedInAdapter = true`

smol-ninja:
I have updated this PR too @andreivladbrg

bob/src/SablierLidoAdapter.sol
```diff
@@ -344,6 +457,34 @@ contract SablierLidoAdapter is
...
+        // Interaction: search the request IDs in all the finalized batches. If any request ID is not finalized, the
+        // corresponding hint will be zero.
+        uint256[] memory hints =
+            ILidoWithdrawalQueue(LIDO_WITHDRAWAL_QUEUE).findCheckpointHints(requestIds, 1, lastIndex);
+
+        // Get the ETH balance before claiming.
+        uint256 ethBefore = address(this).balance;
+
+        // Interaction: Claim all request IDs. It will revert if any request ID is not finalized.
+        ILidoWithdrawalQueue(LIDO_WITHDRAWAL_QUEUE).claimWithdrawals(requestIds, hints);
```
andreivladbrg, line 489: should we also clear `_lidoWithdrawalRequestIds` in this function?

bob/src/SablierLidoAdapter.sol
```diff
@@ -227,6 +245,95 @@ contract SablierLidoAdapter is
...
+        }
+        // Otherwise, its just one request.
+        else {
+            // Initialize array length to 1.
+            amounts = new uint256[](1);
+            amounts[0] = stETHAmount;
+        }
+
+        // Interaction: Approve Lido withdrawal queue to spend the exact stETH amount.
+        IStETH(STETH).approve(LIDO_WITHDRAWAL_QUEUE, stETHAmount);
```
andreivladbrg, line 331: we already have these in the constructor:

```solidity
// Approve wstETH contract to spend stETH, required for wrapping.
IStETH(STETH).approve(WSTETH, type(uint128).max);

// Approve Curve pool to spend stETH, required for unwrapping.
IStETH(STETH).approve(CURVE_POOL, type(uint128).max);
```

should we add it there or can it be a problem security wise?

bob/src/SablierLidoAdapter.sol
```diff
@@ -344,6 +457,34 @@ contract SablierLidoAdapter is
...
+        // Interaction: Claim all request IDs. It will revert if any request ID is not finalized.
+        ILidoWithdrawalQueue(LIDO_WITHDRAWAL_QUEUE).claimWithdrawals(requestIds, hints);
+
+        // Calculate the ETH received from claiming.
+        uint256 ethReceived = address(this).balance - ethBefore;
+
+        // Interaction: Wrap ETH to get WETH.
+        IWETH9(WETH).deposit{ value: ethReceived }();
+
+        return ethReceived.toUint128();
```
andreivladbrg: should we change this as i did in my PR?

```solidity
uint256 ethReceived = address(this).balance - ethBefore;
uint128 ethReceivedU128 = ethReceived.toUint128();
IWETH9(WETH).deposit{ value: ethReceived }();
return ethReceivedU128;
```

first checks (overflow check in `.toUint128()`), then interaction/effects

bob/src/interfaces/external/ILidoWithdrawalQueue.sol
```diff
@@ -0,0 +1,67 @@
...
+
+    /*//////////////////////////////////////////////////////////////////////////
+                              STATE-CHANGING FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Claim a batch of withdrawal requests if they are finalized sending locked ether to the owner.
```
andreivladbrg:
```suggestion
    /// @notice Claim a batch of withdrawal requests if they are finalized sending locked ETH to the owner.
```
?

bob/src/interfaces/external/ILidoWithdrawalQueue.sol
```diff
@@ -0,0 +1,67 @@
...
+    ///  - Any request is not finalized or already claimed.
+    ///  - `msg.sender` is not an owner of the requests.
+    function claimWithdrawals(uint256[] calldata _requestIds, uint256[] calldata _hints) external;
+
+    /// @notice Request the batch of stETH for withdrawal. Approvals for the passed amounts should be done before.
+    /// @param _amounts an array of stETH amount values. The standalone withdrawal request will be created for each item
+    /// in the passed list.
+    /// @param _owner address that will be able to manage the created requests. If `address(0)` is passed, `msg.sender`
+    /// will be used as owner.
+    /// @return requestIds an array of the created withdrawal request ids.
```
andreivladbrg:
```suggestion
    /// @return requestIds an array of the created withdrawal request IDs.
```

bob/src/SablierLidoAdapter.sol
```diff
@@ -344,6 +457,34 @@ contract SablierLidoAdapter is
...
+
+        // Interaction: Since Lido processes withdrawals in batches, we need to find the number of total finalized
+        // batches occurred so far. This is used in the next step.
+        uint256 lastIndex = ILidoWithdrawalQueue(LIDO_WITHDRAWAL_QUEUE).getLastCheckpointIndex();
+
+        // Interaction: search the request IDs in all the finalized batches. If any request ID is not finalized, the
+        // corresponding hint will be zero.
+        uint256[] memory hints =
+            ILidoWithdrawalQueue(LIDO_WITHDRAWAL_QUEUE).findCheckpointHints(requestIds, 1, lastIndex);
+
```
andreivladbrg, line 484: since this can be called by users and the request function can only be called by us (comptroller)

wouldn't it be a good idea to save a `_lidoCheckpointAtRequest` instead of iterating from 1 here?

andreivladbrg (commented):
left few more comments:

let's add in the `redeem` natspec that if the lido withdrawal queue was called, it can revert if not enough time has passed

bob/src/SablierLidoAdapter.sol
smol-ninja: We should not add it to the constructor since its a function that will be called may be once or twice in 5 years period. And therefore, approving it only when its required would be a better approach than max approving it in the constructor.

bob/src/SablierLidoAdapter.sol
smol-ninja: From the safety point of view, I would keep it like this. `findCheckpointHints` uses binary search so its very cheap to find an element here. And So far only 1000 checkpoints have been reached.

bob/src/SablierLidoAdapter.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1436/changes/c4aa2e1ec7a71d53518945c28e1533f0b98354b3

bob/src/SablierLidoAdapter.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1436/changes/c4aa2e1ec7a71d53518945c28e1533f0b98354b3

bob/src/interfaces/external/ILidoWithdrawalQueue.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1436/changes/c4aa2e1ec7a71d53518945c28e1533f0b98354b3

bob/src/interfaces/external/ILidoWithdrawalQueue.sol
smol-ninja: https://github.com/sablier-labs/lockup/pull/1436/changes/c4aa2e1ec7a71d53518945c28e1533f0b98354b3

bob/src/SablierLidoAdapter.sol
andreivladbrg: good point, lets keep it in the function

bob/src/SablierLidoAdapter.sol
andreivladbrg: alright, lets keep current version

bob/src/SablierLidoAdapter.sol
andreivladbrg: actually, does it introduce a bug? xD sorry for suggesting it

we have this check in the request function:

```solidity
// Check: Lido withdrawal has not already been requested for this vault.
if (_lidoWithdrawalRequestIds[vaultId].length > 0) {
    revert Errors.SablierLidoAdapter_LidoWithdrawalAlreadyRequested(vaultId);
}
```

so, it would be possible to request again for the same vault ID

smol-ninja:
@andreivladbrg ready for you. We will "Squash and Merge" with this one.

bob/src/SablierLidoAdapter.sol
smol-ninja: Yeah. Thats why we need invariant tests asap. Reverting it now.

bob/src/SablierLidoAdapter.sol
```diff
@@ -356,8 +469,40 @@ contract SablierLidoAdapter is
                           PRIVATE STATE-CHANGING FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
 
-    /// @dev Converts wstETH to WETH using Curve exchange, with oracle-based slippage protection.
-    function _wstETHToWeth(uint128 wstETHAmount) private returns (uint128 wethReceived) {
+    /// @dev Claims finalized Lido withdrawals for a vault and wraps the received ETH into WETH.
+    function _claimLidoWithdrawals(uint256 vaultId) private returns (uint128 ethReceivedU128) {
```
andreivladbrg: let's call it `wethReceived` as in `_swapWstETHToWeth`

the return value

bob/src/SablierLidoAdapter.sol
smol-ninja: https://github.com/sablier-labs/lockup/compare/772d7cc73fd77ea032cd771bc035969d62ee1a51..44f49040fe0dc09b8d5f9ac4235e216e5d530282

bob/src/SablierLidoAdapter.sol
smol-ninja: What do you mean? This is what we have rn:

```solidity
function _swapWstETHToWeth(uint128 wstETHAmount) private returns (uint128 wethReceived)
```

andreivladbrg: yes, thats the `_swapWstETHToWeth` function

i am talking about 

```solidity
function _claimLidoWithdrawals(uint256 vaultId) private returns (uint128 ethReceivedU128)
```

to:

```solidity
function _claimLidoWithdrawals(uint256 vaultId) private returns (uint128 wethReceived)
```

smol-ninja: https://github.com/sablier-labs/lockup/pull/1436/changes/6c4d57ba3206f77b275e738332bb5783a16cab62

bob/src/interfaces/ISablierBob.sol
```diff
@@ -159,9 +159,12 @@ interface ISablierBob is IComptrollerable, ISablierBobState {
     /// - If no adapter is configured for the vault, a fee in the native token is applied.
     /// - If an adapter is configured for the vault, a fee, in the deposit token, is deducted from yield generated by
     /// the adapter.
+    /// - If unstake via Lido withdrawal queue contract is triggered, redeem will revert until the withdrawal from the
+    /// Lido queue is finalized.
```
andreivladbrg, line 163: would it be useful to mention that the default is the curve swap?

bob/src/SablierLidoAdapter.sol
andreivladbrg: > the function would revert with SablierLidoAdapter_NoWstETHToWithdraw if the vault is not staked in adapter

why would it revert? we don't clear `_vaultTotalWstETH[vaultId]`mapping after unstaking

bob/src/interfaces/ISablierBob.sol
smol-ninja: I don't think so because default is subjective. Curve works as long as lido withdraw is not triggered which can be triggered at any point in time.

andreivladbrg: alright

bob/src/SablierLidoAdapter.sol
smol-ninja: It would revert with `SablierLidoAdapter_LidoWithdrawalAlreadyRequested` if comptroller tries to call `requestLidoWithdrawal` again.

And if a vault is not staked in adapter, then it would revert with `SablierLidoAdapter_NoWstETHToWithdraw` because `_vaultTotalWstETH[vaultId]` is zero.

andreivladbrg: you are right, we don't need this check

andreivladbrg (approved):

PR #1438, opened 2026-03-06
test(bob): add invariant tests for escrow

smol-ninja:
Partially addresses https://github.com/sablier-labs/lockup/issues/1417.

I came up with a few invariants for Escrow and decided to add them to tests.

### DIFF
--- a/bob/tests/escrow/invariant/Invariant.t.sol
+++ b/bob/tests/escrow/invariant/Invariant.t.sol
@@ -7,7 +7,7 @@ import { Escrow } from "src/types/Escrow.sol";
 
 import { Base_Test } from "../Base.t.sol";
 import { EscrowHandler } from "./handlers/EscrowHandler.sol";
-import { EscrowStore } from "./stores/EscrowStore.sol";
+import { Store } from "./stores/Store.sol";
 
 /// @notice Invariant tests for {SablierEscrow}.
 contract Invariant_Test is Base_Test, StdInvariant {
@@ -16,7 +16,7 @@ contract Invariant_Test is Base_Test, StdInvariant {
     //////////////////////////////////////////////////////////////////////////*/
 
     EscrowHandler internal handler;
-    EscrowStore internal store;
+    Store internal store;
 
     /*//////////////////////////////////////////////////////////////////////////
                                   SET-UP FUNCTION
@@ -26,11 +26,11 @@ contract Invariant_Test is Base_Test, StdInvariant {
         Base_Test.setUp();
 
         // Deploy the contracts.
-        store = new EscrowStore();
+        store = new Store();
         handler = new EscrowHandler({ store_: store, escrow_: escrow });
 
         // Label the contracts.
-        vm.label({ account: address(store), newLabel: "EscrowStore" });
+        vm.label({ account: address(store), newLabel: "Store" });
         vm.label({ account: address(handler), newLabel: "EscrowHandler" });
 
         // Target the handler for invariant testing.
@@ -51,13 +51,13 @@ contract Invariant_Test is Base_Test, StdInvariant {
     /// - Sell side: amount transferred to buyer + amount transferred to comptroller = sell amount.
     /// - Buy side: amount transferred to seller + amount transferred to comptroller = buy amount.
     function invariant_AmountConservationOnFill() external view {
-        for (uint256 i = 0; i < store.totalOrders(); ++i) {
+        for (uint256 i = 0; i < store.orderCount(); ++i) {
             uint256 orderId = store.orderIds(i);
 
             // Skip if order is not filled.
             if (!escrow.wasFilled(orderId)) continue;
 
-            EscrowStore.FillData memory data = store.getFillData(orderId);
+            Store.FillData memory data = store.getFillData(orderId);
             assertEq(
                 data.amountToTransferToBuyer + data.feeDeductedFromBuyerAmount,
                 escrow.getSellAmount(orderId),
@@ -75,11 +75,11 @@ contract Invariant_Test is Base_Test, StdInvariant {
     /// @dev For a given sell token, the escrow's balance equals the sum of sell amounts across all open and expired
     /// orders.
     function invariant_ContractBalancePerSellToken() external view {
-        for (uint256 t = 0; t < store.totalTokens(); ++t) {
+        for (uint256 t = 0; t < store.tokensCount(); ++t) {
             IERC20 token = store.tokens(t);
             uint256 expectedBalance = 0;
 
-            for (uint256 i = 0; i < store.totalOrders(); ++i) {
+            for (uint256 i = 0; i < store.orderCount(); ++i) {
                 uint256 orderId = store.orderIds(i);
                 bool isOpenOrExpired =
                     escrow.statusOf(orderId) == Escrow.Status.OPEN || escrow.statusOf(orderId) == Escrow.Status.EXPIRED;
--- a/bob/tests/escrow/invariant/Invariant.t.sol
+++ b/bob/tests/escrow/invariant/Invariant.t.sol
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.8.22 <0.9.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { StdInvariant } from "forge-std/src/StdInvariant.sol";
+import { Escrow } from "src/types/Escrow.sol";
+
+import { Base_Test } from "../Base.t.sol";
+import { EscrowHandler } from "./handlers/EscrowHandler.sol";
+import { EscrowStore } from "./stores/EscrowStore.sol";
+
+/// @notice Invariant tests for {SablierEscrow}.
+contract Invariant_Test is Base_Test, StdInvariant {
+    /*//////////////////////////////////////////////////////////////////////////
+                                   TEST CONTRACTS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    EscrowHandler internal handler;
+    EscrowStore internal store;
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                  SET-UP FUNCTION
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function setUp() public virtual override {
+        Base_Test.setUp();
+
+        // Deploy the contracts.
+        store = new EscrowStore();
+        handler = new EscrowHandler({ store_: store, escrow_: escrow });
+
+        // Label the contracts.
+        vm.label({ account: address(store), newLabel: "EscrowStore" });
+        vm.label({ account: address(handler), newLabel: "EscrowHandler" });
+
+        // Target the handler for invariant testing.
+        targetContract(address(handler));
+
+        // Prevent these addresses from being fuzzed as `msg.sender`.
+        excludeSender(address(escrow));
+        excludeSender(address(comptroller));
+        excludeSender(address(handler));
+        excludeSender(address(store));
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                     INVARIANTS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev For a given filled order:
+    /// - Sell side: amount transferred to buyer + amount transferred to comptroller = sell amount.
+    /// - Buy side: amount transferred to seller + amount transferred to comptroller = buy amount.
+    function invariant_AmountConservationOnFill() external view {
+        for (uint256 i = 0; i < store.totalOrders(); ++i) {
+            uint256 orderId = store.orderIds(i);
+
+            // Skip if order is not filled.
+            if (!escrow.wasFilled(orderId)) continue;
+
+            EscrowStore.FillData memory data = store.getFillData(orderId);
+            assertEq(
+                data.amountToTransferToBuyer + data.feeDeductedFromBuyerAmount,
+                escrow.getSellAmount(orderId),
+                "Invariant violation: amount transferred to buyer + amount transferred to comptroller != sell amount"
+            );
+
+            assertEq(
+                data.amountToTransferToSeller + data.feeDeductedFromSellerAmount,
+                data.buyAmount,
+                "Invariant violation: amount transferred to seller + amount transferred to comptroller != buy amount"
+            );
+        }
+    }
+
+    /// @dev For a given sell token, the escrow's balance equals the sum of sell amounts across all open and expired
+    /// orders.
+    function invariant_ContractBalancePerSellToken() external view {
+        for (uint256 t = 0; t < store.totalTokens(); ++t) {
+            IERC20 token = store.tokens(t);
+            uint256 expectedBalance = 0;
+
+            for (uint256 i = 0; i < store.totalOrders(); ++i) {
+                uint256 orderId = store.orderIds(i);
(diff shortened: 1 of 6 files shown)


smol-ninja:
@andreivladbrg shall we merge this too?

andreivladbrg:
> @andreivladbrg shall we merge this too?

haven't reviewed it yet - will try to do it later today

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+contract EscrowHandler is Constants, StdCheats, BaseUtils, PRBMathUtils {
+    /*//////////////////////////////////////////////////////////////////////////
+                                     VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Maximum number of orders that can be created during invariant runs.
+    uint256 internal constant MAX_ORDER_COUNT = 1000;
+
+    /// @dev Maximum number of tokens that can be created during invariant runs.
+    uint256 internal constant MAX_TOKEN_COUNT = 20;
```
andreivladbrg: not used

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+        uint40 expiryTime
+    )
+        external
+        instrument("createOrder")
+    {
+        // Limit the number of orders.
+        if (store.totalOrders() >= MAX_ORDER_COUNT) return;
+
+        // Exclude sender from being fuzzed as certain addresses.
+        if (sender == address(0) || sender == address(escrow)) return;
```
andreivladbrg: why not `assume`?

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+        // If no tokens exist or random number is even, create a new one.
+        if (store.totalTokens() == 0 || vm.randomUint() % 2 == 0) {
+            // Generate a random decimals value.
+            uint8 decimals = uint8(vm.randomUint(2, 20));
+
+            // Use next token index as the ID.
+            string memory id = vm.toString(store.totalTokens());
+
+            // Set the caller to the handler's address to deploy a new token.
+            setMsgSender(address(this));
```
andreivladbrg, line 244: we don't need to change the caller here

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+            ERC20Mock token = new ERC20Mock(string.concat("Token", id), string.concat("TKN", id), decimals);
+
+            // Update it in the store.
+            store.pushToken(IERC20(address(token)));
```
andreivladbrg: uneeded casting here, and below when returned

`ERC20Mock` inherits `IERC20`

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+        sellAmount = boundUint128(sellAmount, 1, 1000 ether);
+        minBuyAmount = boundUint128(minBuyAmount, 1, 1000 ether);
+
+        // If expiry time is not zero, bound it to reasonable range.
+        if (expiryTime > 0) {
+            expiryTime = boundUint40(expiryTime, getBlockTimestamp() + 10 days, getBlockTimestamp() + 100 days);
+        }
+
+        // If buyer is not zero, exclude it from being fuzzed as certain addresses.
+        if (buyer != address(0) && buyer == address(escrow)) return;
```
andreivladbrg: don't really get the comment  - why not only check for `buyer == escrow`? we don't need the zero check

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+        store = store_;
+        escrow = escrow_;
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                 HANDLER FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function createOrder(
+        address sender,
```
andreivladbrg: lets call it `seller` instead

bob/tests/escrow/invariant/stores/Store.sol
andreivladbrg, line 1: in the other protocols this is called just `Store`

bob/tests/escrow/invariant/stores/EscrowStore.sol
```diff
@@ -0,0 +1,64 @@
...
+    function pushToken(IERC20 token) external {
+        tokens.push(token);
+    }
+
+    /// @dev Records the fill data from a successful `fillOrder` call.
+    function recordFill(uint256 orderId, FillData calldata data) external {
+        _fillData[orderId] = data;
+    }
+
+    function totalOrders() external view returns (uint256) {
```
andreivladbrg: lets call it `ordersCount`

bob/tests/escrow/invariant/stores/EscrowStore.sol
```diff
@@ -0,0 +1,64 @@
...
+        return orderIds.length;
+    }
+
+    function totalTokens() external view returns (uint256) {
```
andreivladbrg: similar, lets call it `tokensCount`

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+import { EscrowStore } from "../stores/EscrowStore.sol";
+
+/// @notice Handler for the invariant tests of {SablierEscrow} contract.
+contract EscrowHandler is Constants, StdCheats, BaseUtils, PRBMathUtils {
+    /*//////////////////////////////////////////////////////////////////////////
+                                     VARIABLES
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @dev Maximum number of orders that can be created during invariant runs.
+    uint256 internal constant MAX_ORDER_COUNT = 1000;
```
andreivladbrg: thats quite high number - IMO we either lower it or increase the invariant runs

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+        uint128 buyAmount,
+        address buyer
+    )
+        external
+        instrument("fillOrder")
+        adjustTimestamp(timeJumpSeed)
+        orderCountNotZero
+    {
+        // Pick a random order from the store.
+        uint256 orderId = store.orderIds(bound(orderIndexSeed, 0, store.totalOrders() - 1));
```
andreivladbrg: lets split this into 2 lines for increased readability

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,261 @@
...
+            minBuyAmount: minBuyAmount,
+            buyer: buyer,
+            expiryTime: expiryTime
+        });
+
+        // Record in store.
+        store.pushOrderId(orderId);
+    }
+
+    function cancelOrder(
```
andreivladbrg: should we limit the number of canceled orders? so that we prioritize `filledOrders`?

andreivladbrg (changes requested):
sorry for longer wait for the review

left my comments below which are mostly addressed in this commit be2a674cb0bf81a7cebefdc588054727553698ca

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
```diff
@@ -0,0 +1,259 @@
...
+        escrow = escrow_;
+    }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                 HANDLER FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function cancelOrder(uint256 timeJumpSeed)
+        external
+        instrument("cancelOrder")
```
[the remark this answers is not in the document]
andreivladbrg: its ok

bob/tests/escrow/invariant/stores/Store.sol
```diff
@@ -0,0 +1,64 @@
...
+    /*//////////////////////////////////////////////////////////////////////////
+                                     FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function getFillData(uint256 orderId) external view returns (FillData memory) {
+        return _fillData[orderId];
+    }
+
+    function orderCount() external view returns (uint256) {
+        return orderIds.length;
```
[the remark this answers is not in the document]
andreivladbrg: these are tests, its ok

bob/tests/escrow/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,98 @@
...
+    //////////////////////////////////////////////////////////////////////////*/
+
+    EscrowHandler internal handler;
+    Store internal store;
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                  SET-UP FUNCTION
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function setUp() public virtual override {
```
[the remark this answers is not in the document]
andreivladbrg: these are tests, its ok

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: I guess I forgot about using it. I had planned to limit number of tokens but I think we dont need it since `_getOrCreateToken` handles it very well.

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: If `vm.assume` fails, it rejects the entire run (the sequence of calls) but if we `return`, then it continues with the remaining calls. Lets say an invariant run has the following sequence of calls:

1. create
2. fill
3. cancel
4. fill

Now if the fuzed input of `create` is `address(0)`, `assume` would discard the run without continuing with `full`, `cancel`, and `fill` but `return` would continue the remaining calls and would only ignore `create`. This approach is therefore better than using `assume` in invariants.

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: Yeah you are right.

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: Will lower it to 20

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: Thats a good question. IMO invariant runs fairly covers a good number of calls sequence. So it does not matter  if we limit it or not.

This is a typical metrics with 20 runs with 20 depth.
Screenshot 2026-03-26 at 13 46 59

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: We do need it so that the token deployed can use this address + nonce to determine the token address. Otherwise, the caller would keep changing and it will lead to a collision (the reason invariant failed in your commit)

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1438/changes/387f5887007b3cf9b1c8694c6477f0fee9ea5f7d

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1438/changes/387f5887007b3cf9b1c8694c6477f0fee9ea5f7d

bob/tests/escrow/invariant/Invariant.t.sol
```diff
@@ -0,0 +1,98 @@
...
+                if (escrow.getSellToken(orderId) == token && isOpenOrExpired) {
+                    expectedBalance += escrow.getSellAmount(orderId);
+                }
+            }
+
+            assertEq(
+                token.balanceOf(address(escrow)),
+                expectedBalance,
+                "Invariant violation: contract balance != sum of sell amounts of all open orders"
+            );
```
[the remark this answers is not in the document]
smol-ninja: good catch 

https://github.com/sablier-labs/evm-monorepo/pull/1438/changes/387f5887007b3cf9b1c8694c6477f0fee9ea5f7d

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
andreivladbrg: ohh, ok 

lets add an explanatory comment as it is not straight forward why we need it

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
andreivladbrg: good points, lets keep `return`

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
andreivladbrg: don't you think invariants should "mimic" real scenarios?
IMO in reality we will see more filled orders than cancellations

smol-ninja: Invariants are supposed to mimic all scenarios especially the edge case scenarios because we want to find bugs. And these edge case scenarios are far from the real scenarios.

smol-ninja: Also what are you referring to is the happy path and not the real world scenario. In real world, more orders are cancelled than filled (speaking from my personal experience).

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1438/changes/23128fa3bb6fe8fc319a1ee9bee52c6239f74389

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
andreivladbrg: good points on the "mimic" real scenarios, but i still believe we should prioritize `fillOrder` with more calls

but i won't debate further, as supposedly with sufficient ci runs, ultimately the scenarios would be found

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
andreivladbrg: thanks

andreivladbrg (approved):

bob/tests/escrow/invariant/handlers/EscrowHandler.sol
smol-ninja: I agree.

PR #1439, opened 2026-03-06
refactor(flow): remove zero check and event from recover

### LINKED ISSUE #1437: Get rid of zero transfer check in Flow recover function
The zero transfer check doesn't provide much value because this is an admin-only function. We can handle errors ourselves, the UX doesn't have to be impeccable.

**Lockup**

https://github.com/sablier-labs/lockup/blob/ee4ba10adb5888fe7613e927041673246c9b9df7/lockup/src/SablierLockup.sol#L250-L257

**Flow**

https://github.com/sablier-labs/lockup/blob/ee4ba10adb5888fe7613e927041673246c9b9df7/flow/src/SablierFlow.sol#L369-L381

smol-ninja:
Closes https://github.com/sablier-labs/lockup/issues/1437

### DIFF
--- a/flow/src/SablierFlow.sol
+++ b/flow/src/SablierFlow.sol
@@ -367,17 +367,12 @@ contract SablierFlow is
 
     /// @inheritdoc ISablierFlow
     function recover(IERC20 token, address to) external override onlyComptroller {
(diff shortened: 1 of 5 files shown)


andreivladbrg (approved):
👍

PR #1442, opened 2026-03-10
feat(bob): implement enterWithNativeToken function

andreivladbrg:


bob/src/SablierBob.sol
```diff
@@ -188,38 +189,48 @@ contract SablierBob is
...
+        notNull(vaultId)
+        onlyActive(vaultId)
+    {
+        // Cache the vault's adapter.
+        ISablierBobAdapter adapter = _vaults[vaultId].adapter;
 
-        // Check: the vault is still active after the price sync.
-        _revertIfSettledOrExpired(vaultId);
+        // Check: the vault has an adapter.
+        if (address(adapter) == address(0)) {
```
smol-ninja: Why revert when vault has no adapter? It will be useful on other chains to allow users to deposit native token such as POL where there will be no adapter.

bob/src/SablierBob.sol
```diff
@@ -188,38 +189,48 @@ contract SablierBob is
...
-            token.safeTransferFrom(msg.sender, address(this), amount);
+        // Check: the vault's token matches the adapter's WETH.
+        if (token != weth) {
+            revert Errors.SablierBob_VaultTokenNotWETH(vaultId, token, weth);
         }
 
-        // Interaction: mint share tokens to the caller.
-        _vaults[vaultId].shareToken.mint(vaultId, msg.sender, amount);
+        // Interaction: wrap the native token into WETH.
+        IWETH9(token).deposit{ value: msg.value }();
```
smol-ninja, line 209: This restricts this function to work only on ethereum. We wont be able to support other chains.

smol-ninja (changes requested):

bob/src/interfaces/ISablierBob.sol
```diff
@@ -130,6 +130,25 @@ interface ISablierBob is IComptrollerable, ISablierBobState {
...
+    ///
+    /// Requirements:
+    /// - `vaultId` must not reference a null vault.
+    /// - The vault must have ACTIVE status.
+    /// - The vault must have an adapter configured.
+    /// - `msg.value` must be greater than zero and fit in `uint128`.
+    /// - The vault's token must match the adapter's WETH address.
+    ///
+    /// @param vaultId The ID of the vault to deposit into.
+    function enterWithNativeToken(uint256 vaultId) external payable;
```
smol-ninja, line 148: Since this function is tightly coupled with weth contract and hence ethereum chain, what if we rename is to `enterWithEth` and revert if chain is not ethereum?

bob/src/SablierBob.sol
smol-ninja: What if `function enterWithNativeToken` simply wraps ETH into WETH without any other check and then call `enter`?

smol-ninja:
As discussed on Slack @andreivladbrg lets simplify it further with the following approach:

- `enterWithNativeToken` takes `vaultId` and `wToken`
- It calls `IWETH9(wToken).deposit{ value: msg.value }();`
- then it calls existing `enter` function

Wdyt?

We dont check if `wToken` is valid or not and because it calls `enter` with `vaultId` it will only pass if `vaultId.token` matches `wToken`

bob/src/SablierBob.sol
andreivladbrg: yes, thats what we want to achieve, no?

andreivladbrg: that's why we didn't add the weth or the wrapped native token address in the `SablierBob` contract

what's your point? what would you have done differently?

bob/src/SablierBob.sol
andreivladbrg: yeah, we can remove this check

bob/src/interfaces/ISablierBob.sol
andreivladbrg: > what if we rename is to enterWithEth and revert if chain is not ethereum

sounds good to me

andreivladbrg:
> enterWithNativeToken takes vaultId and wToken
It calls IWETH9(wToken).deposit{ value: msg.value }();
then it calls existing enter function

how is it simpler? we already have the `WETH` in the adapter

smol-ninja:
But we want to support all native tokens on all chains. Lets speak on Slack

andreivladbrg:
@smol-ninja forced pushed accordingly to our slack discussion, does it look better? if yes, could you mark the comments as resolved?

bob/src/interfaces/ISablierBob.sol
andreivladbrg: no longer needed

bob/src/SablierBob.sol
```diff
@@ -443,6 +436,47 @@ contract SablierBob is
...
+        // Effect: sync the price from oracle.
+        _syncPriceFromOracle(vaultId);
+
+        // Check: the vault is still active after the price sync.
+        _revertIfSettledOrExpired(vaultId);
+
+        // Interaction: transfer tokens to this contract or the adapter.
+        if (address(adapter) != address(0)) {
+            // Interaction: transfer tokens to the adapter.
+            token.safeTransferFrom(from, address(adapter), amount);
```
andreivladbrg: IMPORTANT: this will revert in case a token follows the latest OZ `ERC20` implementation, but thats a not problem for "wrapped" tokens as `WETH9` skips the allowance check when `src == msg.sender`

bob/src/SablierBob.sol
```diff
@@ -201,15 +201,16 @@ contract SablierBob is
...
         // Cache the vault's token.
         address token = address(_vaults[vaultId].token);
 
-        // Interaction: wrap the native token.
+        // Interaction: call the deposit function in the vault tokens assuming it follows the IWETH9 interface.
+        // Otherwise, it will revert.
         IWETH9(token).deposit{ value: msg.value }();
 
+        // Cast `msg.value` to `uint128`.
+        uint128 amount = msg.value.toUint128();
```
andreivladbrg, line 212: CEI not respected - `toUint128()` checks for overflow

bob/src/SablierBob.sol
smol-ninja: This sounds like a big problem if it would revert for tokens implementing latest OZ implementation. As a solution, what do you think about moving transfer code to the main function itself? It would also simplify the code a lot as we won't need `from` variable anymore

<details><summary>Click to see code</summary>

```solidity
    /// @inheritdoc ISablierBob
    function enter(uint256 vaultId, uint128 amount)
        external
        override
        nonReentrant
        notNull(vaultId)
        onlyActive(vaultId)
    {
        // Enter the vault.
        _enter({ vaultId: vaultId, from: msg.sender, amount: amount, token: _vaults[vaultId].token });

        if (address(adapter) != address(0)) {
            token.safeTransferFrom(msg.sender, address(adapter), amount);
        } else {
            token.safeTransferFrom(from, address(this), amount);
        }
    }

    /// @inheritdoc ISablierBob
    function enterWithNativeToken(uint256 vaultId)
        external
        payable
        override
        nonReentrant
        notNull(vaultId)
        onlyActive(vaultId)
    {
        // Cache the vault's token.
        address token = address(_vaults[vaultId].token);

        // Interaction: call the deposit function in the vault tokens assuming it follows the IWETH9 interface.
        // Otherwise, it will revert.
        IWETH9(token).deposit{ value: msg.value }();

        // Cast `msg.value` to `uint128`.
        uint128 amount = msg.value.toUint128();

        // Enter the vault.
        _enter({ vaultId: vaultId, from: address(this), amount: amount, token: IERC20(token) });

        if (address(adapter) != address(0)) {
            token.safeTransfer(address(adapter), amount);
        }
    }
```
</details>

bob/src/SablierBob.sol
smol-ninja: It does not matter here. Does it? `_enter` has its own CEI. So we have CICEI. Thoughts?

andreivladbrg: fair enough

bob/src/SablierBob.sol
andreivladbrg: as discussed on slack, agree to make the change

tho, i believe this version is simpler: 67c9df54ee98e7d4767c9f1f465e1cca34de790c

smol-ninja (approved):
LGTM

bob/src/SablierBob.sol
smol-ninja: Thanks

PR #1443, opened 2026-03-10
feat(airdrops): add sponsor function to merkle base contract

smol-ninja:
See https://github.com/sablier-labs/ui/pull/345#issuecomment-4023465933 for motivation.

### DIFF
--- a/airdrops/src/abstracts/SablierMerkleBase.sol
+++ b/airdrops/src/abstracts/SablierMerkleBase.sol
@@ -170,16 +170,7 @@ abstract contract SablierMerkleBase is
     }
 
     /// @inheritdoc ISablierMerkleBase
-    function sponsor(
-        IERC20 token,
-        uint128 amount,
-        address biller
-    )
-        external
-        override
-        notZeroAddress(biller)
-        notZeroAddress(address(token))
-    {
+    function sponsor(IERC20 token, uint128 amount, address biller) external override notZeroAddress(biller) {
         // Check: the amount is not zero.
         if (amount == 0) {
             revert Errors.SablierMerkleBase_SponsorAmountZero();
--- a/airdrops/src/abstracts/SablierMerkleBase.sol
+++ b/airdrops/src/abstracts/SablierMerkleBase.sol
@@ -169,6 +169,29 @@ abstract contract SablierMerkleBase is
         });
     }
 
+    /// @inheritdoc ISablierMerkleBase
+    function sponsor(
+        IERC20 token,
+        uint128 amount,
+        address biller
+    )
+        external
+        override
+        notZeroAddress(biller)
+        notZeroAddress(address(token))
+    {
+        // Check: the amount is not zero.
+        if (amount == 0) {
+            revert Errors.SablierMerkleBase_SponsorAmountZero();
+        }
+
+        // Interaction: transfer the tokens from the caller to the biller address.
+        token.safeTransferFrom({ from: msg.sender, to: biller, value: amount });
+
+        // Log the sponsorship.
+        emit Sponsor({ caller: msg.sender, token: token, amount: amount, biller: biller });
+    }
+
     /*//////////////////////////////////////////////////////////////////////////
                             PRIVATE READ-ONLY FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
--- a/airdrops/src/interfaces/ISablierMerkleBase.sol
+++ b/airdrops/src/interfaces/ISablierMerkleBase.sol
@@ -19,6 +19,9 @@ interface ISablierMerkleBase is IAdminable {
     /// @notice Emitted when the min USD fee is lowered by the comptroller.
     event LowerMinFeeUSD(address indexed comptroller, uint256 newMinFeeUSD, uint256 previousMinFeeUSD);
 
+    /// @notice Emitted when campaign owner sponsors the claim fees for eligible recipients.
+    event Sponsor(address indexed caller, IERC20 indexed token, uint128 amount, address indexed biller);
+
     /*//////////////////////////////////////////////////////////////////////////
                                 READ-ONLY FUNCTIONS
     //////////////////////////////////////////////////////////////////////////*/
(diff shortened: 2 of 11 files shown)


airdrops/src/interfaces/ISablierMerkleBase.sol
```diff
@@ -102,4 +105,25 @@ interface ISablierMerkleBase is IAdminable {
     /// - The new fee must be less than the current {minFeeUSD}.
     /// @param newMinFeeUSD The new min USD fee to set, denominated in 8 decimals.
     function lowerMinFeeUSD(uint256 newMinFeeUSD) external;
+
+    /// @notice Sponsors the claim fees for eligible recipients.
+    ///
+    /// @dev Emits a {Sponsor} event.
+    ///
+    /// Notes:
+    /// - This function only makes the payment. The claim fees are updated only after the payment has been verified
```
andreivladbrg, line 114: "The claim fees are updated only after the payment has been verified"

is this information needed?

why not simply a "mirror" `ERC20.transferFrom` function in the campaigns instead of "sponsor"

andreivladbrg (commented):
we do not need `batch` as well, right?

smol-ninja:
> we do not need `batch` as well, right?

Not as far as I know.

airdrops/src/interfaces/ISablierMerkleBase.sol
smol-ninja: > why not simply a "mirror" ERC20.transferFrom function in the campaigns instead of "sponsor"

Because we need a method to distinguish between random transfers from sponsored txns. Without that how would you know which transfer is to sponsor campaign fees and which is not?

> The claim fees are updated only after the payment has been verified

Yes to make sure that campaign owner does not confuse it with automated fee updates. The function is just to allow fee payment and then indexer need to pick it up, update things in the UI and mb we need to intervene into updating the fees via fee dashboard api.

airdrops/src/abstracts/SablierMerkleBase.sol
```diff
@@ -169,6 +169,29 @@ abstract contract SablierMerkleBase is
...
+    /// @inheritdoc ISablierMerkleBase
+    function sponsor(
+        IERC20 token,
+        uint128 amount,
+        address biller
+    )
+        external
+        override
+        notZeroAddress(biller)
+        notZeroAddress(address(token))
```
andreivladbrg: isn't this check redundant?

`token.safeTransferFrom` would revert anyway

airdrops/src/interfaces/ISablierMerkleBase.sol
andreivladbrg: > a method to distinguish between random transfers

well, it would be distinguishable as long the function emits an event, and the `transferFrom` was called from campaign -> contract
my idea was to make it more general purpose

but alright, both work, so lets keep current `sponsor`

airdrops/src/abstracts/SablierMerkleBase.sol
```diff
@@ -169,6 +169,29 @@ abstract contract SablierMerkleBase is
...
 
+    /// @inheritdoc ISablierMerkleBase
+    function sponsor(
+        IERC20 token,
+        uint128 amount,
+        address biller
+    )
+        external
+        override
+        notZeroAddress(biller)
```
andreivladbrg: this modifier reverts with `SablierMerkleBase_ToZeroAddress` - isn't this not very precise?

airdrops/src/interfaces/ISablierMerkleBase.sol
```diff
@@ -102,4 +105,25 @@ interface ISablierMerkleBase is IAdminable {
...
+    /// off-chain.
+    /// - Transferring a token other than the designated token will result in a loss of funds. Refer to the website
+    /// for details on the correct token to use for this function.
```
andreivladbrg: what does "designated token" mean here?

`token` input param can be any token, either `USDC` or `DOGE`

IMO it would be simpler to just say: "Refer to the Sablier website in order to sponsor with the correct token."

airdrops/tests/integration/concrete/campaign/shared/sponsor/sponsor.t.sol
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    function test_RevertWhen_AmountZero() external whenBillerNotZeroAddress whenTokenNotZeroAddress {
+        vm.expectRevert(abi.encodeWithSelector(Errors.SablierMerkleBase_SponsorAmountZero.selector));
+        merkleBase.sponsor({ token: usdc, amount: 0, biller: address(comptroller) });
+    }
+
+    function test_WhenAmountNotZero() external whenBillerNotZeroAddress whenTokenNotZeroAddress {
+        // Fund the caller with USDC and approve the campaign.
+        deal({ token: address(usdc), to: users.campaignCreator, give: SPONSOR_AMOUNT });
```
andreivladbrg: the users are already funded with tokens

andreivladbrg (commented):
left few more comments:

airdrops/src/abstracts/SablierMerkleBase.sol
andreivladbrg: 04b22823f1dafdd226b5ab892090e3cbc28d4c36

airdrops/tests/integration/concrete/campaign/shared/sponsor/sponsor.t.sol
andreivladbrg: 04b22823f1dafdd226b5ab892090e3cbc28d4c36

airdrops/src/interfaces/ISablierMerkleBase.sol
smol-ninja: Yeah I have no problem with your approach too so asking Razvan would be the best option. Feel free to suggest it on https://github.com/sablier-labs/ui/pull/345#issuecomment-4023465933

airdrops/src/abstracts/SablierMerkleBase.sol
smol-ninja: `SablierMerkleBase_ToZeroAddress` reverts when called on `sponsor` function and since `sponsor` function only has `biller` as the address parameter, I think its quite precise telling whats wrong with the function call.

airdrops/src/interfaces/ISablierMerkleBase.sol
smol-ninja: "designated token" mean an acceptable token for billing which could be USDC or USDT or BUSD upto billing team to decide.

> IMO it would be simpler to just say: "Refer to the Sablier website in order to sponsor with the correct token."

Thats the second sentence but first sentence is equally important.

airdrops/src/abstracts/SablierMerkleBase.sol
andreivladbrg: i meant the custom error name - the `To` is from `claimTo` function - and here it would be `BillerZeroAddress`

airdrops/src/interfaces/ISablierMerkleBase.sol
andreivladbrg: IMO "with the correct token" is sufficient enough

i am fan of this:  
image

airdrops/src/abstracts/SablierMerkleBase.sol
smol-ninja: Yeah thats a good point. But I thought why not just use an existing relevant error name instead of adding a new one just because variable is called `Biller`.

airdrops/src/interfaces/ISablierMerkleBase.sol
smol-ninja: OK fair point lets do it

andreivladbrg: 061741f4ed7807e02fde7c462f06ed762860090f

airdrops/src/abstracts/SablierMerkleBase.sol
andreivladbrg: yeah, its a small thing, lets keep it like this

airdrops/src/interfaces/ISablierMerkleBase.sol
andreivladbrg: resolving

andreivladbrg (approved):

andreivladbrg (approved):

PR #1447, opened 2026-03-11
refactor(examples): update  examples with latest changes

andreivladbrg:
q: what should we do about the CI? completely ignore for examples?

this should be part of https://github.com/sablier-labs/lockup/pull/1446/ but for a nicer diff, created as a separate PR


misc/examples/package.json
```diff
@@ -11,18 +11,18 @@
...
   },
   "dependencies": {
     "@openzeppelin/contracts": "5.3.0",
     "@prb/math": "4.1.0",
-    "@sablier/airdrops": "2.0.1",
-    "@sablier/flow": "2.0.1",
-    "@sablier/lockup": "3.0.1"
+    "@sablier/airdrops": "file:../../airdrops",
+    "@sablier/flow": "file:../../flow",
+    "@sablier/lockup": "file:../../lockup"
```
andreivladbrg: these should be updated with the package version once we have them, so that we don't have to keep the example "in-sync" with all changes we make in the protocols

justfile
```diff
@@ -125,3 +125,8 @@ setup-env:
         [ ! -L "$dir/.env" ] && ln -sf ../.env "$dir/.env" || true
         [ ! -L "$dir/.prettierignore" ] && ln -sf ../.prettierignore "$dir/.prettierignore" || true
     done
+    # Create symlinks in misc packages
+    for dir in misc/examples misc/benchmarks; do
```
andreivladbrg: i believe we should not add the examples module (in the root `justfile`) as per the "in-sync" argument, to be able to run `just-build-all`

tho, we can include the module for formatting command

andreivladbrg: feel free to suggest any updates on the `misc/examples/justfile` tho

misc/examples/package.json
smol-ninja: Updated in my commits.

justfile
smol-ninja: > i believe we should not add the examples module (in the root justfile) as per the "in-sync" argument, to be able to run just-build-all

Agree.

> feel free to suggest any updates on the misc/examples/justfile tho

I made tons of changes in my comments which has turned `misc` into a package instead of `examples` and `benchmarks` into their own packages. I believe this approach lowers the maintenance cost for us. I have also updated justfile and foundry.toml.

misc/examples/lockup/LockupLinearStreamCreator.sol
```diff
@@ -10,9 +10,9 @@ import { LockupLinear } from "@sablier/lockup/src/types/LockupLinear.sol";
 /// @dev This code is referenced in the docs:
 /// https://docs.sablier.com/guides/lockup/examples/create-stream/lockup-linear
 contract LockupLinearStreamCreator {
-    // Mainnet addresses
-    IERC20 public constant DAI = IERC20(0x6B175474E89094C44Da98b954EedeAC495271d0F);
-    ISablierLockup public constant LOCKUP = ISablierLockup(0xcF8ce57fa442ba50aCbC57147a62aD03873FfA73);
+    // Sepolia addresses
+    IERC20 public constant DAI = IERC20(0x68194a729C2450ad26072b3D33ADaCbcef39D574);
+    ISablierLockup public constant LOCKUP = ISablierLockup(0xAcDc1b0686D38a4aDE97e73e242b30A96761Be64);
```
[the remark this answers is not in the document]
smol-ninja: fixed.

misc/examples/lockup/LockupLinearStreamCreator.sol
```diff
@@ -37,9 +37,11 @@ contract LockupLinearStreamCreator {
...
             cliff: 0, // Setting a cliff of 0
             total: 52 weeks // Setting a total duration of ~1 year
-         });
+        });
 
         // Create the LockupLinear stream using a function that sets the start time to `block.timestamp`
-        streamId = LOCKUP.createWithDurationsLL(params, unlockAmounts, durations);
+        streamId = LOCKUP.createWithDurationsLL({
+            params: params, unlockAmounts: unlockAmounts, granularity: 1 seconds, durations: durations
+        });
```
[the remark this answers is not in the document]
smol-ninja: 1 seconds is fine.

misc/examples/lockup/StreamManagementWithHook.t.sol
```diff
@@ -37,7 +37,7 @@ contract StreamManagementWithHookTest is Test {
         token = new MockERC20(bob);
 
         // Load Sablier Lockup
-        sablierLockup = ISablierLockup(0xcF8ce57fa442ba50aCbC57147a62aD03873FfA73);
+        sablierLockup = ISablierLockup(0xAcDc1b0686D38a4aDE97e73e242b30A96761Be64);
```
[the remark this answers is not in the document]
smol-ninja: fixed.

misc/examples/package.json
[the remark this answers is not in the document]
smol-ninja: fixed

misc/examples/airdrops/MerkleCreator.sol
```diff
@@ -22,15 +25,43 @@ contract MerkleCreator {
...
     ISablierFactoryMerkleLT public constant LT_FACTORY =
-        ISablierFactoryMerkleLT(0x336d464276e2c7C76927d975Ef866Df8a7Ecf8DD);
+        ISablierFactoryMerkleLT(0x239BD5431aDa12F09cA95d0a5d4388A5644268e9);
     ISablierFactoryMerkleVCA public constant VCA_FACTORY =
-        ISablierFactoryMerkleVCA(0x91FdBd7077d615f951a0defA81Ec30Bfd68dbd8D);
-    ISablierLockup public constant LOCKUP = ISablierLockup(0xcF8ce57fa442ba50aCbC57147a62aD03873FfA73);
+        ISablierFactoryMerkleVCA(0xe60Df8e04cE1616a06db8AD11ce71c05dDcB5D88);
+    ISablierLockup public constant LOCKUP = ISablierLockup(0x93b37Bd5B6b278373217333Ac30D7E74c85fBDCB);
+
+    function createMerkleExecute() public returns (ISablierMerkleExecute merkleExecute) {
```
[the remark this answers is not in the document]
smol-ninja: its fine

misc/justfile
```diff
@@ -0,0 +1,96 @@
...
+# Benchmark both Flow and Lockup
+[group("benchmarks")]
+benchmark-all:
+  just benchmark flow
+  just benchmark lockup
+
+# Benchmark a specific protocol
+[group("benchmarks")]
+benchmark *args:
+  FOUNDRY_PROFILE=benchmarks-{{ args }} forge test --show-progress -vv
```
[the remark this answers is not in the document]
smol-ninja: its ok

smol-ninja (approved):
LGTM now. Thanks for starting this PR work Andrei.

misc/justfile
andreivladbrg, line 1: should we have a `build-all` command to build both benchmark and examples?

smol-ninja: We can but its not necessary. Even with other packages, when do you use `build-all`? I usually use individually build or even in the ci we use individual build.

andreivladbrg:
@smol-ninja you did not update the benchmark addresses. was it intentional?

misc/justfile
andreivladbrg: fair enough

my problem is that the justfile from the misc folder imports `import "./node_modules/@sablier/devkit/just/evm.just"` which would display (when you run `just --list`) bunch of commands that would not work

smol-ninja:
> you did not update the benchmark addresses. was it intentional

I have, just running the benchmarks locally. Pushing it once its completed.

misc/justfile
smol-ninja: Which recipes from `just --list` can be problematic?

https://app.warp.dev/block/QFkmH2VoHA5BCPH0KAj6PX

There are separate group for benchmarks and examples which is very explicit. And even under foundry group, there is no build-all recipe.

andreivladbrg: these 

image

---

image

smol-ninja: And how would you fix that? What should `just build` do?

PR #1448, opened 2026-03-12
fix(bob): check the zero oracle price in unstakeFullAmount

andreivladbrg:
addresses the comment https://github.com/Cyfrin/audit-2026-02-sablier-bob-2/issues/26#issuecomment-4044121006


andreivladbrg:
@smol-ninja fyi: before, in the fork tests we were only testing the case where we had a zero price because we were warping in `Base_Test` to `FEB_1_2026` after forking the chain at the latest block

thus `safeOraclePrice` returns because `block.timestamp < _updatedAt`

smol-ninja:
> we were warping in `Base_Test` to `FEB_1_2026` after forking the chain at the latest block

Good catch. We should do the same for Escrow fork tests because this is technically correct approach. Pushed directly to `staging` in https://github.com/sablier-labs/lockup/commit/261af5a31f51d1e9725efff16fd7d782b49a5042

PR #1449, opened 2026-03-12
chore(benchmarks): remove config files

andreivladbrg:
unlike the examples update, i've not included the latest protocols versions as we need first to deploy the contracts on mainnet

i just removed the config files


smol-ninja (approved):
I am not sure why you did this but I trust you.

andreivladbrg:
> I am not sure why you did this but I trust you.

similar to all protocols: `airdrops/` `flow/` `lockup/` - they have the dot (config) files declared at the root of the monorepo

PR #1453, opened 2026-03-14
chore: bump package versions

andreivladbrg:
docs: add new versions in CHANGELOG files

### DIFF
--- a/airdrops/CHANGELOG.md
+++ b/airdrops/CHANGELOG.md
@@ -1,24 +1,30 @@
 # Changelog
 
-All notable changes to this project will be documented in this file. The format is based on
-[Common Changelog](https://common-changelog.org/).
+All notable changes to this project will be documented in this file.
 
-## [3.0.0] - 2026-03-14
+The format is based on [Common Changelog](https://common-changelog.org/).
+
+## [3.0.0] - 2026-03-16
 
 ### Changed
 
+- **Breaking**: Change return type of `totalForgoneAmount()` in `SablierMerkleVCA` from `uint256` to `uint128` ([#1363](https://github.com/sablier-labs/lockup/pull/1363))
 - Refactor `SablierMerkleBase` and `SablierMerkleLockup` constructors to accept `ConstructorParams` struct ([#1403](https://github.com/sablier-labs/lockup/pull/1403))
-- Refactor `DataTypes` into separate type files (`MerkleBase`, `MerkleExecute`, `MerkleInstant`,
-  `MerkleLL`, `MerkleLT`, `MerkleLockup`, `MerkleVCA`) ([#1408](https://github.com/sablier-labs/lockup/pull/1408))
-  - `DataTypes` is kept for backward compatibility and will be deprecated in future versions.
-- Add `granularity` parameter to `SablierMerkleLL` for Lockup Linear streams ([#1366](https://github.com/sablier-labs/lockup/pull/1366))
+- Refactor `DataTypes` into separate type files ([#1408](https://github.com/sablier-labs/lockup/pull/1408))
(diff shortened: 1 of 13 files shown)


flow/CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
...
+
+### Added
+
+- Add `transferFromPayable` function ([#1384](https://github.com/sablier-labs/lockup/pull/1384))
+
+### Removed
+
+- Remove `Recover` event from `ISablierFlow` ([#1439](https://github.com/sablier-labs/lockup/pull/1439))
+- Remove surplus zero check from `recover` function ([#1439](https://github.com/sablier-labs/lockup/pull/1439))
+- Remove `SablierFlow_SurplusZero` error ([#1439](https://github.com/sablier-labs/lockup/pull/1439))
```
andreivladbrg: i think we can merge these 3 into a single point, wdyt? @smol-ninja

airdrops/package.json
smol-ninja, line 16: Note that release should be in the following order so that we can update these deps while deploying

- utils
- lockup
- airdrop / flow / bob

This means we should merge everything into main first and update these deps as we start deploying.

airdrops/CHANGELOG.md
```diff
@@ -3,6 +3,23 @@
 All notable changes to this project will be documented in this file. The format is based on
 [Common Changelog](https://common-changelog.org/).
 
+## [3.0.0] - 2026-03-14
+
+### Changed
+
+- Refactor `SablierMerkleBase` and `SablierMerkleLockup` constructors to accept `ConstructorParams` struct ([#1403](https://github.com/sablier-labs/lockup/pull/1403))
```
smol-ninja, line 12: How is this relevant to changelog? IMO changelog should only focus on the changes that affect users and since this is an interface refactor, we should skip this.

smol-ninja (changes requested):
I created a PR for my suggestions as I found that better than commenting here (since they are many):

https://github.com/sablier-labs/lockup/pull/1456

flow/CHANGELOG.md
smol-ninja: I don't think we need to include error point in changelog. If zero surplus check is removed from `recover` function that inherently means the error is removed too.

Also the first one "Remove `Recover` event from `ISablierFlow`" is a breaking change as hypothetically (since its only used internally) it breaks indexers from tracking recover.

airdrops/CHANGELOG.md
```diff
@@ -3,6 +3,23 @@
...
+- Refactor `DataTypes` into separate type files (`MerkleBase`, `MerkleExecute`, `MerkleInstant`,
+  `MerkleLL`, `MerkleLT`, `MerkleLockup`, `MerkleVCA`) ([#1408](https://github.com/sablier-labs/lockup/pull/1408))
+  - `DataTypes` is kept for backward compatibility and will be deprecated in future versions.
```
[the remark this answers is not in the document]
smol-ninja: Good point. Fixed in https://github.com/sablier-labs/lockup/pull/1456/changes/499131407fdf41ce759c6b16d7b76d96b1f822ab which will later be merged into this PR. So mark it as resolved.

flow/CHANGELOG.md
smol-ninja: Also, since its a breaking change, shouldn't Flow version be `3.0.0`?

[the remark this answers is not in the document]
smol-ninja: Good point. Will discuss. See my [comment here](https://github.com/sablier-labs/lockup/pull/1453#discussion_r2939039283).

smol-ninja (approved):

Issue #1374, opened 2026-01-07
Add monorepo readme


smol-ninja:
Currently there is no `README.md` in the monorepo. So the task is to add a `README` file.

PaulRBerg:
can you add some description plz

andreivladbrg:
addressed in #1458

PR #1458, opened 2026-03-16
docs: monorepo readme

andreivladbrg:
closes https://github.com/sablier-labs/lockup/issues/1374


smol-ninja:
Should we remove badges from each package's readme file? Lets do it after https://github.com/sablier-labs/lockup/issues/1459.

PR #1464, opened 2026-03-18
chore(utils): add denergy chain

andreivladbrg:
we will need to bump the release for this as well - but we will do it after we fix the readme files as well


smol-ninja (approved):
bump only `utils`?

PR #1472, opened 2026-03-19
Update fork tests to use mainnet addresses

### LINKED ISSUE #1335: Fork tests fail on profiles without `via_ir = true`
## Problem

Fork tests fail for profiles that do not have `via_ir` set to `true`, such as `lite` and `default`. This is due to the following reasons:

1. The factory contract on `mainnet` was deployed with `via_ir = true`.
2. During tests, the expected Merkle contract is computed using `vm.computeCreate2Address`, which relies on the `creationBytecodeHash`.
3. The `creationBytecodeHash` is determined by the profile ([see here](https://github.com/sablier-labs/airdrops/blob/main/tests/Base.t.sol#L297)) and, in turn, depends on the value of `via_ir`.
4. Since the forked contracts were deployed with `via_ir = true`, functions like `getMerkleInstantBytecode` must return bytecode generated with `via_ir = true`.

As a result, fork tests only succeed when using the `test-optimized` profile.

### Solution 1

Use `computeMerkle<name>` functions to compute the merkle contract that will be deployed instead of using `getMerkleInstantBytecode` function. That should solve the problem. 

### Solution 2

The ideal approach would involve automatically setting `via_ir` to `true` in Foundry during fork tests. While I haven't figured out an exact solution, one possible solution is to modify `Just` script to run fork tests with the `FOUNDRY_VIA_IR=true` variable (apply to all profiles): 

```Justfile
just fork-test:
  FOUNDRY_VIA_IR=true forge test --mt testFork
```

This would would require separating fork tests from other tests:

```Justfile
just test-lite:
  forge test --nmt testFork
``` 

### Solution 3

An alternative solution involves using `vm.activeFork` to detect active fork tests, and then [here](https://github.com/sablier-labs/airdrops/blob/main/tests/Base.t.sol#L297), we can use `vm.etch` to override the forked contract bytecode with the value `bytes.concat(type(SablierMerkleInstant).creationCode, constructorArgs);` value.

```python
try vm.activeFork():
    # this would reset the onchain bytecode
    vm.etch(merkleFactory, bytes.concat(type(SablierMerkleInstant).creationCode, constructorArgs))
    return return bytes.concat(type(SablierMerkleInstant).creationCode, constructorArgs)
except:
    if not isTestOptimizedProfile():
        return bytes.concat(type(SablierMerkleInstant).creationCode, constructorArgs)
    else:
        return bytes.concat(
             vm.getCode("out-optimized/SablierMerkleInstant.sol/SablierMerkleInstant.json"), constructorArgs
        )
```

smol-ninja:
Updated fork tests to use the mainnet contracts. Also closes https://github.com/sablier-labs/evm-monorepo/issues/1335

### DIFF
--- a/airdrops/tests/fork/merkle-campaign/MerkleExecute.t.sol
+++ b/airdrops/tests/fork/merkle-campaign/MerkleExecute.t.sol
@@ -60,8 +60,6 @@ abstract contract MerkleExecute_Fork_Test is MerkleBase_Fork_Test {
             token: FORK_TOKEN
         });
 
-        // Since the mainnet contracts are deployed using `via_ir`, use `computeMerkleExecute` function to compute the
-        // expected address.
         vars.expectedMerkleCampaign =
             factoryMerkleExecute.computeMerkleExecute(params.campaignCreator, constructorParams);
 
--- a/airdrops/tests/fork/merkle-campaign/MerkleExecute.t.sol
+++ b/airdrops/tests/fork/merkle-campaign/MerkleExecute.t.sol
@@ -60,8 +60,10 @@ abstract contract MerkleExecute_Fork_Test is MerkleBase_Fork_Test {
             token: FORK_TOKEN
         });
 
+        // Since the mainnet contracts are deployed using `via_ir`, use `computeMerkleExecute` function to compute the
+        // expected address.
         vars.expectedMerkleCampaign =
-            computeMerkleExecuteAddress({ params: constructorParams, campaignCreator: params.campaignCreator });
+            factoryMerkleExecute.computeMerkleExecute(params.campaignCreator, constructorParams);
 
         vm.expectEmit({ emitter: address(factoryMerkleExecute) });
         emit ISablierFactoryMerkleExecute.CreateMerkleExecute({
--- a/utils/tests/fork/Comptroller.t.sol
+++ b/utils/tests/fork/Comptroller.t.sol
@@ -28,7 +28,7 @@ contract Comptroller_Fork_Test is Base_Test {
         vm.createSelectFork({ urlOrAlias: "ethereum" });
 
         // Set the current mainnet admin address.
-        currentMainnetAdmin = 0x79Fb3e81aAc012c08501f41296CCC145a1E15844;
+        currentMainnetAdmin = 0x58290bbdb51A4c6B022A81e9cDeD24BE19Ca57fd;
 
         // Deploy the `BaseScriptMock` contract.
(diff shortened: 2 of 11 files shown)


airdrops/tests/fork/merkle-campaign/MerkleExecute.t.sol
```diff
@@ -60,8 +60,10 @@ abstract contract MerkleExecute_Fork_Test is MerkleBase_Fork_Test {
             token: FORK_TOKEN
         });
 
+        // Since the mainnet contracts are deployed using `via_ir`, use `computeMerkleExecute` function to compute the
+        // expected address.
```
andreivladbrg: let's remove these comments, and just keep `factoryMerkleExecute.computeMerkleExecute` no matter the addresses are loaded or deployed

utils/tests/fork/Comptroller.t.sol
```diff
@@ -61,7 +63,7 @@ contract Comptroller_Fork_Test is Base_Test {
         if (protocol == ISablierComptroller.Protocol.Staking) {
             assertEq(minFeeInWei, 0, "Staking: minFeeInWei > 0");
         } else {
-            assertGt(minFeeInWei, 0, "Lockup, Flow, Airdrops: minFeeInWei == 0");
+            assertGt(minFeeInWei, 0, "Airdrops, Bob, Flow, Lockup: minFeeInWei == 0");
```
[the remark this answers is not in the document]
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1472/changes/7b0920081061ef569e39e6c7a34ec3ae529d6d5e

andreivladbrg (commented):
generally looks good, two questions:

wdyt about adding the contract addresses in utils' test constants?

this way we'd have them only in one place

smol-ninja:
> wdyt about adding the contract addresses in utils' test constants

It would increase maintenance cost. With every change, we would have to bump utils. So I am not in favour of that approach.

airdrops/tests/fork/merkle-campaign/MerkleExecute.t.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1472/changes/ed37730768567b484c8f72494b658e73d45e0423

andreivladbrg (approved):
> It would increase maintenance cost. With every change, we would have to bump utils. So I am not in favour of that approach.

yeah, fair enough lets keep it this way

PR #1474, opened 2026-03-20
ci: buildbear workflow integration

smol-ninja:
Copy of https://github.com/sablier-labs/evm-monorepo/pull/1380

### DIFF
--- a/.github/workflows/bb-ci-flow.yml
+++ b/.github/workflows/bb-ci-flow.yml
@@ -1,0 +1,34 @@
+name: "BB CI: Flow"
+run-name: "CI Flow :: ${{ github.event.head_commit.message || github.event.pull_request.title }}"
+
+concurrency:
+  cancel-in-progress: true
+  group: ${{github.workflow}}-${{github.ref}}
+
+on:
+  workflow_dispatch:
+  pull_request:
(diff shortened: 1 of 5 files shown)


smol-ninja:
As discussed over Telegram with @itsshantanu, Buildbear currently supports debugging integration and unit tests only. While this appears to be a good use case, I would need Buildbear to support debugging following test suites before I can consider merging the PR:

- **Fuzz tests** (debugging fuzz tests is a moderate effort task due to fuzzed inputs and resetting bounds within the tests)
- **Fork tests** (debugging fork tests is a moderate effort task due to fuzzed inputs, bounds within the tests and dynamic onchain states)
- **Invariant tests** (debugging invariant tests is a high effort task due to fuzzed inputs, bounds within the tests, and sequences of calls)

Although Buildbear can still be used for integration and unit tests, debugging these tests is relatively simple since the inputs are almost always fixed, and the test sequence is straightforward. Therefore it offers little to no added value. At the same time, it would still run fork tests in the Buildbear sandbox, using and thereforce consuming our RPC calls.

For these reasons, I am closing this PR for now and will wait until Buildbear supports the test suites that are critical for our requirements.

PR #1478, opened 2026-03-21
chore: add denergy previous nft descriptors

smol-ninja:


andreivladbrg (commented):
missing update in the chaind id lib and the comptroller/admin list (if needed)

smol-ninja:
> missing update in the chaind id lib and the comptroller/admin list (if needed)

yeah. Also need to publish utils to npm and then update versions in package json files. I guess you are already on it right?

smol-ninja:
@andreivladbrg a reminder to merge this before publishing release

PR #1481, opened 2026-03-23
chore: bump evm utils dep in bob lockup and flow

andreivladbrg:
i have not included airdrops as we will need to bump lockup there too - so i will do it after we have the 4.0.1 on NPM

note: this PR _must not_ be squased as i included a commit in the changelog

### DIFF
--- a/bob/CHANGELOG.md
+++ b/bob/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Common Changelog](https://common-changelog.org/).
 
 ### Changed
 
-- Bump package version for NPM release ([bd886f1](https://github.com/sablier-labs/evm-monorepo/commit/bd886f1912d11aa9362a8c2024ed317e9f8dffd5))
+- Bump utils package ([#1481](https://github.com/sablier-labs/evm-monorepo/pull/1481))
 
 ## [1.0.0] - 2026-03-16
 
--- a/bob/CHANGELOG.md
+++ b/bob/CHANGELOG.md
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Common Changelog](https://common-changelog.org/).
 
+## [1.0.1] - 2026-03-23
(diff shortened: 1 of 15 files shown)


bob/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Common Changelog](https://common-changelog.org/).
 
+## [1.0.1] - 2026-03-23
+
+### Changed
+
+- Bump package version for NPM release ([bd886f1](https://github.com/sablier-labs/evm-monorepo/commit/bd886f1912d11aa9362a8c2024ed317e9f8dffd5))
```
smol-ninja: Why not a PR here? The commit hash may change over time (PR merge etc). So we should refer https://github.com/sablier-labs/evm-monorepo/pull/1481/.

smol-ninja: Shouldn't we say `Bump utils dependency`. Changelog should contain the reasons why package is bumped

smol-ninja (changes requested):

bob/CHANGELOG.md
andreivladbrg: both work, so we can change it

andreivladbrg: yeah, we should

andreivladbrg: 5ade63173e1d569e1a42b490029b56eb4d52288e

andreivladbrg: 5ade63173e1d569e1a42b490029b56eb4d52288e

smol-ninja (approved):
We should merge https://github.com/sablier-labs/evm-monorepo/pull/1478 first.

andreivladbrg:
> We should merge https://github.com/sablier-labs/evm-monorepo/pull/1478 first.

will merge after this one, so that the ci doesn't fail

PR #1482, opened 2026-03-23
fix: remove bun install from install-all just command

andreivladbrg:
the `install-all` command now runs infinitely

https://github.com/user-attachments/assets/a79de80e-db1f-4615-8e55-0c43fa06272d


smol-ninja (approved):
Good catch btw.

PR #1484, opened 2026-03-23
docs: bump monorepo changelog

andreivladbrg:
chore(misc): use latest packages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,54 +14,30 @@ For detailed per-package changelogs, see each package's `CHANGELOG.md`:
 
 ## March 23, 2026
 
-### Packages
-
-| Protocol | NPM Package                                                              | GitHub Tag        | Version |
-| -------- | ------------------------------------------------------------------------ | ----------------- | ------- |
-| Airdrops | [`@sablier/airdrops`](https://www.npmjs.com/package/@sablier/airdrops)   | `airdrops@v3.0.1` | 3.0.1   |
-| Bob      | [`@sablier/bob`](https://www.npmjs.com/package/@sablier/bob)             | `bob@v1.0.1`      | 1.0.1   |
-| Flow     | [`@sablier/flow`](https://www.npmjs.com/package/@sablier/flow)           | `flow@v3.0.1`     | 3.0.1   |
-| Lockup   | [`@sablier/lockup`](https://www.npmjs.com/package/@sablier/lockup)       | `lockup@v4.0.1`   | 4.0.1   |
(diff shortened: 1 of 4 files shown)


andreivladbrg:
@smol-ninja wdyt about `evm-monorepo@1.0.0` release tag for the full repo?

CHANGELOG.md
```diff
@@ -12,6 +12,41 @@ For detailed per-package changelogs, see each package's `CHANGELOG.md`:
 - [lockup](./lockup/CHANGELOG.md)
 - [utils](./utils/CHANGELOG.md)
 
+## March 23, 2026
```
smol-ninja, line 15: What do you think about not including this changelog at all and change Mar 16 to Mar 23 below.

This is noise imo.

smol-ninja (commented):
> wdyt about evm-monorepo@1.0.0 release tag for the full repo?

I agree but also see my comment below.

CHANGELOG.md
andreivladbrg: why not keep it? its similar to `package/CHANGELOG.md`

smol-ninja: Its noise to me. Having so many lines just to mention that packages are bumped is not helpful. IMO we can break some inconsistency here for the sake of cleanliness. Monorepo is not on NPM so we can choose to follow our own style here.

andreivladbrg: image

IMO having that big title and the dash line makes things easier to read, and the thing that's noisy to me is the table

andreivladbrg: we can re-introduce the:

```md
#### [airdrops@v3.0.1](https://www.npmjs.com/package/@sablier/airdrops/v/3.0.1)
```

and remove the tables, wdyt?

smol-ninja: OK to remove the table. But will you also be removing the bump changelogs? Its a first monorepo release, so lets keep it clean.

We already have mentioned "For detailed per-package changelogs, see each package's CHANGELOG.md:" so let it be part of detailed changelog which is inside each package.

> we can re-introduce the

Sure

If you still dont agree, can you plz tell me why do you think its a good idea to keep the bump stuffs in the monorepo changelog? Why not just call it `v1.0.0` and change the date to Mar 23?

andreivladbrg: we will publish the `evm-monorepo` in GH releases so IMO its good to have the dates and bumps to 1.0.1 1.0.2 etc.

smol-ninja: Though I don't agree (with my reasons above), lets go your way.

andreivladbrg: fb268a9ab13f96933fd5febe39c1664d08b988b0

smol-ninja (approved):
LGTM

PR #1486, opened 2026-03-24
docs: format changelog better

andreivladbrg:


smol-ninja (approved):
Assuming its just a formatting.

PR #1489, opened 2026-03-27
feat: aave adapter

andreivladbrg:
notes: aave v3 is used as v4 is still not live


bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+            IAavePoolDataProvider(AAVE_POOL_DATA_PROVIDER).getReserveTokensAddresses(address(vaultToken));
+        if (aToken == address(0)) {
+            revert Errors.SablierAaveAdapter_TokenNotSupportedByAave(address(vaultToken));
+        }
+
+        // Effect: snapshot the current global yield fee for this vault.
+        _vaultYieldFee[vaultId] = feeOnYield;
+
+        // Interaction: approve the Aave Pool to spend this token.
+        vaultToken.forceApprove(address(AAVE_POOL), type(uint128).max);
```
andreivladbrg: `AAVE_POOL`  is an upgradable contract - it might be safer to not `.approve()` max value, but to do it in the `.stake()` function

smol-ninja: What would be the risk if we max approve it?

andreivladbrg: aave upgrades the pool to a malicious contract (not plausible)

but, from our users POV - it can be seen as "piece of mind"

smol-ninja: > aave upgrades the pool to a malicious contract

Lets say that happens, what will it steal? Adapter would never hold any tokens other than the aTokens and for which there is no approve anyways. Therefore, I think there is no risk in max approving the Aave contract even if its upgradeable. Thoughts?

andreivladbrg: good point, yeah, its safe to keep it with max approve

andreivladbrg:
> notes: aave v3 is used as v4 is still not live

@smol-ninja i think we should keep using v3 as it would take time until the liquidity is migrated, wdyt?

image

smol-ninja:
> think we should keep using v3 as it would take time until the liquidity is migrated, wdyt?

May be. I have started reviewing the code so let me get back to you with my thoughts after I have good understanding of all the pieces including Aave releases.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+import { IAaveAToken } from "./interfaces/external/IAaveToken.sol";
+import { ISablierAaveAdapter } from "./interfaces/ISablierAaveAdapter.sol";
+import { ISablierBobAdapter } from "./interfaces/ISablierBobAdapter.sol";
+import { ISablierBobState } from "./interfaces/ISablierBobState.sol";
+import { Errors } from "./libraries/Errors.sol";
+
+/// @title SablierAaveAdapter
+/// @notice Aave V3 yield adapter for the SablierBob protocol.
+/// @dev This adapter supplies deposited tokens to Aave V3 lending pools to earn interest. Unlike the Lido adapter which
+/// only handles WETH, this adapter is token-agnostic: it retrieves the underlying token and aToken per vault at
```
smol-ninja: There is no need to say "Unlike the Lido adapter" as these two are independent. Imagine adding a third adapter where we are saying "unlike aave and lido".

The entire description should be rewritten only from the point of view of this adapter.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+    IAavePool public immutable override AAVE_POOL;
+
+    /// @inheritdoc ISablierAaveAdapter
+    address public immutable override AAVE_POOL_DATA_PROVIDER;
+
+    /// @inheritdoc ISablierBobAdapter
+    UD60x18 public constant override MAX_FEE = UD60x18.wrap(0.2e18);
+
+    /// @dev Aave uses RAY (1e27) precision for the liquidity index.
+    uint256 private constant RAY = 1e27;
```
smol-ninja, line 47: Why is it called `RAY`? Shouldn't we call it something of our own? Also, why is it a `private constant`? Since constants are part of the bytecode, why does it need to be private?

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+        feeOnYield = newFee;
+
+        // Log the event.
+        emit SetYieldFee(previousFee, newFee);
+    }
+
+    /// @inheritdoc ISablierBobAdapter
+    function stake(uint256 vaultId, address user, uint256 amount) external override onlySablierBob {
+        IERC20 vaultToken = _getVaultToken(vaultId);
+        IAaveAToken aToken = _getAToken(vaultToken);
```
smol-ninja: Since these two are so frequently used, what do you think about keeping them in storage when a vault is registered?

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+
+        // Snapshot the contract's total scaled balance before supply. This includes scaled balances from ALL vaults
+        // that use this token, but the before/after delta isolates exactly what this single deposit contributed.
+        uint256 scaledBefore = aToken.scaledBalanceOf(address(this));
+
+        // Interaction: supply tokens to the Aave Pool. The tokens are already held by this contract (transferred by
+        // SablierBob in `_enter`). Approval was set in `registerVault`.
```
smol-ninja: Ain't these comments verbose? IMO `Interaction: supply tokens to the Aave Pool` is more than enough.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+        AAVE_POOL.supply(address(vaultToken), amount, address(this), 0);
+
+        // The scaled balance delta is the exact scaled amount attributable to this deposit. Aave internally computes
+        // this as `amount * 1e27 / currentLiquidityIndex`. A deposit made when the index is higher yields fewer
+        // scaled tokens per underlying token — this is how deposit timing is encoded without storing timestamps.
+        //
+        // Example: if index = 1.05e27, depositing 200 tokens produces ~190.48 scaled tokens. A user who deposited
+        // 100 tokens when index = 1.0e27 has 100 scaled tokens. At redemption, the first user's 100 scaled tokens
```
smol-ninja: Same a lot of these comments are redundant. We can just mention that `scaledBalanceOf` returns non-rebasing balance of aTokens.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+
+    /// @inheritdoc ISablierBobAdapter
+    function unstakeFullAmount(uint256 vaultId)
+        external
+        override
+        onlySablierBob
+        returns (uint128 totalScaledBalance, uint128 amountReceivedFromUnstaking)
+    {
+        // Get total scaled balance for the vault.
+        totalScaledBalance = _vaultTotalScaledBalance[vaultId];
```
smol-ninja: Lets call it `totalVaultScaledBalance`.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+    /// @dev Scaled balance held for each user in each vault. The scaled balance encodes both deposit amount AND deposit
+    /// time: depositing at a higher liquidity index produces fewer scaled tokens per underlying token. This means
+    /// earlier depositors naturally accumulate more yield per token than later depositors, without needing to store
+    /// timestamps or index snapshots. See {stake} for how this value is derived.
+    mapping(uint256 vaultId => mapping(address user => uint128 scaledBalance)) internal _userScaledBalance;
+
+    /// @dev Total scaled balance held in each vault. This is the sum of all users' scaled balances for the vault.
+    /// Used as the denominator in proportional redemption calculations: each user's share of the withdrawn tokens
+    /// is `userScaled / totalScaled * totalTokensReceived`.
+    mapping(uint256 vaultId => uint128 totalScaledBalance) internal _vaultTotalScaledBalance;
```
smol-ninja: We should use `uint256` for scaled balances. RAY is 1e27 which means we can easily cross `uint128` type.

`(10**18)*(10**27)` exceeds `2**128` which represents 1 token with 18 decimals.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+        // Get total scaled balance for the vault.
+        totalScaledBalance = _vaultTotalScaledBalance[vaultId];
+        IERC20 vaultToken = _getVaultToken(vaultId);
+
+        // Convert the vault's total scaled balance back to underlying tokens using the current liquidity index:
+        //   currentValue = totalScaledBalance * currentIndex / 1e27
+        // This rounds down, so we never ask for more than the contract's aToken balance. The floor division may
+        // leave up to 1 wei of aToken dust, which is negligible compared to the yield generated.
+        uint256 currentIndex = AAVE_POOL.getReserveNormalizedIncome(address(vaultToken));
+        uint256 currentValue = uint256(totalScaledBalance) * currentIndex / RAY;
```
smol-ninja: Lets call it `aTokenBalance`

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,373 @@
...
+
+        // Interaction: withdraw only this vault's portion from Aave. We pass the exact calculated amount (not
+        // type(uint256).max) because this contract holds aTokens for multiple vaults of the same token. Using max
+        // would drain ALL vaults' positions.
+        uint256 actualWithdrawn = AAVE_POOL.withdraw(address(vaultToken), currentValue, address(this));
```
smol-ninja: Instead of `address(this)`, why not use `SABLIER_BOB` directly?

smol-ninja (changes requested):
Nice work. I guess the scaled balance made it easier. I have tons of feedback on the comments and naming of variables but I think you know that too. Left a few comments below on the implementation only.

bob/src/SablierAaveAdapter.sol
andreivladbrg: see the PR body message:

image

bob/src/SablierAaveAdapter.sol
andreivladbrg: this is how aave calls it, and IMO it makes sense to also do the same

bob/src/SablierAaveAdapter.sol
smol-ninja: I read that later. Sorry.

bob/src/SablierAaveAdapter.sol
smol-ninja: Aave's terminologies is very confusing. For ex. they call fixed balance `scaledBalance` and rebasing balance as `balance`. Therefore, my view is that we can go on with our own terminology that is familiar across Sablier systems.

But I will let you decide for this.

andreivladbrg: i don't have a preference, anything works

if you find it that confusing, feel free to come up with better terminology

bob/src/SablierAaveAdapter.sol
andreivladbrg: actually, this was the initial version i wrote, but then we would have the vault token stored in `SablierBob` and in the adapter as well - don't you think it's redundant?

bob/src/SablierAaveAdapter.sol
andreivladbrg: yes, they are - we can later edit them, if you agree with the actual code

bob/src/SablierAaveAdapter.sol
andreivladbrg: yes they are 
image

bob/src/SablierAaveAdapter.sol
andreivladbrg: thats a good point, and we should use `uint256`

bob/src/SablierAaveAdapter.sol
andreivladbrg: sure

bob/src/SablierAaveAdapter.sol
andreivladbrg: good idea, we can do this - we'd save the gas for a erc20 transfer

bob/src/SablierAaveAdapter.sol
smol-ninja: > don't you think it's redundant

Its gas saving too.

andreivladbrg: it depends on how many stakes are done:

```md
Option 1: Store locally

SSTORE (zero → non-zero, cold): 22,100 gas (one-time)
SLOAD (cold, per tx): 2,100 gas each read

Option 2: External call

CALL to cold address: 2,600 gas
SLOAD inside that contract (cold): 2,100 gas
ABI encode/decode + dispatch overhead: ~300 gas
Total per call: ~5,000 gas

So the break-even equation is:
22,100 + 2,100N = 5,000N
22,100 = 2,900N
N ≈ 7.6
```

so given we 7.6 is a relative small number, lets save both the vault and a tokens

andreivladbrg:
@smol-ninja i assume you agree with the rest of the logic besides these comments: [1](https://github.com/sablier-labs/evm-monorepo/pull/1489#discussion_r3023184678), [2](https://github.com/sablier-labs/evm-monorepo/pull/1489#discussion_r3029015953) and [3](https://github.com/sablier-labs/evm-monorepo/pull/1489#discussion_r3032268024)

so i will fix them, and then improve the natspec/comments + tests

bob/src/SablierAaveAdapter.sol
smol-ninja: > SSTORE (zero → non-zero, cold): 22,100 gas (one-time)

This is only when a vault is registered. So This extra overhead is only passed to the vault creator. The depositors only pay for `SLOAD` which is 2100 gas. So its gas saving even for the first stake.

Let me know if I am wrong here.

smol-ninja:
> i assume you agree with the rest of the logic besides these comments: https://github.com/sablier-labs/evm-monorepo/pull/1489#discussion_r3023184678, https://github.com/sablier-labs/evm-monorepo/pull/1489#discussion_r3029015953 and https://github.com/sablier-labs/evm-monorepo/pull/1489#discussion_r3032268024

Yes I agree with the implementation.

bob/src/SablierAaveAdapter.sol
andreivladbrg: there is a problem in the `getTotalYieldBearingTokenBalance` function - it returns `uint128`, which means we can't implement it for `aTokens` (without overflow) unless we change the return value

if we don't want to overflow, we would need to change the common interface, which would be a medium/major change

smol-ninja: As discussed over call, `getTotalYieldBearingTokenBalance` can return the value of `balanceOf` on aToken which is what wallets and explorers display. All public function could return this instead of scaled version.

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

bob/src/SablierAaveAdapter.sol
andreivladbrg: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/a5b2ff65ae920f40cb83588967aa8088d491196e

also added  "aToken" for explicitness

andreivladbrg:
@smol-ninja marked the PR as ready for review, integration tests (yet no fork) and deploy script were added (besides the feedback address)

bob/src/SablierAaveAdapter.sol
smol-ninja: We can continue to call it `RAY` but why is it a private constant and not public constant?

andreivladbrg: Because it’s used for internal calculations, do you think it’s relevant for the adapter’s API?

smol-ninja: We never used private constant. Thats why I asked.

If we make it a public constant then the only effect is the exposure of q public function but that would have no impact on gas cost.

I have no problem with private constant just pointing out that we either use public constant or private variables.

andreivladbrg: the decision is yours

feel free to change it to public if you think its better for consistency

smol-ninja: We can continue with private constant here

bob/scripts/bash/prepare-artifacts.sh
```diff
@@ -28,16 +28,22 @@ mkdir "$artifacts" \
 bob=./artifacts/
 cp out-optimized/SablierBob.sol/SablierBob.json "$bob"
 cp out-optimized/SablierEscrow.sol/SablierEscrow.json "$bob"
+cp out-optimized/SablierAaveAdapter.sol/SablierAaveAdapter.json "$bob"
```
smol-ninja: Should it be above `SablierBob`? Also should we order the list alphabetically?

bob/scripts/bash/prepare-artifacts.sh
```diff
@@ -28,16 +28,22 @@ mkdir "$artifacts" \
...
 
 bob_interfaces=./artifacts/interfaces
 cp out-optimized/IBobVaultShare.sol/IBobVaultShare.json "$bob_interfaces"
 cp out-optimized/ISablierBob.sol/ISablierBob.json "$bob_interfaces"
 cp out-optimized/ISablierEscrow.sol/ISablierEscrow.json "$bob_interfaces"
+cp out-optimized/ISablierAaveAdapter.sol/ISablierAaveAdapter.json "$bob_interfaces"
 cp out-optimized/ISablierLidoAdapter.sol/ISablierLidoAdapter.json "$bob_interfaces"
 
 bob_external=./artifacts/interfaces/external
+cp out-optimized/IAavePool.sol/IAavePool.json "$bob_external"
```
smol-ninja, line 43: I think we made a mistake. Why do we need to export these aave, curve and lido interfaces? A user would anyways use the actual interface from their own repos since our versions are minimal.

So as a suggestion we do not need the following interfaces in this script:

- IAavePool
- IAavePoolAddressesProvider
- IAavePoolDataProvider
- IAaveToken
- ICurveStETHPool
- ILidoWithdrawalQueue
- IStETH
- IWETH9
- IWstETH

which means we dont need `bob_external` at all.

bob/scripts/solidity/AaveAdapterUtils.s.sol
```diff
@@ -0,0 +1,20 @@
...
+
+import { UD60x18 } from "@prb/math/src/UD60x18.sol";
+import { ChainId } from "@sablier/evm-utils/src/tests/ChainId.sol";
+
+/// @notice Aave Adapter utility functions for deploy scripts.
+abstract contract AaveAdapterUtils {
+    UD60x18 internal constant INITIAL_YIELD_FEE = UD60x18.wrap(0.1e18); // 10%
+
+    /// @dev Returns the Aave V3 PoolAddressesProvider for the current chain.
+    /// See https://docs.aave.com/developers/deployed-contracts/v3-mainnet
```
smol-ninja: I don't think this URL exists since its redirecting to default aave docs. 

This should be https://aave.com/docs/resources/addresses

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,342 @@
...
+    mapping(uint256 vaultId => mapping(address user => uint256 scaledATokenBalance)) private _userATokenScaledBalances;
+
+    /// @dev The aToken address for each vault's underlying token.
+    mapping(uint256 vaultId => IAaveAToken aToken) private _vaultATokens;
+
+    /// @dev The underlying token address for each vault.
+    mapping(uint256 vaultId => address token) private _vaultTokens;
+
+    /// @dev The total scaled balance of aTokens for each vault.
+    mapping(uint256 vaultId => uint256 totalScaledBalance) private _vaultATokenTotalScaledBalance;
```
smol-ninja: Note that above variables are plural but this and the next one is singular. We should either use singular versions for mapped variables or plural.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,342 @@
...
+
+    /// @dev Yield fee snapshotted for each vault.
+    mapping(uint256 vaultId => UD60x18 fee) private _vaultYieldFee;
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                    CONSTRUCTOR
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Deploys the Aave V3 adapter.
+    /// @param aavePoolAddressesProvider The address of the Aave V3 PoolAddressesProvider (chain-specific).
```
smol-ninja: No need to say "chain-specific" as its implicit.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,342 @@
...
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @inheritdoc ISablierBobAdapter
+    function getTotalYieldBearingTokenBalance(uint256 vaultId) external view override returns (uint128) {
+        uint256 totalATokenScaled = _vaultATokenTotalScaledBalance[vaultId];
+        if (totalATokenScaled == 0) {
+            return 0;
+        }
+        uint256 currentIndex = AAVE_POOL.getReserveNormalizedIncome(_vaultTokens[vaultId]);
+        return (totalATokenScaled * currentIndex / RAY).toUint128();
```
smol-ninja: Since this logic is used multiple times we can create a helper function that converts scaled amount into normalized one and call it `normalizeWithAaveIndex`.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,342 @@
...
+        // If the user's share is greater than their original deposit, the yield is positive and we need to calculate
+        // the fee.
+        if (userTokenShare > shareBalance) {
+            uint128 yieldAmount = userTokenShare - shareBalance;
+
+            // Calculate the fee.
+            feeAmountDeductedFromYield = ud(yieldAmount).mul(_vaultYieldFee[vaultId]).intoUint128();
+            transferAmount = userTokenShare - feeAmountDeductedFromYield;
+        }
+        // Otherwise, the yield is negative or zero, so no fee is applicable.
```
smol-ninja, line 206: In case of aave lending, there is no concept of slashing. So, its not possible for yield to be negative or zero.

IMO we can safely remove this comparison from here and plainly deduct fee.

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,342 @@
...
+        (address aTokenAddress,,) = IAavePoolDataProvider(AAVE_POOL_DATA_PROVIDER).getReserveTokensAddresses(vaultToken);
+
+        // Check: aToken address is not the zero address.
+        if (aTokenAddress == address(0)) {
+            revert Errors.SablierAaveAdapter_TokenNotSupportedByAave(address(vaultToken));
+        }
+
+        // Effect: store the vault token and aToken for future lookups.
+        _vaultTokens[vaultId] = vaultToken;
+        _vaultATokens[vaultId] = IAaveAToken(aTokenAddress);
```
smol-ninja: Why dont we just call it `aTokens` instead of `_vaultATokens`. These two names are so similar that its very hard to distinguish them as different names.

bob/src/interfaces/external/IAavePool.sol
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.22;
+
+/// @notice Minimal interface for Aave V3 Pool — supply/withdraw only.
```
smol-ninja: Lets remove "supply/withdraw only" as it will need to be updated everytime we add a new function to this interface. It already has `getReserveNormalizedIncome`

bob/src/interfaces/external/IAavePoolAddressesProvider.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.22;
+
+/// @notice Minimal interface for Aave V3 PoolAddressesProvider.
+/// @dev One per chain/market. Used to resolve Pool and DataProvider addresses.
```
smol-ninja: "One per chain/market" this is implicit as its common to have different addresses on a different chain (for ex Sablier addresses)

"Used to resolve Pool and DataProvider addresses" we should remove this too since it needs to be updated if we add more functions to it

bob/src/interfaces/external/IAavePoolAddressesProvider.sol
```diff
@@ -0,0 +1,17 @@
...
+interface IAavePoolAddressesProvider {
+    /*//////////////////////////////////////////////////////////////////////////
+                                READ-ONLY FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    /// @notice Returns the address of the Pool proxy.
+    function getPool() external view returns (address);
+
+    /// @notice Returns the address of the PoolDataProvider.
```
smol-ninja: `PoolDataProvider` should be `pool data provider`

bob/src/interfaces/ISablierAaveAdapter.sol
```diff
@@ -0,0 +1,42 @@
...
+    /// @param vaultId The ID of the vault.
+    /// @return The total amount of underlying tokens received from withdrawing the vault's Aave position.
+    function getTokensReceivedAfterUnstaking(uint256 vaultId) external view returns (uint256);
+
+    /// @notice Returns the total aToken scaled balance for a vault.
+    /// @dev The aToken scaled balance is the sum of all users' aToken scaled balances. It is used as the
+    /// denominator in proportional redemption calculations.
+    /// @param vaultId The ID of the vault.
+    /// @return The total aToken scaled balance for the vault.
+    function getATokenTotalScaledBalance(uint256 vaultId) external view returns (uint256);
```
smol-ninja: I dont like this name as `A` looks very weird. One suggestion from my side:

- `getAaveTokenBalanceScaled`

and `getATokenUserScaledBalance` can be called `getAaveTokenBalanceScaledFor`

We used the same terminology in Flow i.e. `scaled` as a suffix.

If you like this, we can follow similar pattern in the `AaveAdapter` contract too.

bob/tests/bob/fork/LidoAdapter.t.sol
```diff
@@ -7,7 +7,7 @@ import { IWETH9 } from "src/interfaces/external/IWETH9.sol";
 
 import { Fork_Test } from "./Fork.t.sol";
 
-contract Bob_Fork_Test is Fork_Test {
+contract LidoAdapter_Fork_Test is Fork_Test {
```
smol-ninja, line 10: Good idea. But I think there should be a separate `Bob_Fork_Test` test too for non adapter vaults. 

I will add it later in my review fork test PR (since so far all fork tests in Bob are AI generated and hardly went through any review).

bob/tests/bob/utils/Constants.sol
```diff
@@ -5,6 +5,9 @@ import { UD60x18 } from "@prb/math/src/UD60x18.sol";
 import { BaseConstants } from "@sablier/evm-utils/src/tests/BaseConstants.sol";
 
 abstract contract Constants is BaseConstants {
+    // Amounts (WBTC - 8 decimals)
```
smol-ninja: This is redundant and `WBTC_DEPOSIT_AMOUNT` should be part of `// Amounts` section.

bob/tests/bob/utils/Modifiers.sol
```diff
@@ -32,6 +32,14 @@ abstract contract Modifiers is Constants, EvmUtilsBase {
         _;
     }
 
+    modifier givenTotalScaledBalanceNotZero() {
+        _;
+    }
+
+    modifier givenTotalTokensReceivedNotZero() {
```
smol-ninja: Alphabetically not ordered.

bob/tests/bob/utils/Types.sol
```diff
@@ -9,6 +9,8 @@ struct VaultIds {
     uint256 nullVault;
     // A settled vault.
     uint256 settledVault;
+    // A vault with Aave adapter enabled.
+    uint256 vaultWithAaveAdapter;
```
smol-ninja, line 13: Just a note (no need to change right now) but `vaultWithAdapter` should be renamed to `vaultWithLidoAdapter` given that we have two adapters now.

smol-ninja (changes requested):
Left a new review of the code (I didn't review tests yet and only looked at them briefly)

bob/src/SablierAaveAdapter.sol
```diff
@@ -0,0 +1,377 @@
...
+        // would drain ALL vaults' positions.
+        uint256 actualWithdrawn = AAVE_POOL.withdraw(address(vaultToken), currentValue, address(this));
+        amountReceivedFromUnstaking = actualWithdrawn.toUint128();
+
+        // Effect: store the total tokens received. This is the denominator used in {processRedemption} to calculate
+        // each user's proportional share: `userScaled / totalScaled * tokensReceived`.
+        _tokensReceivedAfterUnstaking[vaultId] = amountReceivedFromUnstaking;
+
+        // Interaction: transfer tokens to SablierBob for distribution.
+        vaultToken.safeTransfer(SABLIER_BOB, amountReceivedFromUnstaking);
```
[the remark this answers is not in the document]
smol-ninja: This is a good point. @andreivladbrg `AAVE_POOL.withdraw` can revert if there is not enough liquidity which is possible if the utilization ratio is very high. We should check for available liquidity and revert if its not enough.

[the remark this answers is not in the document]
smol-ninja: No change is required from our side since the aave contract will throw error regardless

bob/scripts/bash/prepare-artifacts.sh
andreivladbrg: yep, it should

bob/scripts/bash/prepare-artifacts.sh
andreivladbrg: we could keep them in the `external` dir for extra explicitness but we can also remove them, anything works on my side

bob/scripts/solidity/AaveAdapterUtils.s.sol
andreivladbrg: yes, it should be: https://aave.com/docs/resources/addresses

i think they changed it after the v4 release as it was a good link

bob/src/SablierAaveAdapter.sol
andreivladbrg: plural to be

bob/src/SablierAaveAdapter.sol
andreivladbrg: yep, it is implicit

bob/src/SablierAaveAdapter.sol
andreivladbrg: i thought about it, but the amounts differ (`totalATokenScaled` vs `userATokenScaled`) - i will see how it looks, but i think i tend more towards keeping it as it would increase the gas cost and it's just a one-liner

bob/tests/bob/fork/LidoAdapter.t.sol
andreivladbrg: > for non adapter vaults

currently the tests are only for vaults with adapter

bob/src/SablierAaveAdapter.sol
andreivladbrg: good idea, we can call it `_aTokens` and also change it to be token-based, not id-based, so that we don't store the same a token in case of multiple vaults with the same token:

```solidity
mapping(uint256 vaultToken => IAaveAToken aToken) private _aTokens;
```

wdyt?

bob/scripts/bash/prepare-artifacts.sh
smol-ninja: Yeah lets keep them for extra explicitness.

bob/src/SablierAaveAdapter.sol
smol-ninja: Ok. I also agree with comparing both and see how it looks + gas cost but having a separate function would also make it less error prone because any change (including renaming) here would mean changing it at all instances.

bob/src/SablierAaveAdapter.sol
smol-ninja: This is an incredible idea.

bob/tests/bob/utils/Modifiers.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/tests/bob/utils/Constants.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/interfaces/ISablierAaveAdapter.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/SablierAaveAdapter.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/SablierAaveAdapter.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/SablierAaveAdapter.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/scripts/bash/prepare-artifacts.sh
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/SablierAaveAdapter.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/interfaces/external/IAavePool.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/interfaces/external/IAavePoolAddressesProvider.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/interfaces/external/IAavePoolAddressesProvider.sol
andreivladbrg: f972801492e7fbb3a1166ed72bb62427d62e6550

bob/src/SablierAaveAdapter.sol
andreivladbrg: you are correct about the slashing                                                                                                        
                                                                                                                                            
but it would be a problem if someone enters the vault and the unstaking is done in the same block, so the yield stays the same. then, because the aave index remains unchanged and the calculation rounds down due to division, it can lead to `userTokenShare <= shareBalance`

so, we should keep the `if`

bob/scripts/solidity/AaveAdapterUtils.s.sol
andreivladbrg: 97f3502c1ba6b999498857e596165ef5dcc073fb

andreivladbrg:
@smol-ninja i responded to all your comments. if the fixes look good, could you mark them as resolved and approve the PR?

bob/src/SablierAaveAdapter.sol
smol-ninja: In that case, `yieldAmount` would calculated as zero anyways. The `if` branch is more useful for negative yield and not zero yield. Therefore, I still think the `if` is redundant.

andreivladbrg: actually it can be: `userTokenShare = shareBalance - 1`

  - alice stakes `100e18`. aave: `scaled = rayDiv(100e18, 1.7e27) = 58823529411764705882` (round-half-up).      
  - same block, unstake. `_calculateUnderlyingAmount` uses floor division (`scaled * index / RAY`), not `rayMul`, so:
 `vaultTokenBalance = floor(58823529411764705882 * 1.7) = 99999999999999999999 = 100e18 - 1`                        
  - in `processRedemption`: `userTokenShare = 100e18 - 1`, `shareBalance = 100e18`                                     
                                                                                                                         
  without the if, `uint128 yieldAmount = userTokenShare - shareBalance` underflows and the function reverts

smol-ninja: I guess its possible

smol-ninja (approved):
Pushed a commit: https://github.com/sablier-labs/evm-monorepo/pull/1489/changes/5a635772656f365a521302741dd429f8c070ffeb. Feel free to revert back some of the changes if you disagree.

I found some of the comments verbose or implicit so changed them while reviewing the PR.

andreivladbrg:
@smol-ninja looks good, merging

PR #1491, opened 2026-03-30
feat: add enumerable custom fee user getter

### LINKED ISSUE #1414: Add getter for users with custom min fee USD
## Problem

There's no way to query which users have a custom min fee USD set in the Comptroller, nor to retrieve their custom fee values in bulk. The `customFeesUSD` data is stored in a `mapping(address => CustomFeeUSD)` inside `ProtocolFees`, which means you can only look up a specific user if you already know their address — there's no enumeration.

This makes it hard to audit or display the full list of users with custom fees from off-chain tools, dashboards, or scripts.

## Solution

Add a getter that returns all users with custom min fee USD values for a given protocol, along with their fee amounts. This would require:

1. Maintaining an array (or `EnumerableSet`) of users with custom fees, updated in `setCustomFeeUSDFor` and `disableCustomFeeUSDFor`
2. A new view function, e.g.:

```solidity
function getCustomFeeUsers(Protocol protocol) external view returns (address[] memory users, uint256[] memory fees);
```

> [!NOTE]
> Since mappings can't be iterated in Solidity, this requires storing an auxiliary data structure (array or `EnumerableSet`) alongside the existing mapping. The storage overhead is minimal but worth noting for gas considerations on writes.

rongarede:
## Summary

Closes #1414.

Adds `getCustomFeeUsers(Protocol)` to `SablierComptroller`, enabling off-chain enumeration of all users with custom fees for a given protocol.

### Implementation

- **Storage**: New `mapping(Protocol => EnumerableSet.AddressSet) private _customFeeUsers` at slot 5, with `__gap` reduced from 45 to 44 (total 50 slots preserved).
- **Getter**: `getCustomFeeUsers(Protocol)` returns parallel `(address[] users, uint256[] fees)` arrays by iterating the EnumerableSet.
- **Set maintenance**: `setCustomFeeUSDFor` calls `_customFeeUsers[protocol].add(user)`; `disableCustomFeeUSDFor` calls `_customFeeUsers[protocol].remove(user)`.

### Gas Impact

| Function | Before (median) | After (median) | Delta |
|----------|-----------------|----------------|-------|
| `setCustomFeeUSDFor` | 51,740 | 118,664 | +66,924 |
| `disableCustomFeeUSDFor` | 29,498 | 45,385 | +15,887 |
| `getCustomFeeUsers` | N/A | ~15,000–240,000 | new |

The increase is expected — EnumerableSet.add/remove maintain an internal array + mapping for O(1) membership checks and O(n) enumeration.

## Test Plan

- [x] 8 BTT fuzz tests with `.tree` file covering:
  - Empty state, single user, multiple users, zero-fee user
  - Duplicate set (idempotency), cross-protocol isolation (all 5 protocols)
  - Disable removes user, disable non-existent user is no-op
- [x] Order-independent assertions (no coupling to EnumerableSet internals)
- [x] Cross-function coverage: `setCustomFeeUSDFor` and `disableCustomFeeUSDFor` shared helpers verify `getCustomFeeUsers` reflects mutations
- [x] All 184 utils tests pass
- [x] Storage layout verified via `forge inspect --diff`
- [x] Pre-commit hooks pass (forge fmt, solhint, prettier)

### DIFF
--- a/utils/src/SablierComptroller.sol
+++ b/utils/src/SablierComptroller.sol
@@ -8,6 +8,7 @@ import { ERC165Upgradeable } from "@openzeppelin/contracts-upgradeable/utils/int
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IERC165 } from "@openzeppelin/contracts/utils/introspection/IERC165.sol";
+import { EnumerableSet } from "@openzeppelin/contracts/utils/structs/EnumerableSet.sol";
 
 import { IComptrollerable } from "./interfaces/IComptrollerable.sol";
 import { ISablierComptroller } from "./interfaces/ISablierComptroller.sol";
@@ -43,6 +44,7 @@ contract SablierComptroller is
     RoleAdminable, // 3 inherited components
     UUPSUpgradeable // 1 inherited component
 {
+    using EnumerableSet for EnumerableSet.AddressSet;
     using SafeERC20 for IERC20;
 
     /*//////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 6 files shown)


smol-ninja:
Hi @rongarede. Thank you so much for taking the time to submit this PR. I really appreciate your effort and contribution. While I would genuinely love to get it merged, we have temporarily paused all Solidity related development to focus on some other important priorities at the moment. 

I hope you understand that we won't be able to review new PRs at this time so please do not create any PRs. But I hope we can collaborate again in the future.

PR #1493, opened 2026-03-30
test(bob): fix fork tests

andreivladbrg:
These tests failed here: https://github.com/sablier-labs/evm-monorepo/actions/runs/23748418335/job/69183460419?pr=1489


smol-ninja (approved):
Should we create this PR against `main` and then run `git reset --hard main` with the staging branch?

andreivladbrg:
> Should we create this PR against `main` and then run `git reset --hard main` with the staging branch?

it should be on main, it was a mistake

updated

PR #1499, opened 2026-04-03
test(bob): fix fork tests

andreivladbrg:
fork tests failed again:

https://github.com/sablier-labs/evm-monorepo/actions/runs/23950679664/job/69857300476?pr=1489

this fix was suggested by claude, it seems to be working

### DIFF
--- a/bob/tests/bob/fork/Fork.t.sol
+++ b/bob/tests/bob/fork/Fork.t.sol
@@ -14,6 +14,13 @@ import { Base_Test } from "./../Base.t.sol";
 /// @dev The FINALIZE_ROLE is held by the stETH contract on mainnet.
 interface IWithdrawalQueueFinalize {
     function finalize(uint256 _lastRequestIdToBeFinalized, uint256 _maxShareRate) external payable;
+    function prefinalize(
+        uint256[] calldata _batches,
+        uint256 _maxShareRate
+    )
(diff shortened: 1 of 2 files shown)


bob/tests/bob/fork/Fork.t.sol
```diff
@@ -14,6 +14,13 @@ import { Base_Test } from "./../Base.t.sol";
 /// @dev The FINALIZE_ROLE is held by the stETH contract on mainnet.
 interface IWithdrawalQueueFinalize {
```
smol-ninja, line 15: Should we move this to mocks and call it a contract?

smol-ninja (changes requested):

bob/tests/bob/fork/Fork.t.sol
andreivladbrg: it's just an interface - mocks are contracts that have actual logic that mimics the deployed contract
so, i'd say no

smol-ninja (approved):

bob/tests/bob/fork/Fork.t.sol
smol-ninja: It was written by AI. So may be we review the entire fork tests later and find a better way to put this interface but lets keep it as is for now.

PR #1501, opened 2026-04-07
Refactor Bob invariants

smol-ninja:
Partially addresses https://github.com/sablier-labs/evm-monorepo/issues/1417


bob/tests/bob/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,199 @@
...
+        _;
+    }
+
+    /// @dev Selects a random token from the store's token list.
+    modifier useFuzzedToken(uint256 tokenIndex) {
+        IERC20[] memory tokens = store.getTokens();
+        tokenIndex = _bound(tokenIndex, 0, tokens.length - 1);
+        currentToken = tokens[tokenIndex];
+        _;
+    }
```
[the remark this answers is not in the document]
smol-ninja: Tokens are created in invariant setup to this will never underflow.

bob/tests/bob/invariant/README.md
```diff
@@ -1,41 +1,47 @@
...
+   - the value of `isStakedInAdapter` can never change from `false` to `true`.
+   - total supply of share tokens = $`\sum_{\text{vault}}`$ deposits - $`\sum_{\text{vault}}`$ shares burned.
 
-#### Adapter Economics
+4. For a given vault with adapter,
 
-14. (Inv 6) Yield fees must never exceed yield earned.
-15. (Inv 24) Total WETH distributed across all redemptions $`\le`$ `wethReceivedAfterUnstaking`.
-16. (Inv 27) No user receives more than their proportional WETH share based on wstETH attribution.
+   - `amountReceivedFromAdapter` $`\ge`$ $`\sum_{user}`$ transfer amount + fee amount during redemption for each user.
```
andreivladbrg, line 19: i think here it would be a good inv: `sum(amount deposited in vault) >= amountReceivedFromAdapter`
`amountReceivedFromAdapter` containing the yield as well

bob/tests/bob/invariant/Invariant.t.sol
```diff
@@ -28,403 +31,358 @@ contract Invariant_Test is Base_Test, StdInvariant {
...
+        setMsgSender(address(comptroller));
+        bob.setDefaultAdapter(IERC20(address(weth)), adapter);
+
+        // Fund Curve pool and Lido mocks so that they are solvent for unstaking.
+        vm.deal(address(curvePool), 1_000_000 ether);
+        vm.deal(address(lidoWithdrawalQueue), 1_000_000 ether);
+
+        // Deploy 13 tokens: ERC20Mock for decimals 6-17, plus WETH for decimal 18.
+        IERC20[] memory tokenList = new IERC20[](13);
+        for (uint8 d = 6; d <= 17; ++d) {
```
andreivladbrg: why do we need these tokens for all decimals? can't we have like 1 for 6 (USDC like) one token for 8 decimals (WBTC like) and one token for 18 decimals (WETH like)

bob/tests/bob/invariant/handlers/BobHandler.sol
```diff
@@ -1,254 +1,147 @@
...
 
-        // Toggle the default adapter as comptroller.
-        setMsgSender(comptrollerAddr);
-        if (withAdapter) {
-            bob.setDefaultAdapter(weth, ISablierBobAdapter(address(adapter)));
-        } else {
-            bob.setDefaultAdapter(weth, ISablierBobAdapter(address(0)));
-        }
+        // Set target price to be above current oracle price.
+        uint128 targetPrice = 2 * uint128(uint256(oracle.price()));
```
andreivladbrg: unneeded casting


```suggestion
        uint128 targetPrice = 2 * oracle.price();
```

bob/tests/bob/invariant/handlers/BobHandler.sol
```diff
@@ -1,254 +1,147 @@
...
     {
-        // Pick a random vault and user.
         uint256 vaultId = _fuzzVaultId(vaultIdSeed);
-        address user = _fuzzUser(userSeed);
+
+        // Snapshot the vault status before making the call.
+        _recordStatus(vaultId);
 
         // Skip if vault is not active.
         if (bob.statusOf(vaultId) != Bob.Status.ACTIVE) return;
```
andreivladbrg, line 81: shouldn't we move this check first and then `_recordStatus()`?

bob/tests/bob/invariant/handlers/BobHandler.sol
```diff
@@ -1,254 +1,147 @@
...
 
-        // Bound deposit amount.
-        uint128 amount = boundUint128(amountSeed, MIN_DEPOSIT, MAX_DEPOSIT);
+        // Pick a user such that 50% chance to reuse an existing depositor.
+        address user = _pickUser(vaultId, newUser, existingUserSeed);
```
andreivladbrg, line 87: why do we want users to be already by an existing depositor?

i'd rather have more unique depositors, or lower it to 10-25%

bob/tests/bob/invariant/handlers/BobHandler.sol
```diff
@@ -1,254 +1,147 @@
...
 
-        // Snapshot before.
-        BobStore.VaultMeta memory meta = store.getVaultMeta(vaultId);
-        uint256 shareBalBefore = meta.shareToken.balanceOf(user);
-        uint256 tokenBalBobBefore = weth.balanceOf(address(bob));
+        // If token is WETH, then deal ETH and wrap it into WETH.
+        if (address(token) == address(weth)) {
+            setMsgSender(user);
+            vm.deal(user, amount);
+            IWETH9(address(weth)).deposit{ value: amount }();
```
andreivladbrg: why do we need to handle `weth` separately as it follows `IERC20`?

also, why not store the `weth` as `IWETH9` type?

bob/tests/bob/invariant/handlers/BobHandler.sol
```diff
@@ -261,83 +154,41 @@ contract BobHandler is Constants, StdCheats, BaseUtils {
...
-        if (shareBalance == 0) return;
+        // Pick an existing depositor.
+        address[] memory existingUsers = store.getUsers(vaultId);
 
-        // For adapter vaults: snapshot wstETH before redeem auto-unstakes.
-        if (meta.hasAdapter && !store.wstETHSnapshotTaken(vaultId)) {
-            _snapshotWstETH(vaultId);
-        }
+        // Skip if no users.
+        if (existingUsers.length == 0) return;
```
andreivladbrg, line 173: isn't this check redundant as we have the modifier: `vaultCountNotZero`

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
```diff
@@ -0,0 +1,83 @@
...
+        address comptroller_
+    )
+        BaseHandler(store_, bob_, adapter_, weth_, wstEth_, oracle_, comptroller_)
+    { }
+
+    /*//////////////////////////////////////////////////////////////////////////
+                                 HANDLER FUNCTIONS
+    //////////////////////////////////////////////////////////////////////////*/
+
+    function requestLidoWithdrawal(
```
andreivladbrg, line 37: don't you think we should restrict the number of calls to this function?

bob/tests/bob/invariant/handlers/BobHandler.sol
andreivladbrg, line 1: why no `syncPriceFromOracle` function in the handler?

bob/tests/bob/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,224 @@
...
+        if (shareBalance == 0) return;
+
+        uint128 minAmount = 1;
+
+        // For adapter vaults, ensure the transfer amount is large enough so proportional wstETH is not zero.
+        if (address(bob.getAdapter(vaultId)) != address(0)) {
+            uint128 fromWstETH = adapter.getYieldBearingTokenBalanceFor(vaultId, from);
+            if (fromWstETH == 0) return;
+            minAmount = uint128((shareBalance + fromWstETH - 1) / fromWstETH);
+            if (minAmount > shareBalance) return;
```
andreivladbrg: is this check needed? 

if `fromWstETH == 0` can `minAmount > shareBalance`?

bob/tests/bob/invariant/handlers/BobHandler.sol
```diff
@@ -349,156 +200,51 @@ contract BobHandler is Constants, StdCheats, BaseUtils {
...
-        bob.unstakeTokensViaAdapter(vaultId);
+        // Snapshot the vault status before making the call.
+        _recordStatus(vaultId);
 
-        // Record WETH received.
-        uint128 wethReceived = uint128(adapter.getWethReceivedAfterUnstaking(vaultId));
-        store.setUnstakeResults(vaultId, wethReceived);
-    }
+        // Snapshot the isStakedInAdapter before making the call.
+        store.setPrevIsStakedInAdapter(vaultId, bob.isStakedInAdapter(vaultId));
```
andreivladbrg: shouldn't we add this in `redeem` handler as well?

bob/tests/bob/invariant/README.md
```diff
@@ -1,41 +1,47 @@
...
-18. (Inv 12) No ETH should remain stuck in the `SablierBob` contract.
-19. (Inv 18) Per-vault share token `totalSupply` equals the sum of all holder balances.
-20. (Inv 28) `_vaultTotalWstETH` equals the sum of all `_userWstETH` for adapter vaults while staked.
-21. (Inv 29) When shares are burned in `redeem`, `_userWstETH` is cleared for adapter vaults.
-22. (Inv 72) `processRedemption` conservation: `transferAmount + fee` equals the user's proportional WETH share.
+   - `block.timestamp` $`\ge`$ expiry
+
+7. For a settled vault,
+
+   - lastSyncedPrice $`\ge`$ targetPrice and `block.timestamp` $`\lt`$ expiry
```
andreivladbrg, line 31: we can also have `lastSyncedPrice` is constant once `SETTLED`

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
andreivladbrg, line 1: making this a different contract will lead to a lot of calls to these not so relevant functions (`requestLidoWithdrawal` `setYieldFee`) IMO 

i assume this was created to have more calls in the `BaseHandler` contract: `increaseOraclePrice` `simulateLidoSlashing` `simulateLidoYield`, right? (given the functions are targeted in both lido and bob handlers)

bob/tests/bob/invariant/handlers/BaseHandler.sol
```diff
@@ -0,0 +1,224 @@
...
+        adjustTimestamp(timeJumpSeed)
+    {
+        uint256 currentPrice = uint256(oracle.price());
+        uint256 newPrice = currentPrice * 110 / 100;
+        oracle.setPrice(newPrice);
+    }
+
+    /// @dev Raises the wstETH exchange rate, simulating a Lido slashing event. A higher rate means each wstETH unwraps
+    /// to less stETH.
+    function simulateLidoSlashing(
```
andreivladbrg, line 122: i think it would be a good idea to have handler functions as well for `simulateCurveSlippage` and price fluctuations for `stETH/ETH` , i.e. +/- 0.5% occasionally

andreivladbrg (commented):
great work, left some comments below

bob/tests/bob/invariant/README.md
smol-ninja: This would not be a universal invariant because due to varying Curve exchange rates (impacting Curve swap path) as well as Lido slashing (impacting Lido queue path), `amountReceivedFromAdapter` could be less than `sum(amount deposited in vault)`. See [simulateLidoSlashing](https://github.com/sablier-labs/evm-monorepo/blob/bob-invariants/bob/tests/bob/invariant/handlers/BaseHandler.sol#L121) handler I implemented to simulate the same.

bob/tests/bob/invariant/Invariant.t.sol
smol-ninja: My reason for including all all decimals was to check the accuracy of the yield fee calculation. But now I think having only 3 tokens is also sufficient. So, I agree with you.

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: Yeah it should be like that.

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: Having existing depositors help testing the cases of yield calculations where deposits are made at different period in time. These cases are more complex than calculating yield for someone who made only one deposit.

Why do you think its a good idea to have more unique depositors? I think the same user depositing multiple times is more critical piece here.

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: We need to handle `weth` separately because of wrapping/unwrapping mechanism. If this does not use `deposit`, then unwrapping would fail as `weth` contract may not have enough eth to unwrap.

> why not store the weth as IWETH9 type

Then we would be casting it to `IERC20` multiple times.

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: there could be a vault with no depositor so this check is not redundant.

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
smol-ninja: Yes I think we should.

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: I forgot about it.

bob/tests/bob/invariant/handlers/BaseHandler.sol
smol-ninja: I suppose you mean - if `fromWstETH > 0` can `minAmount > shareBalance`. Then yes, this check is redundant.

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: Yes we should.

bob/tests/bob/invariant/README.md
smol-ninja: Thats a good idea. I will add it.

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
smol-ninja: > lead to a lot of calls

Why would there be lots of calls when they are limited to `MAX_ADMIN_CALLS`? Also, how is it related to having a different contract?

> this was created to have more calls

This was created because Lido adapter is a separate contract. For aave, we can have a separate handler contract - making it easier to maintain.

> given the functions are targeted in both lido and bob handlers

Yes but they are separate functions.

bob/tests/bob/invariant/handlers/BaseHandler.sol
smol-ninja: `simulateCurveSlippage` and `simulateLidoYield` essentially have the same impact. Changing slippage changes the output of curve swap. `simulateLidoYield` does the same. This is because unwrapping and curve swap are called together in the unstake function. So having `simulateCurveSlippage` doesn't add any value to the invariant testing.

Re. price fluctuations handler, I would say the same thing.

bob/tests/bob/invariant/Invariant.t.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/handlers/BaseHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/handlers/BobHandler.sol
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/README.md
smol-ninja: https://github.com/sablier-labs/evm-monorepo/pull/1501/changes/49cac12e78c91edad7b4cd43f1c40e9883236de7

bob/tests/bob/invariant/handlers/BobHandler.sol
andreivladbrg: > We need to handle `weth` separately because of wrapping/unwrapping mechanism. If this does not use `deposit`, then unwrapping would fail as `weth` contract may not have enough eth to unwrap.
> 
> >

alright

> Then we would be casting it to IERC20 multiple times

that's not true. `IWETH9` inherits `IERC20`: 9fb41409406fe20fa0341a148c76b61dc7d0124f

smol-ninja: Thanks looks good.

bob/tests/bob/invariant/README.md
andreivladbrg: right

in that case, wdyt about changing the `simulateLidoSlashing` function to support both ways, price increases and decreases? currently it's only going 1% up

smol-ninja: There are two functions: `simulateLidoYield` and `simulateLidoSlashing`. One makes the price go up and another makes it go down. So its already supporting both ways.

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
andreivladbrg: > Why would there be lots of calls when they are limited to `MAX_ADMIN_CALLS`? Also, how is it related to having a different contract?
> 
> >

if we "manually" restrict the num of calls to certain functions, that's something different (was not the case in `requestLidoWithdrawal`, but having a different target handler would lead to more calls to `requestLidoWithdrawal` if we were to place it in the `BobHandler`, as the invariants call all functions equally in a target, and having more functions there would lead to a lower number. that's why we have a create stream handler in the lockup

bob/tests/bob/invariant/handlers/BobHandler.sol
andreivladbrg: good point about the yield. tho the yield is generated by the `sum(deposits)`, not by individual deposits

i think it would be a good idea to have more unique depositors, as this is a more realistic scenario rather than on having multiple deposits into a single vault

but it's a small thing, let's keep it this way

smol-ninja: > good point about the yield. tho the yield is generated by the `sum(deposits)`, not by individual deposits

Yes but what I wanted to test is that if a user deposits multiple times in the same vault, would the yield be calculated correctly? If the calculations work for a user with multiple deposits then the calculation would definitely work for a user with only one deposit.

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
smol-ninja: Thats a fair argument. So what I am thinking is to merge these functions into a new contract called `Adapter.sol` (which can include aave adapter handlers too in the future) and then inherit Bob handler from Adapter handler so there is only one target contract.

Do you agree?

bob/tests/bob/invariant/handlers/BaseHandler.sol
andreivladbrg: right

bob/tests/bob/invariant/README.md
andreivladbrg: right

bob/tests/bob/invariant/handlers/LidoAdapterHandler.sol
andreivladbrg: let's keep it this way currently, and will think of an alternative when i'll work on the aave adapter

andreivladbrg (approved):

PR #1506, opened 2026-04-10
docs: remove root changelog in favor of per-package changelogs

smol-ninja:
Removed root changelog.

Also should we remove [evm-monorepo tags](https://github.com/sablier-labs/evm-monorepo/releases/tag/evm-monorepo%40v1.0.1) too?


PaulRBerg:
Thanks. And yes

PR #1516, opened 2026-04-22
ci: add scheduled deep and fork workflows

### LINKED ISSUE #1317: Add temporarily removed CIs
In [PR #1305](https://github.com/sablier-labs/lockup/pull/1305)

we removed `ci-fork` and `ci-deep` due to issues caused by the complex `monorepo` setup in the GH workflows.

We should also add the `runs` fields in `ci-module` so that we can pass them individually for each protocol.

andreivladbrg:
Closes https://github.com/sablier-labs/evm-monorepo/issues/1317

### DIFF
--- a/.github/workflows/ci-deep.yml
+++ b/.github/workflows/ci-deep.yml
@@ -1,0 +1,88 @@
+name: "CI: Deep Fuzz & Invariant"
+
+on:
+  workflow_dispatch:
+    inputs:
+      integrationFuzzRuns:
+        default: "100000"
+        description: "Integration: number of fuzz runs."
+        required: false
+      invariantRuns:
+        default: "20000"
+        description: "Invariant runs: number of sequences of function calls generated and run."
+        required: false
+      invariantDepth:
+        default: "200"
+        description: "Invariant depth: number of function calls made in a given run."
+        required: false
+  schedule:
+    - cron: "0 3 1 * *" # at 3:00 AM UTC on the 1st of every month
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        module: [airdrops, bob, flow, lockup, utils]
+    uses: "sablier-labs/gha-utils/.github/workflows/forge-build.yml@main"
+    with:
+      cache-path: &cache-path |
+        ${{ matrix.module }}/cache
+        ${{ matrix.module }}/node_modules
+        ${{ matrix.module }}/out
+        ${{ matrix.module }}/out-optimized
+      foundry-version: &foundry-version "v1.5.1"
+      name: "Build"
+      working-directory: ${{ matrix.module }}
+
+  test-integration:
+    needs: ["build"]
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - module: airdrops
+            match-path: "tests/{integration,utils}/**/*.sol"
+          - module: bob
+            match-path: "tests/*/integration/**/*.sol"
+          - module: flow
+            match-path: "tests/{integration,utils}/**/*.sol"
+          - module: lockup
+            match-path: "tests/{integration,utils}/**/*.sol"
+          - module: utils
+            match-path: "tests/{integration,utils}/**/*.sol"
+    uses: "sablier-labs/gha-utils/.github/workflows/forge-test.yml@main"
+    with:
+      cache-path: *cache-path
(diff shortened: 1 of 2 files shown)


smol-ninja:
I was also working on this but will review yours now.

andreivladbrg:
@smol-ninja The current job names in the CI display look like this:

```
CI: Deep Fuzz & Invariant / test-fuzz (bob, tests/*/integration/fuzz/**/*.sol) / Fuzz tests (bob)
CI: Deep Fuzz & Invariant / test-invariant (airdrops, tests/invariant/**/*.sol) / Invariant tests (airdrops)
```

<details><summary>see image</summary>
<p>

image


</p>
</details> 

Is this fine, or would you prefer something cleaner with the protocol name first? e.g.:

```
CI: Deep Fuzz & Invariant / airdrops / Build / Build
CI: Deep Fuzz & Invariant / bob / Fuzz / Fuzz
CI: Deep Fuzz & Invariant / airdrops / Invariant / Invariant
```

.github/workflows/ci-deep.yml
```diff
@@ -0,0 +1,92 @@
...
+
+concurrency:
+  cancel-in-progress: true
+  group: ${{ github.workflow }}-${{ github.ref }}
+
+on:
+  workflow_dispatch:
+    inputs:
+      integrationFuzzRuns:
+        default: "100000"
```
smol-ninja: This looks a lot. What about 10000?

.github/workflows/ci-deep.yml
```diff
@@ -0,0 +1,92 @@
...
+        description: "Integration: number of fuzz runs."
+        required: false
+      invariantRuns:
+        default: "20000"
```
smol-ninja: 1000 runs are also enough given depth is 200.

smol-ninja (commented):
Two comments:

1. Don't we need `notify-on-failure` job that we had before?
2. I thought about reusing the workflows from `ci-module` but it adds more code due to addition of new variables so I think the isolated approach may be better here.

.github/workflows/ci-deep.yml
andreivladbrg: the numbers were taken from: https://github.com/sablier-labs/flow/blob/main/.github/workflows/ci-deep.yml

.github/workflows/ci-deep.yml
andreivladbrg: the numbers were taken from: https://github.com/sablier-labs/flow/blob/main/.github/workflows/ci-deep.yml

.github/workflows/ci-deep.yml
andreivladbrg: since its once per month, thats fine IMO

smol-ninja:
Re. the names, the naming is fine since most of it is truncated by the GH UI except the first and the last words, so its good to have the protocol name in the end. But I'd suggest:

- For deep fuzz and invariant: CI: Deep / .... / Invariant tests (airdrops)
- For scheduled fork: CI: Fork / .... / Invariant tests (airdrops)
 
i.e. dropping Fuzz & Invariant from the ci deep name. Its a deep test so by default you can assume it covers fuzz and invariant. And ci-fork is just a scheduled test and its not deep.

.github/workflows/ci-deep.yml
smol-ninja: But the GitHub Actions also has some constraints on the runs available in free plan. This will run `100000` across 5 protocols so I am not sure if it would be fine even for once a month. We can keep it this way, and lower it if exceeds the the actions compute on the free plan, we can lower it later. 

Having said that, I still insist to keep it lower as its a very large number.

andreivladbrg: alright, lowered it
0f51442943bda6e43459bf7c5c60d8406d1988f7

.github/workflows/ci-deep.yml
andreivladbrg: 0f51442943bda6e43459bf7c5c60d8406d1988f7

andreivladbrg:
> i.e. dropping Fuzz & Invariant from the ci deep name. Its a deep test so by default you can assume it covers fuzz and invariant. And ci-fork is just a scheduled test and its not deep.

dropped it in 0f51442943bda6e43459bf7c5c60d8406d1988f7

smol-ninja (approved):
thanks

PR #1519, opened 2026-04-23
ai: explicitely mention to look in the root "utils/"

andreivladbrg:
i had this issue multiples time - so i think its a good idea to explicitely mention this

### DIFF
--- a/.agents/skills/foundry-testing/references/sablier-conventions.md
+++ b/.agents/skills/foundry-testing/references/sablier-conventions.md
@@ -116,6 +116,22 @@ ______________________________________________________________________
 
 ______________________________________________________________________
 
+## Monorepo Import Resolution
+
+**This is a monorepo.** When test files or contracts import from `@sablier/evm-utils/`, do **NOT** look in
+`node_modules/`. Instead, resolve the source from the **`utils/`** directory at the repo root.
+
+For example, `import { BaseTest } from "@sablier/evm-utils/src/tests/BaseTest.sol"` maps to
+`utils/src/tests/BaseTest.sol`.
+
+This applies to all consumer packages: `airdrops/`, `bob/`, `flow/`, and `lockup/`.
+
+**Version check:** Before modifying or referencing utils code, verify that the version in `utils/package.json` matches
+the `@sablier/evm-utils` dependency version declared in the working package's `package.json`. A mismatch means the
+package may be using a published release rather than the local source.
+
+______________________________________________________________________
+
 ## Commands
 
 ```bash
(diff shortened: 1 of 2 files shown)


.agents/skills/foundry-testing/references/sablier-conventions.md
```diff
@@ -116,6 +116,22 @@ ______________________________________________________________________
 
 ______________________________________________________________________
 
+## Monorepo Import Resolution
+
+**This is a monorepo.** When test files or contracts import from `@sablier/evm-utils/`, do **NOT** look in
+`node_modules/`. Instead, resolve the source from the **`utils/`** directory at the repo root.
```
smol-ninja, line 122: Why? Doesn't packages use evm-utils from node_modules?

.agents/skills/foundry-testing/references/sablier-conventions.md
```diff
@@ -116,6 +116,22 @@ ______________________________________________________________________
...
+
+For example, `import { BaseTest } from "@sablier/evm-utils/src/tests/BaseTest.sol"` maps to
+`utils/src/tests/BaseTest.sol`.
+
+This applies to all consumer packages: `airdrops/`, `bob/`, `flow/`, and `lockup/`.
+
+**Version check:** Before modifying or referencing utils code, verify that the version in `utils/package.json` matches
+the `@sablier/evm-utils` dependency version declared in the working package's `package.json`. A mismatch means the
+package may be using a published release rather than the local source.
```
smol-ninja, line 131: mismatch can happen when utils is in WIP but version is not upgraded (which is typically done during releases)

.agents/skills/foundry-testing/references/sablier-conventions.md
andreivladbrg: they do, but in case of me, it gets more "confused" when it is looking at the `node_modules` - thats why i mentioned to look into the `package.json` version field

.agents/skills/foundry-testing/references/sablier-conventions.md
andreivladbrg: thats true. would like to mention this?

.agents/skills/foundry-testing/references/sablier-conventions.md
smol-ninja: But looking at `package.json` is not enough. For ex, you might be making some changes to utils so you can have a situation where the version in `package.json` matches but the codes are different.

IMO, it should always look into `node_modules` instead of local `utils` when analyzing lockup/flow/airdrops etc.

.agents/skills/foundry-testing/references/sablier-conventions.md
smol-ninja: See https://github.com/sablier-labs/evm-monorepo/pull/1519#discussion_r3130305510. I am arguing towards not mentioning this at all. Its doing the right thing by looking into node_modules. May be the issue that you are facing (Claude getting confused) is something else. 

You can also add these custom instructions to claude memory. Just say add it to your memory.

smol-ninja (approved):
Assuming that there will not be any new development, these changes to claude skills can be helpful. Above concerns only apply if we continue the development in this repo and in that case, I think this should be a part of .ai directory in the repo (gitignored).

andreivladbrg:
thanks

